PR #8, opened 2018-07-10
Unwrap HTML tags in slugify filter

### LINKED ISSUE #7: Strip out HTML tags in slugify filter
For our website, we implemented the ability to render titles as raw HTML, so that content editors can italicize titles embedded within other titles. However, we noticed that the `formatPermalink` function doesn't strip out HTML tags when generating the slug. Since doing so would not impact projects that do not enter HTML into title fields, I'd like to see the `slugify` filter expanded to strip out HTML tags. 

https://github.com/area17/twill/blob/4d58468db4c2aec3bcbb7ed8c8de3dfb78c5240b/frontend/js/mixins/formatPermalink.js#L27

https://github.com/area17/twill/blob/4d58468db4c2aec3bcbb7ed8c8de3dfb78c5240b/frontend/js/utils/filters.js#L4-L16

I'll submit a PR for this shortly!

IllyaMoskvin:
Fixes #7. This makes it so that if you include simple HTML tags in your titles, they will be unwrapped when the titles are transformed into permalink suggestions. This is non-DOM solution for speed.

I do have an alternative solution that uses the DOM:

```javascript
  slugify: function (value) {
    const a = 'àáäâèéëêìíïîòóöôùúüûñçßÿœæŕśńṕẃǵǹḿǘẍźḧ·/_,:;' + 'ąàáäâãåæćęęèéëêìíïîłńòóöôõøśùúüûñçżź'
    const b = 'aaaaeeeeiiiioooouuuuncsyoarsnpwgnmuxzh------' + 'aaaaaaaaceeeeeeiiiilnoooooosuuuunczz'
    const p = new RegExp(a.split('').join('|'), 'g')

    // Perform preliminary normalization
    value = value.toString().toLowerCase().trim()

    // Strip (unwrap) HTML tags using DOM
    const div = document.createElement('div')
    div.innerHTML = value
    value = div.innerText

    // Preform regex replacement
    return value
      .replace(/\s+/g, '-') // Replace spaces with -
      .replace(p, c =>
        b.charAt(a.indexOf(c))) // Replace special chars
      .replace(/&/g, '-and-') // Replace & with 'and'
      .replace(/[^\w-]+/g, '') // Remove all non-word chars
      .replace(/--+/g, '-') // Replace multiple - with single -
  },
```

I know it's generally considered bad practice to parse HTML using regex due to the myriad of edge-cases, but I thought that it might be the more appropriate method in this specific situation.

Even in the case of our website, it's rare for titles to include HTML, and the HTML that will be entered is quite limited. Since the permalink is meant to update as the user types, I feel that speed should be a priority over correctness, and AFAIK, regex is faster than DOM.

### DIFF
--- a/frontend/js/utils/filters.js
+++ b/frontend/js/utils/filters.js
@@ -7,6 +7,7 @@ const filters = {
     const p = new RegExp(a.split('').join('|'), 'g')
 
     return value.toString().toLowerCase().trim()
+      .replace(/<\/?[^>]+(>|$)/g, '') // Unwrap HTML tags
       .replace(/\s+/g, '-') // Replace spaces with -
       .replace(p, c =>
         b.charAt(a.indexOf(c))) // Replace special chars


jeremytubbs:
I would approve this... maybe strip tags for display elsewhere? That way we do not see tags in the table or edit view.

ifox:
Hi @IllyaMoskvin!

Thanks for contributing to Twill, it's awesome to have you part of the open source community.

I can definitely see your use case. I'm a little bit concerned about the implications though. It seems like introducing HTML in titles is opening other issues, like the one @jeremytubbs actually identified. 

I think title should be purely text to avoid having to strip tags for display. 

However, if you need to give editors the ability to add simple tags to title for styling/intent purposes, which is definitely something Twill should not prevent you to do, I would suggest adding a dedicated field to the form and leave the CMS title empty of HTML. The main title column would become the non styled version of your title, which is going to render fine everywhere in the CMS. It might be wise to also use it in SEO metatags as I'm not sure how social networks would react to HTML tags in a title. Since you have full control over how you use the data added through Twill's UI in the frontend part of your application, it is not a problem to render a different field as your title inside the page than the one you would use for say metatags.

Another issue I see is someone wanting to write an article about the history of the <blink> tag (silly, I know :)). If we strip tags from CMS listings, we have a problem with part of the article title not showing up if the article is something like "The history of the silly <blink> tag".

ifox:
And... I got tricked by the very same issue in here. My message meant to be "the history of the silly `<blink>` tag".

IllyaMoskvin:
Hey all – 

Alright, I can see your reasoning – given that Twill targets developers, writing an article about a tag is not a silly use-case at all! I agree that inviting HTML into titles is not a good fit for Twill's core. 

However, I do think this is a good opportunity to talk about how this feature might be implemented in a project where it _is_ a good fit. Cards on the table – I'm interested in how we might go about letting the application override parts of Twill's frontend.

## 1. Fork Twill

This is both the most obvious and the least convenient method. You'd be able to implement this change easily, but you'd incur technical debt from having to keep your fork up-to-date with the upstream.

## 2. Use dedicated field for HTML titles

This is good for all the reasons @ifox mentioned – there'd be no escaped tags scattered throughout the admin interface, and you'd have more control over where you choose to render plaintext vs. HTML titles in your templates.

For normal in-page rendering, we can make the HTML title field optional. If it exists for a given instance, render it raw, else fall back to rendering the required plaintext title field and escape it.

A downside is that if an article with an HTML title needs to be renamed, two fields now need to be changed. This is a rare enough scenario that it could be ignored.

However, I don't trust content editors to (1) only fill out the HTML title field as-needed, and (2) to keep HTML out of the plaintext title. I see two solutions to this issue, most useful in tandem.

First, add [form validation rules](https://laravel.com/docs/5.5/validation#form-request-validation): one to reject the HTML title field if it does _not_ contain HTML, and for our use-case, another to reject the plaintext title if it _does_ contain HTML.

Secondly, copy [views/partials/create.blade.php](https://github.com/area17/twill/blob/1.1/views/partials/create.blade.php) from Twill to the following path in your application:

```
resources/views/admin/partials/create.blade.php
```

Twill's backend is really good about checking to see if the application contains its own version of a template, before falling back on the ones that come packaged with it:

https://github.com/area17/twill/blob/0c4578437ba3538c61c9a27c568bff0a33073103/views/layouts/listing.blade.php#L87

https://github.com/area17/twill/blob/4d58468db4c2aec3bcbb7ed8c8de3dfb78c5240b/src/TwillServiceProvider.php#L272

Copying `create.blade.php` allows us to add a `note` to the `title` field:

```php
@formField('input', [
    'name' => $titleFormKey ?? 'title',
    [...]
    'note' => "Don't use HTML here!"
])
```

The verbiage can use some work, but I'd be happy with this solution.

## 3. Extend filters.js?

Twill's _backend_ is really good about allowing the application to override Twill's templates. I'm not as familiar with Vue.js to say whether the same is true for its frontend at this time.

From prior conversation, A17 suggested that – since we already override `create.blade.php` –  we can change the field's `onChange` attribute to execute our own function, instead of `formatPermalink`:

```php
@formField('input', [
    'name' => $titleFormKey ?? 'title',
    [...]
    'onChange' => 'formatPermalink',
])
```

However, I don't know how to configure the build process to append new functions to the admin, and if there are any namespacing rules in place for these function references. So for now, I'll leave it be.

We also considered the subject of configurable hooks, but there's ambiguity in when hooks should fire. Even in this case, different slug transformations might need to fire at different stages in the `slugify` process, potentially requiring several hook points. Hooks are too situational to be useful holistically.

More broadly, it seems reasonable that there ought to be a way to extend `filter.js` to override only the `slugify` method in the child. I'm not sure to what extent this is possible, but it would be _amazing_ if Twill could do something similar to how it handles the PHP backend, and allow the application to override parts of its frontend in some manner.

Perhaps there's a way to resolve the `@` alias in `webpack.mix.js` to check if the path exists below a given root within the application, and if not, fall back to using the files in Twill's `frontend/js`?

ifox:
Thanks again for your thorough breakdown here @IllyaMoskvin!

As noted in #7, we're investigating your third point for next releases. It is definitely how we see it too. It should be possible to override very specific parts of the frontend codebase by following a convention like we already have on the PHP side.

PR #44, opened 2018-10-19
Fix the bug that no error message pops up when user give incorrect credential on login page

yanhao-li:
This commit separated the notification component from footer, and include the new toaster component on login page to give a simple fix of the error.

As the next step, we may should remove all old notification part from footer and include the new toaster component in layout.


ifox:
Looking good to me @yanhao-li thanks a lot!

As a next step, the reset password page, which currently use the shared footer, could be setup to have the Twill logo below the form field, like on the login page, but that's more of a branding choice than a feature or bug, so I'm good with merging this now to fix the wrong credentials issue you identified.

PR #52, opened 2018-10-29
Create a simple way to detect the presence of an image

ferpetrelli:


ifox:
Fixed pull request merged in #53.

PR #53, opened 2018-10-29
Implement an easy way to check for images existence

ferpetrelli:


ifox:
Thanks @ferpetrelli! Definitely helpful.

PR #64, opened 2018-10-31
Cartalyst/tags supported Laravel 5.7.  Removed note in docs related to the issue

yanhao-li:


ifox:
Thanks @yanhao-li!

Would you mind separating your 2 changes in 2 different PRs?  I'm all for the database connection check before install, but it is a different concern than updating the docs about cartalyst/tags. 

At the same time, if you can cleanup the branch you've sent this PR from, that'd be great. I'd like to avoid polluting git history with merge commits. I can always squash before merging, but since I'm asking you to separate in 2 pull requests, I guess you can cleanup your branch first, if you don't mind.

PR #66, opened 2018-11-01
check database connection before twill:install

yanhao-li:
The twill:install command entails database connected, however, if the user failed to do that, this command will still try to migrate the database and incur errors. In that case, rerun the command can't fix the error, users may have to removed all migrated tables manually and run the command again. 

This commit will do the database check before running php artisan twill:install to prevent this.


ifox:
Awesome @yanhao-li, thanks!

PR #91, opened 2018-11-26
Fixed the "offset 'en'" error while updating a setting form which include translate disabled field

yanhao-li:
This PR include the following two commits,
The first commit fixed the error happened when update the setting, by improving the setting update strategy.
The second commit changed A17TextField's value render strategy by converting the passed value to a reasonable string, avoid the situation that in some cases, it will display "[object object]", not only in this case, but also may happen elsewhere.


yanhao-li:
Awesome, Thanks @ifox

PR #100, opened 2018-12-04
To enable HTML rendering for ItemList's name property

yanhao-li:
The purpose of this commit is to enable ItemList name property to display html content when find 'html' key passed in.


ifox:
Hi Yanhao!

I think we should avoid using `html` as the key to identify the intent of a developer to render the browser `name` field as HTML, to avoid potential conflicts with a model attribute named `html`. What do you think about `renderHtml` instead?

yanhao-li:
@ifox sounds good, will push a new commit later today!

PR #113, opened 2018-12-13
Update to Laravel Mix 4

nsivertsen:
Laravel Mix 4 includes Webpack 4, which is much faster than Webpack 2, which is included with Mix 2. On my machine this has reduced build times by half.

Notes:

- [Overview of changes in Mix 4](https://laravel-mix.com/docs/4.0/upgrade)
- Mix 4 comes with Babel 7, which [no longer has](https://babeljs.io/docs/en/babel-preset-stage-2) `babel-preset-stage-*`presets. This means all features from stage 2 have to included individually as plugins. It appears Twill only uses class properties from stage 2, so I only included that.
- Mix 4 uses Dart Sass by default. We can't use that here, because Dart Sass doesn't allow the `/deep/` combinator used by Vue (and neither Dart Sass nor Node Sass allow `>>>`).

  For Sass inside Vue components, Mix will detect which version of Sass to use automatically based on which version you have installed, meaning `node-sass` has to be included in `package.json`. For `*.scss` files, we have to include `{ implementation: require('node-sass') }` as a third argument to `mix.sass()`.

  Small caveat: from my testing it appears that when using the `twill-build` scripts from *your own project*, when compiling Vue components that use Sass, Mix [looks](https://github.com/JeffreyWay/laravel-mix/blob/da4c2d0518dc9342eb15e2da16be5dd3e0e22fe1/src/components/Sass.js#L13) [there](https://github.com/JeffreyWay/laravel-mix/blob/da4c2d0518dc9342eb15e2da16be5dd3e0e22fe1/src/Mix.js#L70-L78) for `node-sass`, and if it doesn't find it, it resorts to Dart Sass, which of course fails because of `/deep/`. This is fixed by including `node-sass` in your own project. If it is installed, Mix will prefer that.


ifox:
Hi @nsivertsen, so nice to see you here!

Thank you so much for your work on this. I actually went through the same process as I followed along the Mix 3 and 4 releases, first updating to Babel 7, then to Webpack 4.

It appears we both ended up at the same conclusion regarding `dart-sass` and `node-sass`, however I would really like to dig deeper into trying to move away from `node-sass`, as you're probably aware of its issues regarding its installation on various OS configurations as well as its stability when new versions of node are released.  Also, as I'm working on moving away from requiring developers to setup npm scripts manually, replacing it with an more streamlined artisan commands set that will make updates easier and will hide the complexity of the build configuration, requiring developers to include `node-sass` in their project kinda goes against it.

I was so happy to see that Mix now supports the Dart implementation by default but was pretty bummed seeing the lack of `/deep/` or `>>>` support.

@mrdoinel @m4n1ok any ideas?

m4n1ok:
Hi all, 

It could be awesome if we can have classic artisan commands and remove node-sass requirement.
Currently, my main idea is to remove all `/deep/` in Vue components and use a different way to override child component style. Because css deep selector is now deprecated and like the philosophy chosen by `dart-sass` to not use invalid selectors.  

As solution, I think about a combination of `<style scoped>`and expose some elements in non scoped `<style>`in same component. 
Another way will be to use old friend `!important` on some rules... 

I don't know what is the best way to follow. I can try some experimentations soon.

nsivertsen:
I agree that it would be better not to require including `node-sass`. So I can close this pull request?

Using `dart-sass` would also make clean builds inside containers a lot faster (that's how I'm developing a Twill site at the moment).

I also agree with @m4n1ok that moving away from nonstandard selectors is probably better.

Another thing to keep in mind is that the current setup of building the assets in the `vendor/area17/twill` context and then copying them over to the user's project breaks source maps. I see two ways of fixing that:

1. rewriting the paths of source maps in the build script
2. building in the context of the project.

One way to achieve 2. and also to hide the complexity that you mentioned might be to package up the frontend code in an npm package that the user imports and which includes a script that can be run from the project. Something like:

```json5
{
  // ...
  scripts: {
    production: "*user's own frontend build script here*",
    prod: "twill build && npm run prod"
  },
  devDependencies: {
    "twill": "^1.2.0",
    // ...
  },
  // ...
}
```

nsivertsen:
Another problem with the current build script is that this part:

```bash
cd vendor/area17/twill && npm ci && npm run prod && cp -R public/* ${INIT_CWD}/public
```

...replaces `mix-manifest.json`. The new build script should probably handle merging the manifests for the frontend and the backend.

ifox:
Hi @nsivertsen!

With @m4n1ok's work on removing the deep selector, we are going to be able to merge this after a rebase and removal of node-sass 👌. I'm testing the various components for any regression, will let you know when I merge #127!

nsivertsen:
Hi @ifox :) That's great news!

In fact, I have been working on a build script that solves all of those problems. I'll need a bit more time to make it presentable, but in essence it works like this:

The user installs Twill's npm package as a local package from the vendor folder:

```
npm install ./vendor/area17/twill
```

This adds the executable `twill`, so now you can do stuff like this:

```
twill build --blocks resources/js/blocks --output public/assets/admin
twill watch --blocks resources/js/blocks --output public/assets/admin
```

And you can leave the options off if you want to keep the defaults and just do

```
twill build
twill watch
```

A few notes:

1. Installing the package like this means that shared dependencies between Twill and the user's setup can actually be shared in `node_modules`, speeding up install times
2. Nothing is ever copied, so all source maps stay intact
3. Cross-platform (using fs-extra)
4. Much faster than the current build script

Give me a few days to clean everything up and I'll give you a demo.

In the meantime, I'll rebase this PR.

nsivertsen:
Happy new year to everyone! I've rebased and switched out `node-sass` for `sass`. Once #127 is merged, I'll rebase one last time and then we should be good to go! (Note that currently building doesn't work with this PR because `node-sass` is already gone, but we still have `/deep/` selectors.)

I've been on a pretty tight deadline, and so haven't got around to cleaning up that build script mentioned above, but it's been running without problems for the past month. The site I'm using it for (my first Twill site) is going live at the end of the month. Can't wait to share that with you!

tchapi:
Hi all
Is the manifest merging issue fixed somehow ? On a fresh install of Twill and Laravel, using Mix for my frontend, I still have the issue. Maybe I overlooked something.

Thanks for your help !

Issue #128, opened 2018-12-19
updateRepeater and getFormFieldsForRepeater need to be updated


yanhao-li:
Currently in order to make the contentFields repeater to get to work, we have to define the repeater name and relation name to have the same name, which make no sense.  Per `HandleRepeater.php` and #51 .

I am feeling the `updateRepeater` and `getFormFieldsForRepeater` functions need to be updated to have an extra `$repeaterName` parameter passed in.

PR #129, opened 2018-12-19
update handleRepeater trait functions

yanhao-li:
Upon the issue https://github.com/area17/twill/issues/128


ifox:
Hi @yanhao-li, thanks for this!

I agree with you about the idea, however the changes you are making here introduce a breaking change for Twill users. They would have to update all calls to getFormFieldsForRepeater and updateRepeater. I do not think this is worth releasing a major version so I would suggest adding this parameter as the last parameter of both function, defaulting to the relation parameter when not provided.

yanhao-li:
Hi @ifox,

You are right! I have updated the pull request.
Please let me know if there still have problems!

PR #144, opened 2019-01-10
add 'hidden' attribute to input field

zipavlin:
Add a 'hidden' html attribute to input field. Hidden field(s) might be needed when auto populating field value. For example, when auto populating a 'title' field.

### DIFF
--- a/views/partials/form/_input.blade.php
+++ b/views/partials/form/_input.blade.php
@@ -6,6 +6,7 @@
     $placeholder = $placeholder ?? false;
     $maxlength = $maxlength ?? false;
     $disabled = $disabled ?? false;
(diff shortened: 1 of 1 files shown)


ifox:
Hi @zipavlin,

Hidden fields can definitely be very useful and we actually already have a `@formField` for that, under `views/partials/forms/_hidden.`, that you can use like this:

```php
@formField('hidden', [
    'name' => 'hidden_field_name',
    'value' => 'hidden_field_value'
])
```

It doesn't have to overhead of the more complex input field, but maybe there's a feature of the input field that you're looking at using on an hidden field? If not, please close this and I will add the `hidden` field to the docs on my end!

PR #149, opened 2019-01-14
Add dynamic title for repeater items

zipavlin:
Add option to set title of repeater item (block) dynamically based on a field inside repeater item.

capture

Use by setting a ```titleField``` in repeater array in ```cofig/twill.php```:
```php
return [
    "block_editor" => [
        'blocks' => [...],
        ...
        'repeaters' => [
            'accordion_item' => [
                'title' => 'Accordion Item',
                'titleField' => 'title', // name of field inside repeater item that will be used instead of default 'title' 
                'trigger' => 'Add accordion item',
                'component' => 'a17-block-accordion_item',
                'max' => 10,
            ],
        ],
        ...
    ]
];
```


frontend/js/components/Repeater.vue
```diff
@@ -77,7 +77,22 @@
...
         get () {
           if (this.savedBlocks.hasOwnProperty(this.name)) {
-            return this.savedBlocks[this.name] || []
+            const blocks = this.savedBlocks[this.name]
+            if (blocks && blocks.length > 0) {
+              const repeaterName = this.name.replace(/(blocks-\d+_)/, '')
+              const titleFieldToMap = this.$store.state.repeaters.availableRepeaters[repeaterName] && this.$store.state.repeaters.availableRepeaters[repeaterName].titleField ? this.$store.state.repeaters.availableRepeaters[repeaterName].titleField : null
+              if (titleFieldToMap) {
+                return blocks.map(block => {
+                  const titleField = this.$store.state.form.fields.find(field => field.name === `blocks[${block.id}][${titleFieldToMap}]`)
```
m4n1ok: I'm not very fan of that here. I'm thinking about performances issues with a repeater with a lot of blocks.

Currently, I see one different solution:
Pass `titleField` to map as a `prop` inside `Blocks.vue` component and move logic to get the value of field inside `Blocks.vue` component.

**Pros:**
* It will be update only one block and not try to update all blocks when a user update field value.
* We can use this behavior with content editor.

**Cons:**
* It will add a dependency to vuex store inside `Blocks.vue` component.

frontend/js/components/Repeater.vue
```diff
@@ -77,7 +77,22 @@
...
+                  if (titleField && titleField.value) block.title = titleField.value
```
m4n1ok: There is an issue with `titleField.value` if we have a form with multi languages enabled.
The returned value is an object with value for each language. We need to detect if the form have multi languages and get the value for the current language.

screenshot 2019-01-17 at 22 25 02

philippdaun:
Any progress on that feature? As I need this functionality for an upcoming project I'd be happy to give it a stab myself. Not too familiar with the codebase, however, so not sure what has changed in the releases since.

ifox:
Added in #1096, thanks for suggesting this!

Issue #126, opened 2018-12-18
Pages module


zipavlin:
First of, thanks for sharing your work, it looks amazing!

I am trying to create a pages module, but I am not sure how to go about it.

Documentation states that '_Pages are unstructured data models most often used for static content, such as an About page [...]_', which I understand should not be prefixed with 'pages'. I also don't know how to arrange it in page/subpage/... convention. Do you maybe have any more detailed docs about it?

I will also need pages to have different page templates with different fields. I read #54 and will try to implement it somehow, but I would appreciate if you have any ideas how to do that or where to start?

Also related: is it/or would be possible to implement a sort of addons system. For example a 'pages' addon that would scaffold all of this (tree structure, templates, ...) and leave user to populate template' fields, because it seems like something that a lot of projects requiring CMS would actually need. 

Thanks for your help!

zipavlin:
Thank you @noxify the gists works perfectly! Now, I just need to figure how to make block(s) not flexible (so I don't need to have pages table with 50 columns) and I have everything :)

Your solutions for pages prefix doesn't work in my case, but I might need to try again, as there was no change after implementing either or both of those ($indexOptions, $permalinkBase).

Thanks again for your help, you are awesome! 👍

ifox:
Hi @zipavlin, sounds good!

Docs are open to contributions, the `docs` folder is a [Vuepress](https://vuepress.vuejs.org/) project 🤓

Like I just mentioned on https://github.com/area17/twill/issues/46#issuecomment-452409567, a guides section would definitely makes sense, great suggestion!

frixxx:
Has anyone used this nested module stuff? Is there a way to get nested slugs?

F.e. i have the following items:
- Item A - Nesting Depth 1
   - Item B - Nesting Depth 2

And now i want to have the slug /en/item-a/item-b) for the Item B and /en/item-a for the Item A.
Did anyone do this? Maybe someone could help?

amirakbulut:
> Has anyone used this nested module stuff? Is there a way to get nested slugs?
> 
> F.e. i have the following items:
> 
> * Item A - Nesting Depth 1
>   
>   * Item B - Nesting Depth 2
> 
> And now i want to have the slug /en/item-a/item-b) for the Item B and /en/item-a for the Item A.
> Did anyone do this? Maybe someone could help?

Have you solved this yet?

vacoup95:
> > Has anyone used this nested module stuff? Is there a way to get nested slugs?
[...]

PR #171, opened 2019-01-23
Add documentation for creating nested module in Twill

yanhao-li:
Reference: https://github.com/area17/twill/issues/126


vrdvishwas:
vendor/kalnoy/nestedset/src/NodeTrait.php"
line: 1181
message: "Node must exists."
type: "LogicException"

got this error,

On load it works for one move, but after second move it showing this error

Stiropor:
@vrdvishwas I just came to same issue. After some digging around I found that adding `self::fixTree();` in the beginning of module's `saveTreeFromIds()` method fixed the problem. Works like a charm now.

Issue #97, opened 2018-12-01
Physically remove uploaded images


awgul:
Hey,

I set `MEDIA_LIBRARY_CASCADE_DELETE` to `true`.
But now I can only mark multiple images (earlier I could mark only one image).
And when I hit delete the images are not physically deleted at all.

This `MEDIA_LIBRARY_CASCADE_DELETE` is working here like: select multiple files or not.

I use endpoint_type: local.

ifox:
Hi @awgul!

There are no relationship in code between `MEDIA_LIBRARY_CASCADE_DELETE` and being able to select multiple images in the media library. Are you sure you were in 2 different contexts when testing that? Like once by opening the media library from the navigation, and once by opening the media library through a medias form field. When you open it from the navigation, you're able to select multiple, but if you're coming from a Add image button on a medias form field, you're only able to select one image, since you're trying to attach it to the form, not manage your collections of images.

Regarding the files not being deleted, I'll check on my end and will get back to you but this should definitely be the case when set to true.

ifox:
Hey @awgul, same here, closing to focus on fixing all related issues in #173. Thanks!

Issue #79, opened 2018-11-12
File Library upload fails


milewski:
Hello i have the following env vars:

```
MEDIA_LIBRARY_ENDPOINT_TYPE=local
MEDIA_LIBRARY_LOCAL_PATH=uploads/
MEDIA_LIBRARY_IMAGE_SERVICE=A17\Twill\Services\MediaLibrary\Local

FILE_LIBRARY_ENDPOINT_TYPE=local
FILE_LIBRARY_LOCAL_PATH=files/
FILE_LIBRARY_IMAGE_SERVICE=A17\Twill\Services\FileLibrary\Disk
```

and i get the following error:
![image](https://user-images.githubusercontent.com/2874967/48336563-8de95a80-e69b-11e8-8f82-4e56ded4565b.png)

i have also tried to override the file_library manually on twill.php 

```php
'file_library' => [
    'disk' => 'libraries',
    'endpoint_type' => 'local', ///<------
    'cascade_delete' => env('FILE_LIBRARY_CASCADE_DELETE', false),
    'local_path' => env('FILE_LIBRARY_LOCAL_PATH'),
    'file_service' => A17\Twill\Services\FileLibrary\Disk::class, ///<------
    'acl' => env('FILE_LIBRARY_ACL', 'public-read'),
    'filesize_limit' => env('FILE_LIBRARY_FILESIZE_LIMIT', 50),
    'allowed_extensions' => ['woff'],
],
```

but it somehow is still attempting to connect to S3...

Quickly workaround is to set these env var with dummy values:

```
S3_KEY=ANYTHING
S3_SECRET=ANYTHING
S3_BUCKET=ANYTHING
```

i guess the intended behavior is to work without these env vars if the endpoint is not set to S3.. 

**Second Problem:**

My model uses `HasFiles`, repository uses: `HandleFiles`, and my form.blade.php has

```php
@extends('twill::layouts.form')

@section('contentFields')

    @formField('files', [
        'name' => 'map',
        'label' => 'Map file',
        'note' => 'it needs to be the a proper obj file'
    ])

@stop
```

![image](https://user-images.githubusercontent.com/2874967/48337758-d22a2a00-e69e-11e8-9ce2-5a00a26ccd91.png)

i can pick a file and save it, however it is not persisted to the `fileables` table on db. am i missing some required settings on the model? or somewhere? there is no error and i can see the payload is being sent to the server "correctly": 

[...]

PR #173, opened 2019-01-25
Fix local media/file library issues

amrnn90:
Currently, File/Media Library stores the images under the public folder using a local path provided by the user config. However, when trying to delete the images (when using the cascade config) or when trying to get an uploaded file's size, it uses Laravel's Filesystem API (which uses a predefined storage location locally). 
This causes multiple issues such as #79 and #97 

This PR attempts to fix these issues by using Laravel's Filesystem API even when storing local files, and also providing disks configuration for both media/files depending on the user's chosen `endpoint_type` config.

Another issue that is being fixed is the cascade delete option when bulk deleting library assets, I did some refactoring in  **FileRepository**, **MediaRepository** and **ModuleRepository** (which should also solve an unrelated issue and make delete events trigger for other bulk deleted models).

Since we are now using Laravel's storage to store files locally, users would need to run `php artisan storage:link`.

And finally, the `local_path` config is now optional and is relative to the storage's public root path.

Please let me know if there are any issues or modifications that should be made, thanks.

### DIFF
--- a/config/disks.php
+++ b/config/disks.php
@@ -1,12 +1,32 @@
 <?php
 
+$localRootPrefix = storage_path('app/public/');
+$localUrlPrefix = env('APP_URL') . '/storage/';
+
+$mediaLocalConfig = [
+    'driver' => 'local',
+    'visibility' => 'public',
+    'root' => $localRootPrefix . trim(config('twill.media_library.local_path'), '/ '),
+    'url' => $localUrlPrefix . trim(config('twill.media_library.local_path'), '/ '),
+];
+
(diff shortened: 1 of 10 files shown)


ifox:
HEY @amrnn, 

Everything's looking super good and I'm eager to test this and merge it now that I'll have more time to review. My only concern is that we probably would consider those changes breaking, but I'll confirm once I get my hands on it!

Thank you so much for contributing.

amirakbulut:
Still waiting for this to contribution to be accepted..

Issue #136, opened 2019-01-07
Repeater always create new row and delete old one.


yanhao-li:
While using Repeater in the contentField as static content, the repeater's associate module always behave as creating new row and deleting the old one, even when actually update it.

From the best of my knowledge, there may have a slip in the updateRepeater function.

```php
    public function updateRepeater($object, $fields, $relation, $model = null)
    {
        $relationFields = $fields['repeaters'][$relation] ?? [];

        $relationRepository = $this->getModelRepository($relation, $model);

        // if no relation field submitted, soft deletes all associated rows
        if (!$relationFields) {
            $relationRepository->updateBasic(null, [
                'deleted_at' => Carbon::now(),
            ], [
                $this->model->getForeignKey() => $object->id,
            ]);
        }

        // keep a list of updated and new rows to delete (soft delete?) old rows that were deleted from the frontend
        $currentIdList = [];

        foreach ($relationFields as $index => $relationField) {
            $relationField['position'] = $index + 1;
            if (isset($relationField['id']) && starts_with($relationField['id'], $relation)) {
                // row already exists, let's update
                $id = str_replace($relation . '-', '', $relationField['id']);
                $relationRepository->update($id, $relationField);
                $currentIdList[] = $id;
            } else {
                // new row, let's attach to our object and create
                $relationField[$this->model->getForeignKey()] = $object->id;
                unset($relationField['id']);
                $newRelation = $relationRepository->create($relationField);
                $currentIdList[] = $newRelation['id'];
            }
        }

        foreach ($object->$relation->pluck('id') as $id) {
            if (!in_array($id, $currentIdList)) {
                $relationRepository->updateBasic(null, [
                    'deleted_at' => Carbon::now(),
                ], [
                    'id' => $id,
                ]);
            }
        }
    }
```

The below line should always return false if I'm thinking it right. 
```
    if (isset($relationField['id']) && starts_with($relationField['id'], $relation)) {
``` 

For the reason that the `$relationField` will be passed from $formRequest directly, `function getFormFieldsForRepeater` will not be called, thus the id will always be pure integer rather than the `$relation . '-' . $relationItem->id` format.

Can anyone else using Repeaters in their cases confirm this? Thanks in advance!

zipavlin:
@yanhao-li Yeah, this is in fact happening! Did you find a fix for this?

ifox:
I'm looking into this @yanhao-li @zipavlin, thanks for reporting.

amrnn:
[...]

PR #258, opened 2019-05-06
Fixed a issue that may cause repeaters unable to be updated

yanhao-li:
This commit is to fix the issue adverted in https://github.com/area17/twill/issues/136

### DIFF
--- a/frontend/js/utils/getFormData.js
+++ b/frontend/js/utils/getFormData.js
@@ -83,7 +83,7 @@ export const gatherRepeaters = (rootState) => {
         delete repeaterBlock.type
 
         // and lastly we want to keep the id to update existing items
-        repeaterBlock.id = repeaterItem.id
+        fields.id = repeaterItem.id
 
         return Object.assign(repeaterBlock, fields)
       })


frontend/js/utils/getFormData.js
```diff
@@ -83,7 +83,7 @@ export const gatherRepeaters = (rootState) => {
         delete repeaterBlock.type
 
         // and lastly we want to keep the id to update existing items
-        repeaterBlock.id = repeaterItem.id
+        fields.id = repeaterItem.id
```
ifox: @yanhao-li what about removing this line completely?  `repeaterBlock` should have been assigned `repeaterItem.id` in `id` by `buildBlock` already, but I might be missing something! Can you try it out? Thanks!

yanhao-li: @ifox You're right, the `repeaterBlock` indeed has already been assigned `repeaterItem.id` by `buildBlock`, the tricky part here is, on line 88
```php
        return Object.assign(repeaterBlock, fields)
```

Where `fields` is from:
```php
        let fields = repeaterBlock.content
```

However, for repeaters used on static form, the `fields` will include the related entity's id from database, and it override the `id` of `repeaterBlock`.

I'm not sure if there is any specific reason that you decided to keep the `id` while building block, if not, we could choose to get rid of the `id` from `fields` instead.

The only potential problem is, if the repeater state hasn't been persisted, it will still delete old block and create a same new one, because this is not a idempotent operation, we can solve it in the future by providing a more detailed http response.

but for now, this is the easiest way I can see to solve this specific issue. I have tested the updated repeater both on static form or inside block editor, didn't find any problems.

ifox: You're totally right, thank you so much for the investigation. I'm perfectly fine with merging this fix like this. We can improve in the future when refactoring, but we really need to get this fix in for now!

PR #259, opened 2019-05-07
Changed function names to $this->functionname

brammittendorff:
We should use `$this->` for local class function calling.


ifox:
Good catch @brammittendorff. To be honest this is pretty much dead code at the moment in Twill because our uploader currently doesn't make use of chunked requests. So we might clean things up at some point but still makes sense to merge this in the meantime. Thank you!

PR #261, opened 2019-05-08
Changed & for ? in parameter from thumbnail url

brammittendorff:
When using images from amazon, or trough varnish, or local images with `&no-cache` will resolve in a xml's or 404 not found request:

```
https://s3-eu-west-1.amazonaws.com/tpd/logos-domain/4a3199790000640005044e8e/255x0.png&no-cache
```

```
https://s3-eu-west-1.amazonaws.com/tpd/logos-domain/4a3199790000640005044e8e/255x0.png?no-cache
```

This will result in errors in you're console:

aaaa

Found by @dravenshorst

### DIFF
--- a/frontend/js/components/MediaField.vue
+++ b/frontend/js/components/MediaField.vue
@@ -462,7 +462,7 @@
 
           // try to load the media thumbnail
           let append = '?';
-          if (this.media.thumbnai.indexOf('?') > -1) {
+          if (this.media.thumbnail.indexOf('?') > -1) {
             append = '&';
           }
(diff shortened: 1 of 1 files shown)


ifox:
Hi @brammittendorff, makes sense. Imgix urls coming out of Twill always already have at least one parameter. So here it would be best to dynamically add a question mark or an ampersand depending on the existence of a query string in the url.

brammittendorff:
@ifox Fixed it. Here is a fiddle where you can see it working: https://jsfiddle.net/ojty67gb/2/

PR #264, opened 2019-05-13
migrations break due different type of column

fredylg:
migrations break due different type of column for foreign index


ifox:
Hi @fredylg , thank you so much for contributing to Twill!

Can you elaborate? On which database are you facing this issue?

filipac:
I have the same error when migrating, foreign key errors.

ifox:
Hi there, 

this has been fixed in https://github.com/area17/twill/commit/b3fd5819b0cde430c7cda3543d4242927892de05! Going forward Twill will default to big integers for primary and foreign keys, like Laravel since 5.8. There is an option provided to keep using regular integers for backwards compatibility.

Thank you for your contribution!

PR #277, opened 2019-05-27
Detect custom types starting with "file"

tomascardenal:
This allows passing custom media library types that are still handled as file list instead of media grid fallback when type is not "file" exactly while still maintaining compatibility.

PR #279, opened 2019-05-27
Add param to pass the type from PHP into Vue component

IonAgorria:
If other than default type is desired to be changed in files/medias form fields the type parameter must be passed into component blade.


ifox:
Hi @IonAgorria, thank you so much for contributing to Twill!

Is this related to #277? Is the idea being able to implement custom tabs in the media library that work with our medias and files form fields?

ifox:
See https://github.com/area17/twill/pull/277#issuecomment-523464051

Issue #193, opened 2019-02-28
getTranslationModelNameDefault not using configured namespace


jeepey:
Hi, 

Appreciate all the great work on this project.

I ran into an issue with the namespace of my Translations. Currently these are determined in the HasTranslation trait as follows:

```php
public function getTranslationModelNameDefault() 
{
        return "App\Models\Translations\\" . class_basename($this) . 'Translation';
}
```

Wouldn't it be better to use:

```php
public function getTranslationModelNameDefault() 
{
        return config('twill.namespace') . "\Models\Translations\\" . class_basename($this) . 'Translation';
}
```

Same thing for the revisions I guess.

Many thanks!

ifox:
Good catch @jeepey!

PR #280, opened 2019-05-27
Fix namespace

antonioribeiro:
Had two problems setting a new namespace for Twill (`App\Admin`):

1. It could not find some classes anymore, because they were hardcoded.
2. When making a new module it did not respect the new namespace path.

This PR fixes both.

### DIFF
--- a/src/Helpers/helpers.php
+++ b/src/Helpers/helpers.php
@@ -64,3 +64,27 @@ if (!function_exists('fireCmsEvent')) {
         Event::$method($eventName, [$eventName, $input]);
     }
 }
+
+if (!function_exists('twill_path')) {
+    function twill_path($path = '')
+    {
+        // Split to separate root namespace
+        preg_match('/(\w*)\W?(.*)/', config('twill.namespace'), $matches);
+
+        // Namespace App is unchanged in config?
+        if ($matches[0] === 'App') {
+            return app_path($path);
+        }
+
+        // If it it still starts with App, use the left part, otherwise use the whole namespace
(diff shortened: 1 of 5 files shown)


ifox:
Hi @antonioribeiro, thank you so much for this!

It fixes #193.

src/Helpers/helpers.php
```diff
@@ -63,4 +63,28 @@ function fireCmsEvent($eventName, $input = [])
...
+        // Split to separate root namespace
+        preg_match('/(\w*)\W?(.*)/', config('twill.namespace'), $matches);
+
+        // Namespace App is unchanged in config?
+        if ($matches[0] === 'App') {
+            return app_path($path);
+        }
+
+        // If it it still starts with App, use the left part, otherwise use the whole namespace
+        // This can be a problem for those using a completely different app path for the application
```
ifox: Everything LGTM @antonioribeiro, I'd just like to discuss your comment here. Can you elaborate?

antonioribeiro: Just an edge case for those not doing things the way most developers do these days, maybe. 

Let's say someone, insted of this 

	"psr-4": {
	    "App\\": "app/"
	},

Does this 

	"psr-4": {
	    "Area17\\App\\": "area17/app/"
	},

That namespace & path resolution may not work, since it takes in consideration everything starts with App or only the very first part of the namespace.

ifox: Right, makes sense! I think that's perfectly fine. We might want to add a note in the config file about it, but like you said, very edge case.

Issue #265, opened 2019-05-14
Twill adds Gate::before callback


zipavlin:
After 6 hours of solving and error where all my user gates failed immediately, I found that Twill registers before gate that checks if user is published - attribute I don't use on my not-twill users.

Is is possible to add gate only for twill users?

ifox:
Hi @zipavlin, thank you for reporting this and sorry for the delay in providing you with an answer!

I'm struggling to figure out if Laravel would allow us to register Gate callbacks for a specific auth guard. It doesn't seem possible though unless I'm missing something. I'd love to be able to do something like `Gate::guard('twill_users')->before...` to avoid those conflicts.

However, as I was researching this, @lorrenrules actually submitted a PR in #282! I need to review and test as I'm a bit worried that not relying on an actual Gate callback anymore could introduce some regression, but maybe not at all! @lorrenrules I'm sure you thought about this while making this change (looks great btw!), any concerns?

lorrenrules:
@ifox I also had trouble finding a way to separate the gate to specific guards. I agree that it doesn't seem possible at this time. I did my best to refactor it as identical to its current process as possible, keeping the same checks used in the before callback in each gate's callback. Definitely test it a bunch, as I am admittedly rather new to Laravel, and not fully confident there isn't something I could have overlooked. Hope it does well, and if not, I would be stoked to keep working at it with you.

ifox:
Merged #282 as part of Twill 1.2.2!

PR #282, opened 2019-05-31
Remove call on Gate before method, and refactor into gate definitions…

lorrenrules:
… to prevent failure on non-twill user authorization.


ifox:
Thank you so much for contributing @lorrenrules, please see my comment in https://github.com/area17/twill/issues/265#issuecomment-497768833.

PR #290, opened 2019-06-04
add option to prefix s3 uploads

### LINKED ISSUE #288: Possible to store files to S3 into a subfolder?
Hi on dev I use `MEDIA_LIBRARY_ENDPOINT_TYPE=local` and `MEDIA_LIBRARY_LOCAL_PATH=uploads/medias/` however when changing the `MEDIA_LIBRARY_ENDPOINT_TYPE` to `s3` it doesn't store the files under `uploads/medias/` is there an specific config to set the output dir for S3 endpoint?

milewski:
By default twill upload files to s3 in a flat manner, this PR adds an env var called: `S3_ROOT=` that is used to prefix the destination path:

before:

`4a38573e-04fe-40d2-892b-e0a42af55a61/image.png`

after:

```
S3_ROOT=uploads/
uploads/4a38573e-04fe-40d2-892b-e0a42af55a61/image.png
```

closes #288

### DIFF
--- a/config/disks.php
+++ b/config/disks.php
@@ -7,6 +7,7 @@ return [
         'secret' => env('S3_SECRET', env('AWS_SECRET', env('AWS_SECRET_ACCESS_KEY'))),
         'region' => env('S3_REGION', env('AWS_REGION', env('AWS_DEFAULT_REGION', 'us-east-1'))),
         'bucket' => env('S3_BUCKET', env('AWS_BUCKET')),
+        'root' => env('S3_ROOT', env('AWS_ROOT')),
         'use_https' => env('S3_UPLOADER_USE_HTTPS', env('S3_USE_HTTPS', env('AWS_USE_HTTPS', true))),
     ],
 ];
--- a/frontend/js/components/media-library/Uploader.vue
+++ b/frontend/js/components/media-library/Uploader.vue
@@ -151,7 +151,7 @@
       _onSubmitCallback (id, name) {
         this.$emit('clear')
         // each upload session will add upload files with original filenames in a folder named using a uuid
-        this.unique_folder_name = this.unique_folder_name || qq.getUniqueId()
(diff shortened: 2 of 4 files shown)


stiknoltz:
I've run into a couple issues that I think relate to this merge/commit.

### Issue 1

I believe this one is pretty straightforward and I'm happy to provide a pull request to solve if it looks good. 

The logic supplying the `unique_folder_name` value `area17/twill/frontend/js/components/media-library/Uploader.vue` on [line 154 in v1.2.2](https://github.com/area17/twill/blob/97cb4785980d5826ac3dce1f1b986228bb3d8c9b/frontend/js/components/media-library/Uploader.vue#L154) should be changed from 

```
this.unique_folder_name = this.uploaderConfig.endpointRoot + (this.unique_folder_name || qq.getUniqueId())
```

to 

```
this.unique_folder_name = (this.unique_folder_name || (this.uploaderConfig.endpointRoot + qq.getUniqueId()))
```

Otherwise, on multiple uploads, the `endpointRoot` value is continually re-appended to the `unique_folder_name` string.

### Issue 2

When a file is uploaded to S3 through the 'file_library' and using a custom `S3_ROOT`, the upload will fail due to the `size` check made in `area17/twill/src/Repositories/FileRepository.php`

This occurs because the `key` value that is assigned to `$path` for the `getSize` call is returned from the Uploader already containing the `S3_ROOT` value carried over via `endpointRoot`

The `getSize` method tests the file's existence first using `$this->assertPresent($path);` which hands it off to the `has` method on the `AwsS3Adapter` which prepends the `pathPrefix` [containing the `S3_ROOT` value]. So the `S3_ROOT` value is doubled at the beginning of the path string and the check fails. 

An easy solution to this issue is eluding me—as I'm not fully familiar w/ all the possible scenarios—but I'll continue to look at it.

stiknoltz:
Please disregard the above. I realized that Issue 1 is already solved in the HEAD of branch 1.2 and I've presented a possible solution for Issue 2 in #554

Issue #292, opened 2019-06-05
Undefined index: max


vacoup95:
>        'usp_repeater' => [
>                 'title' => 'USP',
>                 'trigger' => 'Add new',
>                 'component' => 'a17-block-usp_repeater',
>             ],

Block defined as above allows itself to be created in Twill. But once you save it and re-open it using the listview it will show the error: Undefined index: max

You can fix it by adding a 'max' value to the array, but sometimes you do not want to set a maximum.

amirakbulut:
Just set max to an extreme high integer.

yanhao-li:
Hi @vacoup95 @amirakbulut ,

I can confirm this is a bug, just created a PR to fix it: https://github.com/area17/twill/pull/297.

Thank you for reporting it!

PR #297, opened 2019-06-11
Fixed the Undefined index error reported in #292

yanhao-li:
This PR added a defensive check on the max index of blockTypeConfig to prevent the error `Undefined index max` from happening.


ifox:
Hi @yanhao-li! Thank you :)

Did you make sure to check how the frontend is reacting to not being provided with the `max` option?

yanhao-li:
@ifox 

Yes, when `max` is not provided, add new repeater item is simply not limited. Seems you guys have taken this into account when designed the component: https://github.com/area17/twill/blob/1.2/frontend/js/components/Repeater.vue#L71-L73

then missing the defensive check of `max` in HandleBlock should be a bug.

yanhao-li:
ref #292

Issue #295, opened 2019-06-11
PHPDoc - docblocks


stevanpavlovic:
As far as I could see, Twill does not have any docblocks in the codebase. What do you think about having those? We could contribute by providing some and make life easier for ones who want to go deeper into code. I would like to hear your opinions.

PR #301, opened 2019-06-12
 Add PHPDoc (docblocks) with param and return types to all PHP methods and classes

stevanpavlovic:
This is related to #295. I've added the docblocks to all methods and classes. What I am missing in most of the places are method descriptions. So what I did is that I only added annotations `@param` and `@return`. Adding those method descriptions is a huge procedure and I need a lot of time to do that since I would basically have to briefly explain the whole systems' functionality. I am not giving up on that, I am just postponing that.

But first I would really love that we merge this one. I would also like to hear from you guys about these changes.

### DIFF
--- a/migrations/create_files_tables.php
+++ b/migrations/create_files_tables.php
@@ -5,6 +5,11 @@ use Illuminate\Database\Schema\Blueprint;
 
 class CreateFilesTables extends Migration
 {
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
     public function up()
     {
         Schema::create('files', function (Blueprint $table) {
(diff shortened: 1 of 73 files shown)


yusufkandemir:
I think it's so much better to use type declarations instead of PHP doc blocks. It improves safety and decreases verbosity.

stevanpavlovic:
@yusufkandemir I agree, but for now, that could be really risky for the implementation. I wanted to prepare everything for doing strict type hinting because a lot of untested stuff could break to invalid type hinting. This way, we will have everything implicitly described so we can do the typehinting in a more "secure" way.

yusufkandemir:
@stevanpavlovic I agree with that. Refactoring is best when done progressively.

ifox:
Merged in 4f80c83d and reviewed in 6f4e9c92. Thank you so much @stevanpavlovic!

PR #308, opened 2019-06-22
Expanded the documentation section regarding nested modules with the …

raymondtri:
…function to enable browsing parent as well as child modules.

Without overriding the getBrowserItems function on the module controller, nested modules which are not top level parents are not accessible via the browser.

### DIFF
--- a/docs/.sections/crud-modules.md
+++ b/docs/.sections/crud-modules.md
@@ -88,7 +88,7 @@ Schema::create('table_name_singular_revisions', function (Blueprint $table) {
 // related content table, holds many to many association between 2 tables
 Schema::create('table_name_singular1_table_name_singular2', function (Blueprint $table) {
     createDefaultRelationshipTableFields($table, $table1NameSingular, $table2NameSingular)
-    // will add the following inscructions to your migration file 
+    // will add the following inscructions to your migration file
     // $table->integer("{$table1NameSingular}_id")->unsigned();
     // $table->foreign("{$table1NameSingular}_id")->references('id')->on($table1NamePlural)->onDelete('cascade');
     // $table->integer("{$table2NameSingular}_id")->unsigned();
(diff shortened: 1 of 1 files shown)


yanhao-li:
Hey @raymondtri ,

Thank you for contributing to Twill's documentation!

These days I'm also working on a project which is using the browser on nested module, however, everything works fine on my side. 
``
I'm just curious what kind of errors did you get when you not override the `getBrowserItems `?
The other thing that confused me is, I can't understand how your overriding could solve it, because it didn't change anything from my perspective.

This is your override method:
```
protected function getBrowserItems($scopes = [])
{
    return $this->repository->get(
        $this->indexWith,
        $scopes,
        $this->orderScope(),
        request('offset') ?? $this->perPage ?? 50,
        true
    );
}
```

The original ones are:
```
    protected function getBrowserItems($scopes = [])
    {
        return $this->getIndexItems($scopes, true);
    }

    protected function getIndexItems($scopes = [], $forcePagination = false)
    {
        return $this->transformIndexItems($this->repository->get(
            $this->indexWith,
            $scopes,
            $this->orderScope(),
            request('offset') ?? $this->perPage ?? 50,
            $forcePagination
        ));
    }

    protected function transformIndexItems($items)
    {
        return $items;
    }
```

It seems the overriding you made is just lift `getIndexItems` method one level up, it shouldn't bring any changes. Please correct me if I'm wrong.

Thanks!

raymondtri:
Hey @yanhao-li,

Sure no problem explaining it. Remember this only applies when the module being accessed via browser is nested (following existing Twill documentation). So when you are selecting something that has parents (following the documentation), you must skip the transformIndexItems step. In the documentation it says to add the following transformIndexItems to your controller:

`
protected function transformIndexItems($items)
    {
        return $items->toTree();
    }
`

This creates problems when trying to browse all instances of a model due to the fact that the returned items are now nested, and the browser component does not have any logic to handle that.

The long term solutions would be to either:

1. Add logic to the browser component so that it knows to check for nested modules
2. Add some sort of transformForBrowser method that is slated to run after the transformIndexItems method, to un-transform the transformIndexItems. (Seems loop-y).

But for now, my solution (skipping transformIndexItem) works as a way to prevent nesting of modules when the $items->toTree() method is called inside of the transformIndexItems. It also overrides the getBrowserItems in a way that is less loop-y than something like a transformForBrowser.

Best,
Ray

PR #341, opened 2019-08-06
Improve the module maker by letting user choose traits to include through questions

yanhao-li:
This PR proposed to add a `checkOption` method in `ModuleMake.php`, which will prompt choice questions while creating a new module, for user to decide which trait has to be brought in, instead of provide `-BTSMFPR` options manually.

The command now will work as the image shows below, if no options are provided:

image

It will only prompt for options that hasn't been explicitly defined in command, and the default answer is `false`:

image

### DIFF
--- a/src/Commands/ModuleMake.php
+++ b/src/Commands/ModuleMake.php
@@ -43,13 +43,13 @@ class ModuleMake extends Command
     {
         $moduleName = $this->argument('moduleName');
 
-        $blockable = $this->option('hasBlocks') ?? false;
-        $translatable = $this->option('hasTranslation') ?? false;
-        $sluggable = $this->option('hasSlug') ?? false;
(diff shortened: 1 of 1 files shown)


ifox:
Thank you @yanhao-li, that was a great idea!

While reviewing I made a couple of nice additions for 2.0:
- Generate model and migration content dynamically depending on provided options, removing comments, providing a greatly improved developer experience (this is at the price of ugly stubs for now – we should revisit our approach by using a proper PHP code generator in the future)
- Add `--all` option to enable all traits without any prompt
- When providing no option, prompt defaults to yes for all options
- When providing one or multiple options, prompt defaults to no for all other options
- It is possible to use artisan's `--no-interaction` option to skip the prompt

Merged in https://github.com/area17/twill/commit/c7b4bf22d0cc6cbc91ed7cb1d43d22dd9b94665c.

Issue #356, opened 2019-08-26
create unittests


Gummibeer:
Atm there is not a single test for anything in the package!?

This should be changed - it would make it easier for contributors to know if they break something and give all a better feeling to merge/release anything.

Gummibeer:
I would like to help but one of the repo admins has to enable StyleCI, TravisCI/CircleCI and all the great tools.^^
I would think that we should start with a simple setup and don't start with a 9x% coverage. This would allow to add unittests for upcoming bugs and also add the planned ones time after time.

I would like to help, and also offer to create the basic php unittests setup. But I can't help with frontend/vue.

Let me know if you want me doing it.

titantwentyone:
I would like to help with this also. Think it would be hugely beneficial to the project. I'm just starting out with Twill and have dived into the source so no expert (yet!) but will help if I can.

ifox:
Hi @Gummibeer! Sorry for the delay in getting back to you. We're working on setting all the tools up right now with the help of @antonioribeiro! Once this is ready, we will of course welcome community contributions to improve our test coverage and code quality.

Thank you so much for offering your help too @titantwentyone!

antonioribeiro:
I just sent a PR to provide the test environment and some of those tools @Gummibeer mentioned: 

https://github.com/area17/twill/pull/419

PR #357, opened 2019-08-26
upgrade translatable from dimsav to astrotomic

Gummibeer:
This PR upgrades from `dimsav/laravel-translatable` to `astrotomic/laravel-translatable`. This could be breaking because several older versions are dropped and the namespace changed.
So for everyone who used the package in his own code action is required.

https://github.com/Astrotomic/laravel-translatable/issues/10

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,21 +52,21 @@
     "php": "^7.0",
     "ext-json": "*",
     "ext-pdo": "*",
-    "laravel/framework": "~5.3.0|~5.4.0|~5.5.0|~5.6.0|~5.7.0|~5.8.0",
-    "dimsav/laravel-translatable": "^6.0.1|^7.0.0|^8.0.0|^9.0.0|^10.0.0",
+    "astrotomic/laravel-translatable": "^11.3",
+    "barryvdh/laravel-debugbar": "^2.4|^3.1",
     "cartalyst/tags": "^3.0.0|^4.0.0|^5.0.0|^6.0.0|^7.0.0|^8.0.0",
     "doctrine/dbal": "^2.9",
-    "league/flysystem-aws-s3-v3": "^1.0.13",
-    "myclabs/php-enum": "^1.5.0",
-    "imgix/imgix-php": "^3.0.0",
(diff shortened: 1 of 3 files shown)


Gummibeer:
@ifox it's @Gummibeer 😉 like the 🍺

Gummibeer:
Hey,
I'm a big fan of [LTS is a lie](https://connortumbleson.com/2019/01/28/lts-is-a-lie). But yes, I know that several developers still depend on it.
So it's a big decision to make - for sure. At @spatie we go an even harder way and only support the latest version of laravel in every release - so atm only 5.8.
There's no particular reason like L5.5 isn't compatible (it should be) - it's just a simple decision how many versions to support and test with. Because we build every supported version in a PR.
https://github.com/Astrotomic/laravel-translatable/blob/c7d3eefe7bd8390fa325b555583a127134c7ce9b/.circleci/config.yml#L186-L193

I could offer you to release a v11.4 which adds support for L5.5 back in.
On the other hand the feature support for L5.5 ends end of the month (Aug 30th, 2019 - https://laravel-news.com/laravel-release-process). So it would be a security fix only LTS after the next Twill release.
I'm a fan of forcing moving forward. It makes the life as maintainer a lot easier and is most time not a big problem. Primary because most Laravel releases aren't this breaking that it would need days/weeks to do. And even if - there's [Laravel Shift](https://laravelshift.com/shifts) at your side to do it.

But I can't decide for you which support strategy you want to go with. So just let me know your decision and as long as it doesn't require special work it would be fine for me to add L5.5. The moment it requires a single character more than L5.6 I will move it out of supported versions.

Gummibeer:
@ifox but anyway I worry **much much much much much much much** more about #356 ❗️
This would prevent me from using it in real/commercial projects - I have to really trust a new release and can't trust it if there's no automated testing with a high coverage which runs every commit/PR/merge and blocks a release if it fails.

yanhao-li:
Hey @Gummibeer 

I have included your changes in this PR: [Added support for Laravel 6](https://github.com/area17/twill/pull/389), closing this for now.

Thank you for the contribution!

Issue #365, opened 2019-09-07
Potential problem with migration file timestamp prefix


yanhao-li:
Line 271 in `TwillServiceProvider`
https://github.com/area17/twill/blob/9eaedc4b8911cff9eb168913ec7929ef596295c7/src/TwillServiceProvider.php#L271

1. `30000` should be `300000` => Missing a digit
2. When creating a new migration in the project with `php artisan make:migration` or `php artisan twill:module projects` , the newly generated migration file will have a smaller timestamp than 300000, (ex. 142916) it will cause the file migrated before the Twill's migration files in a fresh project, which will threw error because some table have foreign keys linked to the Twill's system tables (ex. twill_users)

---
Possible solution:
Change `300000` to `100000`, but need more thought and discussion.

yanhao-li:
Please see the discussion here: https://github.com/area17/twill/pull/372

PR #372, opened 2019-09-09
Fixed issue #365

yanhao-li:
This PR is designed to fix the issue https://github.com/area17/twill/issues/365
The timestamp used in migration file now will have 6 digits with leading zero, and avoid the problem described in the issue.

### DIFF
--- a/src/TwillServiceProvider.php
+++ b/src/TwillServiceProvider.php
@@ -262,13 +262,12 @@ class TwillServiceProvider extends ServiceProvider
     private function publishMigration($migration, $publishKey = null)
     {
         $files = new Filesystem;
-        $this->migrationsCounter += 1;
 
         if (!class_exists($migration)) {
             // Verify that migration doesn't exist
             $migration_file = database_path('migrations/*_' . snake_case($migration) . '.php');
             if (empty($files->glob($migration_file))) {
-                $timestamp = date('Y_m_d_', time()) . (30000 + $this->migrationsCounter);
+                $timestamp = date('Y_m_d_', time()) . sprintf('%06d', $this->migrationsCounter);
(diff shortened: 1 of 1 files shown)


ifox:
Hi @yanhaoli!

I did some research on this and concluded that the best approach is to switch to using `loadMigrationsFrom` provided by Laravel instead of publishing migrations during the install. In the same way developers currently can publish Blade views, they will still be able to publish migrations, but with `loadMigrationsFrom` we keep control of the order of our migrations instead of relying of something generated (like right now and in other proposals from the community) or fixed to a specific date for all of them (like this PR). Having the ability to date each new migration will allow us to decide wether a new migration is intended to run before the developer's own post Twill install migrations if we ever need to reach to that. New migration will be dated to the date we actually add them to Twill's codebase, but let's say our migration helpers are modified to add a foreign key referencing a new Twill table, then we would need that table to be created before the developer's migration are executed. 

Here's the catch though. Our migrations helpers are a tricky territory. We should probably never modify them if we want to avoid major pain in existing codebases. If we do modify them, we would have to provide a way to retroactively apply the changes to all tables that have been created with them in the past. If we do add features that requires adding columns or indexes to core module tables, we will create new helpers that will be able to be called individually in new migration files and from the existing helpers. Also, to avoid any duplicate operations we should always verify existence before executing queries in all our database migrations.

I can take care of the update but wanted to keep you in the loop and document the thinking for others.

ifox:
Fixed as described above in 7fced605 and ee489635.

Migrations are now loaded by the package and do not have to be published. Even when migrating from a Twill application that previously published migrations files, this change won't be breaking because Twill's migrations have been rewritten to check for the existence of tables or columns depending on the use case. A new option in `twill` config has been added to prevent Twill's migrations from loading automatically for those that would want to customize them: `twill.load_default_migrations`.

PR #374, opened 2019-09-10
Adding prefix to migration file

mcylinder:
Added a ’00’ to timestamp to ensure twill migrations run after default laravel migrations
`2014_10_12_100000_create_password_resets_table.php`

and before any new entity tables
`2019_09_10_133302_create_articles_table.php`

Directory sorts by string and moving a project had migration issues.  New tables were running before base Twill tables. (for example)
`2019_09_10_003001_create_tags_tables.php`
replaces
`2019_09_10_3001_create_tags_tables.php`


ifox:
Thank you so much for your contribution @mcylinder. After giving this more thoughts, we are going to move forward with the approach Laravel itself is using as seen in #372.

PR #377, opened 2019-09-11
Parse twill version from composer's installed.json

ghobaty:
With the latest update to `v1.2.2`, I believe you might have forgotten to bump up the version value in VERSION file causing the attribution in the footer to show an old version: `v1.2.1`

With this PR, I am proposing a different way to get the installed twill version: via composer `installed.json` file which should always yield the correct version, and if something goes wrong, it defaults to the value from `VERSION` file

### DIFF
--- a/src/TwillServiceProvider.php
+++ b/src/TwillServiceProvider.php
@@ -90,7 +90,7 @@ class TwillServiceProvider extends ServiceProvider
             'blocks' => Block::class,
         ]);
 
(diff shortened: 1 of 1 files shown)


yanhao-li:
Hi @ghobaty ,

Inspired by your PR, we have decided to use a constant VERSION in `TwillServiceProvider` to record the current version. I have created a new PR here: https://github.com/area17/twill/pull/404

Thank you so much for your contribution!

PR #387, opened 2019-09-16
make getLanguageLabelFromLocaleCode to use php intl's "Locale::getDisplayLanguage"

ghobaty:


ghobaty:
The only head-up I can think of that this requires intl extension

ifox:
Thank you @ghobaty!

I think the `intl` extension should be bundled in most PHP installations, but there's no harm in being safe by keeping our existing mapping like you did.

This change should also automatically use the current locale to display a language, which will be great in combination with #405 :)

PR #389, opened 2019-09-18
Added support for Laravel 6

yanhao-li:
### Changelog
- Added support for Laravel 6
- Superseded the all the `str_` and `array_` helpers with `Illuminate\Support\Str` and `Illuminate\Support\Arr` classes. https://laravel.com/docs/6.x/upgrade#helpers
- Updated some places using `url()` helper to avoid potential url generation problem. 
- Updated the dependencies to support Laravel 6
- Removed Laravel debugbar & Inspector
- Replaced Translatable with it's new repository https://github.com/dimsav/laravel-translatable


src/TwillServiceProvider.php
```diff
@@ -136,8 +132,6 @@ private function registerAliases()
 
         if (config('twill.debug.use_inspector', false)) {
```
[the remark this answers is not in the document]
yanhao-li: @ifox looks like I accidentally added it back during the rebasing. Has amended the commit and force-pushed.

yanhao-li:
Added a new commit to remove the deprecated code from codebase that was used to add backward compatibility for laravel 53 and 54.

src/Exceptions/Handler.php
```diff
@@ -116,28 +117,14 @@ public function render($request, Exception $e)
...
             return $this->handleUnauthenticated($request, $e);
         } elseif ($e instanceof ValidationException) {
             return $this->convertValidationExceptionToResponse($e, $request);
         }
 
-        if ($this->config->get('app.debug', false) && $this->config->get('twill.debug.use_inspector', false)) {
-            return Inspector::renderException($e);
-        }
-
         if ($this->config->get('app.debug', false) && $this->config->get('twill.debug.use_whoops', false)) {
```
ifox: The `twill.debug.use_whoops` configuration key and other Whoops relation keys can be removed as well which means we can remove this and the `renderExceptionWithWhoops` method.

yanhao-li: ok, done.

ifox:
Hi @yanhao-li, one very last thing before I can merge this: removing the `ddd` helper, now that Ignition comes with a `ddd` helper and that ours is not relevant anymore.

https://flareapp.io/blog/1-introducing-ddd-a-new-global-helper-for-laravel

PR #392, opened 2019-09-22
Add Once Option to Create Super Admin Command on Twill Install

droplister:
This is a very low priority idea for the `twill:install` command. A few of the operations that run check for file existence and only run if a file does not exist, but no existence check is performed for super admin accounts. This change makes this command's results slightly more consistent, in the instance where the command has been run already and someone has run it again.

So, if you were to run `twill:install` twice, for whatever reason, you would currently get...

```
Publishing complete.
Nothing to migrate.
Publishing complete.
Copied Directory [/vendor/area17/twill/dist] To [/public]
Publishing complete.
Let's create a superadmin account!

 Enter an email:
 >

```

Which feels inconsistent to me. This change would make it report...

```
Publishing complete.
Nothing to migrate.
Publishing complete.
Copied Directory [/vendor/area17/twill/dist] To [/public]
Publishing complete.
A super admin account exists.
```

More of a draft of an idea than anything else, thanks!

### DIFF
--- a/src/Commands/CreateSuperAdmin.php
+++ b/src/Commands/CreateSuperAdmin.php
@@ -14,7 +14,7 @@ class CreateSuperAdmin extends Command
      *
      * @var string
      */
-    protected $signature = 'twill:superadmin';
+    protected $signature = 'twill:superadmin {--once}';
 
     /**
      * The console command description.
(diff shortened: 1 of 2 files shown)


droplister:
Emphasis on "draft" because of the number of ways you could do this. Like, you could of course add this check inside of the twill:install command instead and it have it more self-contained, etc.

ifox:
Hi @droplister!

Thank you so much for your contribution. I understand the intention however I think I would prefer the approach your identified in your last comment, as I think the command to create superadmin users shouldn't deal with this aspect. An approach that would work best for everyone here would be to check for the existence of a superadmin user during the Twill install command before initiating the process of creating one. If a superadmin user already exists, the install command could first ask the user if creating a new superadmin user is desired or not.

PR #397, opened 2019-09-24
Support Glide Presets

yanhao-li:
This PR add the support for Glide presets by setting `glide.presets` attribute in `config/twill.php`.
Verified it works well.


src/Services/MediaLibrary/Glide.php
```diff
@@ -63,6 +63,7 @@ public function __construct(Config $config, Application $app, Request $request)
             'cache' => $this->config->get('twill.glide.cache'),
             'cache_path_prefix' => $this->config->get('twill.glide.cache_path_prefix'),
             'base_url' => $baseUrl,
+            'presets' => $this->config->get('twill.glide.presets')
```
ifox:
```suggestion
            'presets' => $this->config->get('twill.glide.presets', [])
```

Suggesting this to avoid any issues for users defining the `glide` key in their Twill config but not defining the `presets` key. In the current setup this would never be a problem because of how configurations are merged but down the line I feel like it's safer to set a default from here too.

Issue #162, opened 2019-01-20
Group fields in JSON column


nsivertsen:
Hi there!

I was wondering whether there's already some infrastructure in place to group a set of fields in a module in such a way that they are saved to one JSON column in the database, both for fields on the module itself and for fields in blocks?

The reason for doing this is that I'm using the connected fields partial to show or hide some fields based on other fields and I'd prefer not to have a database column for every possible field. Also, I can foresee adding more fields to this field group later on.

I've managed to save a set of fields to one JSON field in the database using the `prepareFieldsBeforeSave` hook the the module repository, but I'm having trouble getting them back into the form. If I add them to the `$fields` array in the `getFormFields` hook they are ignored.

Is there maybe a better way?

Thank you!

zipavlin:
@nsivertsen this is a great idea! Please share if you will manage to get it to work.

It would be a step towards having page/field templates (#126)

ifox:
Hi @nsivertsen,

Like I just mentioned in #164, this is a great idea.

We do not have the infrastructure immediately in place to be able to do that, however our blocks content works exactly like that so we should be able to leverage that part of the codebase to implement something at the module level.

In the meantime that's definitely something that should be achievable through the repository.
Which fields were you using? `input`, `wysiwyg`, `select`, `radios`, `multi_select`, `checkbox` and `checkboxes` should be compatible when added manually to `$fields` in `getFormFields`. 

Another approach that should work for more field types would be to deconstruct the JSON into multiple Eloquent model accessors, but that would be way to much infrastructure to solve an issue that should make things simpler.

yanhao-li:
If you are trying to render the `select` formFields use `getFormFields`, it might related to https://github.com/area17/twill/issues/175

PR #410, opened 2019-10-08
Group Fields in JSON column

yanhao-li:
This PR is intended to provide a solution per discussion in https://github.com/area17/twill/issues/162.

A new trait `HandleFieldsGroup` was added, users have to create new json columns manually on the table, and for each column, the fields that needed to saved as group should be declared as `$fieldsGroups` in the repository, as the below example demonstrated.

```php
class ProjectRepository extends ModuleRepository
{
    protected $fieldsGroups = [
        'fields_group' => [
            'field1',
            'field2',
            'field3'
        ]
    ];
}
```

### Supported Fields:
- `input`
- `wysiwyg`
- `date_picker`
- `select`
- `multi_select`
- `map` (haven't tested, but should work)
- `color`
- `checkbox`
- `checkboxes`

### Further improvements
- Support Browsers and Repeaters
- Added the `fieldsGroups` attribute to `repository.stub`
- Added it as a new option on ModuleMake CLI to add the json column automatically

Any comments/thoughts are welcomed!

### DIFF
--- a/src/Repositories/Behaviors/HandleFieldsGroups.php
+++ b/src/Repositories/Behaviors/HandleFieldsGroups.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace A17\Twill\Repositories\Behaviors;
+
+use Illuminate\Support\Arr;
+
+trait HandleFieldsGroups
+{
+
+    /**
+     * @param \A17\Twill\Models\Model|null $object
+     * @param array $fields
+     * @return array
+     */
+    public function prepareFieldsBeforeSaveHandleFieldsGroups($object, $fields)
+    {
+        foreach ($this->fieldsGroups as $group => $groupFields) {
+            $fields[$group] = json_encode(Arr::only($fields, $groupFields));
+            Arr::forget($fields, $groupFields);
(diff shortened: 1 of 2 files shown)


src/Repositories/Behaviors/HandleFieldsGroups.php
```diff
@@ -0,0 +1,61 @@
...
+                    $object->setAttribute($field_name, $field_value);
+                }
+                $fields = array_merge($fields, $decoded_fields);
+            }
+        }
+
+        return $fields;
+    }
+
+    protected function HandleFieldsGroups($fields) {
```
ifox:
```suggestion
    protected function handleFieldsGroups($fields) {
```

yanhao-li:
Because of the design of `MultiSelected` component, the stored json data for `multi_select` and `checkboxes` must be casted into the form in order to be displayed on the form
```php
[
  ['id' => 'new york'],
  ['id' => 'london']
]
```
It's obviously not ideal, but may not in the scope of this PR, we should open another issue to improve the component.

PR #414, opened 2019-10-10
Define a default locale on table mediables

antonioribeiro:
In SQLite you have to set a default value for "NOT NULL" columns, otherwise you will, while migrating, receive a 

```
Migrating: 2019_10_10_30011_add_locale_column_to_twill_mediables

   Illuminate\Database\QueryException  : SQLSTATE[HY000]: General error: 1 Cannot add a NOT NULL column with default value NULL (SQL: alter table "mediables" add column "locale" varchar not null)
```


ifox:
Hi @antonioribeiro!

That makes total sense and is also actually important to existing application so that existing `mediables` records are updated to use the first locale provided in config during that migration.

PR #415, opened 2019-10-11
Add quotes when populating store from form field

zipavlin:
I ran into an error when using _select_ field in settings field.

Blade was passing string value without quotes to ```window.STORE.form.fields``` which failed because JS considers value as variable and was unable to load it (as it does not exists), resulting in select field not being populated with value saved in DB.

I came across this when using _select_ in _settings_ page only.


zipavlin:
Please disregard this PR, I branched from wrong branch.

PR #421, opened 2019-10-14
Display error when the blocks dir is not found

antonioribeiro:
Currently we get this error in terminal: 

image


antonioribeiro:
Those constants will be used in testing.

PR #424, opened 2019-10-15
Support azure storage

m4n1ok:


config/disks.php
```diff
@@ -27,7 +27,26 @@
...
 
+$azureConfig = [
+    'driver' => 'azure',
+    'account' => [
+        'key' => env('AZURE_ACCOUNT_KEY'),
+        'name' => env('AZURE_ACCOUNT_NAME'),
+    ],
+    'container' => env('AZURE_CONTAINER', 'public'),
+    'endpoint-suffix' => env('AZURE_ENDPOINT_SUFFIX', 'core.windows.net'),
+    'protocol' => env('AZURE_UPLOADER_USE_HTTPS', env('AZURE_USE_HTTPS', true)),
```
ifox: `protocol` feels weird being a Boolean value. Can we use `use_https` for consistency with the S3 driver?

src/Helpers/media_library_helpers.php
```diff
@@ -10,11 +10,23 @@
...
 }
 
+if (!function_exists('azureEndpoint')) {
+    /**
+     * @param string $disk
+     * @return string
+     */
+    function azureEndpoint($disk = 'libraries')
+    {
+        $scheme = config("filesystems.disks.{$disk}.protocol") ? 'https://' : '';
```
ifox: See previous comment about the `protocol` config key.

src/AzureBlobStorageServiceProvider.php
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    /**
+     * Register bindings in the container.
+     *
+     * @return void
+     */
+    public function register()
+    {
+        $this->mergeConfigFrom(__DIR__ . '/../config/disks.php', 'filesystems.disks');
```
ifox: This already happens in Twill main service provider: https://github.com/area17/twill/blob/7ca1b4bfef36995bb9ec5d3fda20170c5808a1bc/src/TwillServiceProvider.php#L213

Any reason this would be needed here?

src/Http/Controllers/Admin/MediaLibraryController.php
```diff
@@ -289,4 +303,21 @@ private function getExtraMetadatas()
...
         });
     }
+
+    /**
+     * @param $sasUrl
+     * @return JsonResponse
+     */
+    public function isValidSas($sasUrl)
+    {
+        return response($sasUrl);
```
ifox: The injected `responseFactory` should be preferred over the `response` helper, like in other methods returning responses in this class.

src/Http/Controllers/Admin/MediaLibraryController.php
```diff
@@ -289,4 +303,21 @@ private function getExtraMetadatas()
             return [$field['name'] => $fieldInRequest];
         });
     }
+
+    /**
+     * @param $sasUrl
+     * @return JsonResponse
+     */
+    public function isValidSas($sasUrl)
```
ifox: `isValidSas` is technically doing the exact same thing as `policyIsSigned`. Same comment for `isNotValidSas` and `policyIsNotValid`. I agree existing naming doesn't make sense for the Azure implementation because it is not about signing a policy, but we should probably think about renaming to account for both and avoid the code duplication. I think `uploadIsSigned` and `uploadIsNotValid` probably make sense to be more agnostic.

src/Http/ViewComposers/MediasUploaderConfig.php
```diff
@@ -50,9 +50,15 @@ public function compose(View $view)
 
         $mediasUploaderConfig = [
             'endpointType' => $endpointType,
-            'endpoint' => $endpointType === 'local' ? $this->urlGenerator->route('admin.media-library.medias.store') : s3Endpoint($libraryDisk),
+            'endpoint' => $endpointType === 'local'
```
ifox: This and other ternary should probably be refactored now that we have more than 2 drivers. This is fine for now as we will probably revisit it when migrating away from FineUploader.

config/disks.php
m4n1ok: Yes I agree.

src/AzureBlobStorageServiceProvider.php
m4n1ok: Any valable reason here 😅

src/Http/Controllers/Admin/MediaLibraryController.php
m4n1ok: @ifox, I agree with you. I did a refactor of `SignS3UploadListener` / `SignAzureUploadListener` to have only one agnostic interface `SignUploadListener`.
Let me know if new implementation is better.

src/Http/ViewComposers/MediasUploaderConfig.php
m4n1ok: Should we refactor those now ? 
What's your recommendations on that (helpers functions / switch case...) ?

src/Http/Controllers/Admin/MediaLibraryController.php
ifox: Perfect, thank you!

src/Http/ViewComposers/MediasUploaderConfig.php
ifox: All good for now. Once we refactor the uploader to not use FineUploader anymore, we'll revisit.

src/AzureBlobStorageServiceProvider.php
```diff
@@ -0,0 +1,42 @@
...
+     * Perform post-registration booting of services.
+     *
+     * @return void
+     */
+    public function boot()
+    {
+        Storage::extend('azure', function ($app, $config) {
+            $connectionString = sprintf(
+                'DefaultEndpointsProtocol=%s;AccountName=%s;AccountKey=%s;EndpointSuffix=%s',
+                isset($config['use_https']) ? 'http' : 'https',
```
ifox: This doesn't sound right anymore.
```suggestion
                ($config['use_https] ?? true) ? 'https' : 'http',
```

PR #425, opened 2019-10-15
Fix published being sent as string on json

antonioribeiro:
In some installations we might get this kind of JSON error sometimes:

image

It's caused by a string being converted to JSON, which is escaped by PHP to become a "proper" JSON:

``` php
json_encode("0")
```

The piece of code doing this is trying to convert a boolean to a JSON string, but depending on the database server, that boolean might not be a boolean, but a string, it's the case of SQLite, which doesn't have a boolean type:

``` javascript
window.STORE.publication = {
    published: {{ json_encode(isset($item) ? $item->published : false) }},
}
```

But in this case we don't even need to convert it to JSON, we can just render a string (true/false):

``` javascript
window.STORE.publication = {
    published: {{ isset($item) && $item->published ? 'true' : 'false' }},
}
```

And we should get the perfect JSON boolean:

image

### DIFF
--- a/src/Models/User.php
+++ b/src/Models/User.php
@@ -49,6 +49,8 @@ class User extends AuthenticatableContract
         ],
     ];
 
+    protected $casts = ['published' => 'boolean'];
+
     public function __construct(array $attributes = [])
     {
         $this->table = config('twill.users_table', 'twill_users');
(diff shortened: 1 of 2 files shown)


antonioribeiro:
I just pushed a second fix for the same problem:

``` php
protected $casts = ['published' => 'boolean'];
```

Which also works using the previous version of the code. But I think we should not json_encode() strings to make them "jsonable", that might be a problem in some other parts of the package. If what we are plotting is not an array, I think we might just print them instead of trying to convert, because PHP will try to escape it to make it safer, and that can cause more harm than good.

PR #428, opened 2019-10-17
Fix Laravel 6.3 installation

antonioribeiro:
This is what I was getting when trying to install on a fresh Laravel 6.3 app:

image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,7 +52,7 @@
     "php": "^7.0",
     "ext-json": "*",
     "ext-pdo": "*",
-    "laravel/framework": "~5.6.0|~5.7.0|~5.8.0|~6.0.0",
+    "laravel/framework": "~5.6.0|~5.7.0|~5.8.0|^6.3",
     "astrotomic/laravel-translatable": "^11.3",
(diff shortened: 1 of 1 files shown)


ifox:
@antonioribeiro should we allow installing with any version of Laravel above 6.0 instead? I feel like some users might try installing on a recent project they have, but before updating their Laravel version

antonioribeiro:
@ifox That's actually what it does, but everything >= 6.3 == 6.3.1 and 6.4 will be installed

ifox:
Yup, got it. My point is that someone might be on Laravel 6.1 when they first try to install Twill.

antonioribeiro:
> Yup, got it. My point is that someone might be on Laravel 6.1 when they first try to install Twill.

It's only a problem if the person sets its composer.json as

```
"laravel/framework": "6.1.*",
```

But this way they won't get any fixes and improvements, not sure if it makes sense while installing a new package like Twill.

If they leave it in the [default](https://github.com/laravel/laravel/blob/master/composer.json):

```
"laravel/framework": "^6.2",
```

Composer will update both Laravel (to 6.3.x) and install Twill.

PR #431, opened 2019-10-19
fixed the http & https issue from APP URL

imknight:
original example for Laravel on APP_URL is with Http, this always creates url conflict on twill

### DIFF
--- a/config/disks.php
+++ b/config/disks.php
@@ -1,7 +1,7 @@
 <?php
 
 $localRootPrefix = storage_path('app/public/');
-$localUrlPrefix = request()->getScheme() . '://' . env('APP_URL') . '/storage/';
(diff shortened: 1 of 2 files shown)


pnzz7:
I think with this solution you encounter a problem when using a reverse proxy. Most reverse proxies (like traefik or any cloud service) send a http request to the application which will result in an url beginning with the http scheme even though the user originally requested over https

trueChazza:
> I think with this solution you encounter a problem when using a reverse proxy. Most reverse proxies (like traefik or any cloud service) send a http request to the application which will result in an url beginning with the http scheme even though the user originally requested over https

Any way to fix this?

For context, I am running traefik which does result in all urls beginning with http.
Browser will not submit anything to insecure endpoints over secure connection.

imknight:
this has been merge, pull the latest and see if you still having problem?

Issue #329, opened 2019-07-24
Permanently delete soft deleted / trashed records 


vacoup95:
Because most customers that the company I work for works with need to be able to permanently delete records (for many reasons obviously) I would like to requrest this feature...

And also, is there a temp workaround, because I deliver my product thursday and now I found out Twill does not allow permanent deletion which I find really weird that nobody thought of this??

yanhao-li:
Hey @vacoup95, for now I would suggest adding a new function or overriding the `delete` function from `ModuleController` to do the force delete.

vacoup95:
@yanhao-li thanks, do you know when this feature will be added? I feel like everyone from Twill just disappeared :(

yanhao-li:
Hahaha that’s because these days our team are attending the laracon 2019, sorry for the inconvenience!
I think the feature itself is not hard to implement, but we have more things need to be considered, like how to maintain browser relationship after force deleting? How to deal with nested modules? I can totally understand this is a rigid demand, but these questions need to be took care of properly, and maybe our answers/solution can’t perfectly fit your use case. I will still encourage you to build your own solution for now, and we hear you, and will implement it ASAP. @vacoup95

PR #454, opened 2019-11-14
Permanently delete trashed records

sauron:
Ref. Issue: #329.
- Added ability to destroy a.k.a. hard delete each trashed record.
- Added ability to bulk destroy the selected trashed records.
- Added ability to enable/disable the destroy function. Enabled by default.


ifox:
Merged in a8addc71 and updated in 578da43d. Thank you @sauron!

PR #456, opened 2019-11-15
Update migrations_helpers to support Laravel 6.

shad0w-rep:
Use Str::random Facade instead of helper. Required for Laravel 6 support.
this was missed in the L6 pull request that got merged.


ifox:
Great catch, thank you!

PR #471, opened 2019-11-22
Add name and version to allow reuse of dependencies in Workspaces

antonioribeiro:
To fix those error messages while using Workspaces:

```
warning Missing name in workspace at "/Users/antoniocarlos/code/area17/flv/test-compile/vendor/area17/twill", ignoring.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "twill",
+  "name": "@area17/twill",
   "version": "2.0.0",
   "private": true,
   "scripts": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,4 +1,6 @@
 {
+  "name": "twill",
```
luislavena: Perhaps will be better to scope the package for possible future usage?

```suggestion
  "name": "@area17/twill",
```

antonioribeiro:
Still unsure about the version too. This version is supposed to be related to the JavaScript part of Twill, not sure if it must be the same version of the PHP package.

ifox:
This has been added as part of ae62d29400ee61e46ab83cf8faad14d39d6ab319. Thank you so much guys!

PR #484, opened 2019-12-10
Fix join() in Glide service

tchapi:
There is still a deprecated `join()` in the codebase that makes **Glide** choke in PHP 7.4. This PR hopeful fixes it 🎉


tchapi:
Any plan on fixing this ? This breaks the whole CMS on PHP 7.4 ...

tchapi:
🆒

PR #498, opened 2020-01-08
Fix deeper Twill namespaces

antonioribeiro:
While testing things with the ModuleMake command, I found some issues with namespace configuration:

- When using deeper namespaces, like `App\Twill\Admin`, files were not copied to the right directory.
- Directory creation was not recursive, so I got a bunch of `permission denied` on things like `app/Twill/Admin/Http/Requests/Admin/BlogRequest.php`
- And the namespace declaration (`namespace App\Twill\Admin\Models;`) coming from the stubs are now being changed to comply with the new namespace.

Created some helpers to improve readability and dry things a bit.


ifox:
Merged in d64a158. Thank you @antonioribeiro!

PR #502, opened 2020-01-14
Fix for settings table

farnsworthMars:
The primary key should have been a `bigIncrements` - I noticed when trying to install on a Laravel 6.x from dev-master.


karl007:
better use (like at all other places)
```php
        Schema::create('settings', function (Blueprint $table) {
	    createDefaultTableFields($table, true, false);
            $table->string('key')->nullable()->index();
            $table->string('section')->nullable()->index();
        });
```
for be compatible with `config('twill.migrations_use_big_integers')`

PR #506, opened 2020-01-20
add required namespaces for Arr and Str in views and blade macros

karl007:


yanhao-li:
@karl007, Hi, thank you for the PR!

Since the blade template is in global namespace, I assume we don't need the fully qualified class name no? 

As I'm using the master branch every day and didn't run into any issues with it yet, could you please elaborate on the problem and use cases? 

On the other hand, you could also try registering the alias in `config/app.php`:
```
'aliases' => [
        'Str'       => Illuminate\Support\Str::class,
]
```

Personally speaking, we will use the full class name in most cases, but for these two, if the short alias could be used, I will prefer avoiding the fully qualified class name.

Thank you!

karl007:
Hi yanhao-li.
I got the exceptions "class Str/Arr not found" when I installed twill on an existing Laravel 6.11 project.
I only got the dashboard running after fixing these namespace issues.

If this works without these fixes on your side, 
a) It depends on my project, maybe because it was upgrade from a 5.4 version and something like the alias is now default in 6.x and I missed it during the update process
b) you use a version below 6.11. and this works only below this

yanhao-li:
@karl007 
If you are upgrading from 5.4 and haven't added them to the aliases, that makes sense to see this exception. I believe by adding them should fix the issue.

I'm okay with merging this PR if more people are facing the same issues like this.

Thank you!

PR #512, opened 2020-01-24
Do tests detect syntax errors?

szepeviktor:


szepeviktor:
Not really:

- if the file with the syntax error loads while running unit test: phpunit fails with "Parse error"
- Scrutinizer waits for never-arriving coverage data

szepeviktor:
`jakub-onderka/php-parallel-lint` is the best choice for catching syntax errors and stop further testing.

antonioribeiro:
Awesome, @szepeviktor, thank you for raising this!

`php-parallel-lint` looks nice, but I have two concerns about it: 
- No commits in almost 1 year.
- Currently raising deprecations warnings on PHP 7.4.

Do you see an alternative?

For now I'm using PHP-CS-Fixer to lint it and tell Travis to stop on errors as you can see here: https://travis-ci.org/antonioribeiro/twill/jobs/641672781.

Still unsure about Scrutinizer, because Travis is supposed to stop on errors and never hit the instruction that sends the coverage to it anyway, is it supposed to upload a coverage for a failing build?

szepeviktor:
Okay, PHP-CS-Fixer seems to spot syntax errors.

Scrutinizer may seem to be a useful service, I see no real added value in it.

antonioribeiro:
Scrutinizer also helps spoting errors and smells on code. Maybe phpstan could do it at the same extent, but we better keep it until we fix all static analysis errors we are having before ditching Scrutinizer. And, yes, I like the code quality chart they give us anyway... :)

ifox:
Fixed in #517.

PR #513, opened 2020-01-24
Validate Composer config in CI

szepeviktor:
PHP Extension and Composer packages installation goes into "install"


szepeviktor:
It will **fail** as currently the lock file is out of date.

ifox:
Rebased on latest master in #533.

Issue #514, opened 2020-01-24
Introduce Larastan


szepeviktor:
Larastan reveals many problem hidden in the codebase.

Please remove PHPStan as a dependency and add `nunomaduro/larastan`

Somehow `'media-library' => false,` should be set in config/enabled.php before analysis.

There are problems.

szepeviktor:
e.g. in `src/Repositories/Behaviors/HandleBrowsers.php` we need

```php
use A17\Twill\Models\Behaviors\HasMedias;
```

antonioribeiro:
Just sent a PR to add the Larastan plugin, hoping you will help us fix all 1107 errors it's now pointing to our codebase :trollface:

PR #516, opened 2020-01-25
Add larastan plugin to phpstan

antonioribeiro:
And change `composer.json` to use `"major.minor"` versions of Laravel and some other packages. As we are on SemVer now, we can ask for `"^6.0"` and Composer will still give us `"6.13"`, unless the user asks for `"6.9"` on its own `composer.json` file.

I also bumped into some version clashes with Symfony 5, because specifying minor and patch versions may reduce the range Composer has to calculate dependencies for all packages involved.

Closes https://github.com/area17/twill/issues/514

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,26 +52,27 @@
     "php": "^7.1",
     "ext-json": "*",
     "ext-pdo": "*",
-    "laravel/framework": "~5.6.0|~5.7.0|~5.8.0|^6.3",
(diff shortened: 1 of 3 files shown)


szepeviktor:
Analysis should start on Level 0 and move up to Level max.

szepeviktor:
@antonioribeiro You could write a one-liner PHP script to load all files in migrations/

Issue #74, opened 2018-11-09
Failed to updated nested repeaters


milewski:
Hello i have a nested repeater

whenever i try to update it i get this error:

```
error: {type: "ErrorException", message: "Undefined index: type",…}
file: "/home/vendor/area17/twill/src/Repositories/BlockRepository.php"
line: 34
message: "Undefined index: type"
type: "ErrorException"
``` 

this is my current setup

**admin/regions/form.blade.php**

```php
@extends('twill::layouts.form')

@section('contentFields')
 
    @formField('repeater', ['type' => 'regions'])

@stop

```

**admin/blocks/region.blade.php**

```php
@formField('input', [
    'name' => 'name',
    'label' => 'Name'
])

@formField('repeater', ['type' => 'locations'])
```

any idea what is the cause of the error?

/// edit
i suspect that the nested repeater data is being sent as `blocks:{}...` instead of `repeaters: {}` as if i move the `$fields['repeaters'] = $fields['blocks'];` it seems to work properly but then the "HandleBlocks" trait stop working...

ifox:
Hi @milewski,

nested repeaters are not supported, but I do not see a technical limitation that would prevent us from implementing it. Design-wise it can quickly become messy though.

I guess this is worth adding a note to our docs (as well as improving repeaters documentation as a whole because I know it is a bit weak right now).

milewski:
hi @ifox 

Also `block_editor` is not supported inside repeaters?

```
@formField('block_editor', [
    'blocks' => ['quote']
])
```

this example works outside repeaters but inside i see like this screenshot: 

![image](https://user-images.githubusercontent.com/2874967/48297371-6058b180-e4e1-11e8-8461-1077c141f81f.png)

ifox:
Hi Rafael,

Block editor is not supported inside repeaters, but repeaters are supported in blocks.

I guess a matrix of fields compatibility would be very helpful in our docs! Sorry about the confusion.

PR #521, opened 2020-01-31
Added support for nested repeaters in block

yanhao-li:
This PR is to add the nested repeaters as discussed in https://github.com/area17/twill/issues/74

I found out the fields passed to `afterSaveHandleBlocks` have included all information we needed to support nested repeaters, it's merely not be handled properly. 

This PR refactored the `HandleBlocks` trait a little bit by adding the following two methods:

- `createBlock` to create a block and recursively traverse it's child blocks and create them.
- `getChildBlocks` to recursively convert a child block's fields to the structure needed for creating a child block.

### DIFF
--- a/src/Repositories/Behaviors/HandleBlocks.php
+++ b/src/Repositories/Behaviors/HandleBlocks.php
@@ -70,16 +70,30 @@ trait HandleBlocks
         $blockRepository = app(BlockRepository::class);
 
         $blockRepository->bulkDelete($object->blocks()->pluck('id')->toArray());
-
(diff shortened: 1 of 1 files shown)


ifox:
Merging as this adds a new feature without breaking changes but noting that [`BlockController@preview`](https://github.com/yanhao-li/twill/blob/fix/support-nested-repeaters/src/Http/Controllers/Admin/BlocksController.php#L25) and [`HandleBlocks@hydrateHandleBlocks`](https://github.com/yanhao-li/twill/blob/fix/support-nested-repeaters/src/Repositories/Behaviors/HandleBlocks.php#L17) will also need to be updated for blocks and pages to render nested repeaters content during previews.

PR #528, opened 2020-02-06
Added an extra prop to repeater field to override the name

yanhao-li:
Added an extra prop `name` to repeater field used for overriding the repeater name.

### DIFF
--- a/views/partials/form/_repeater.blade.php
+++ b/views/partials/form/_repeater.blade.php
@@ -1,22 +1,26 @@
+@php
+    $name = $name ?? $type;
+@endphp
+
 <a17-repeater
     type="{{ $type }}"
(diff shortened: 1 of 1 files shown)


philippdaun:
@yanhao-li Just curios, what's the use case for that?

yanhao-li:
Hi @philippdaun,
Thanks for rasing this up, as I realize my PR description do miss a lot of details/background.

One of my projects requires two or more same repeater on the form but serves for different purposes. Currently, I have to duplicate the repeater setting in `config/twill.php` to achieve this, which is a little bit redundant, so the extra `name` prop is served for that purpose. 

I'm also thinking about adding more props to the repeater to override the settings. (like `max`, `component`, `trigger`), all those props shall fallback to the setting, but will make the field more customizable.

philippdaun:
@yanhao-li Makes perfect sense!

PR #529, opened 2020-02-07
Use real bold font in wysiwyg editor

philippdaun:
This change improves editor productivity by rendering bold passages in Inter's bold style instead of medium. Currently, bold sections in wysiwyg editors are barely detectable, making text hard to scan. See screenshots below for a before/after comparison.

**Before**:

Screenshot 2020-02-07 at 15 50 23

**After**:

Screenshot 2020-02-07 at 15 49 53

I've made the preload hint in `head.blade.php` optional, since it makes sense to only load the bold style if there's an actual wysiwyg field in the current edit form.

### DIFF
--- a/frontend/js/components/Wysiwyg.vue
+++ b/frontend/js/components/Wysiwyg.vue
@@ -336,6 +336,10 @@
         font-weight:700;
       }
 
+      b, p b, p strong, strong {
+        font-weight:700;
+      }
+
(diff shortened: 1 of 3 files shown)


ifox:
Rebased and merged in bbb1bb99d5f8599666a63ad212a47648a2f5a565. Thank you so much for this @philippdaun, it is indeed much better now!

Issue #184, opened 2019-02-12
Translation of Title


Equruler:
Just a simple question. If the module has translation, can the title be "not translatable"?

sauron:
Hey @Equruler,
When you create a new CMS module by using the ```twill:module``` command, you'll find a commented piece of code in the Model as:
```php
    // uncomment and modify this as needed if you use the HasTranslation trait
    // public $translatedAttributes = [
    //     'title',
    //     'description',
    //     'active',
    // ];
```
As the instruction states you only need to uncomment that snippet and remove the *title* from that array.

Keep in mind that the migration should reflect this change, e.g.:
in the migration you'll have to modify the creation of the translation table from
```php
        Schema::create('<your_model>_translations', function (Blueprint $table) {
            createDefaultTranslationsTableFields($table, 'setting');
            $table->text('title')->nullable();
            $table->text('description')->nullable();
            ...
        });
```
to
```php
        Schema::create('<your_model>_translations', function (Blueprint $table) {
            createDefaultTranslationsTableFields($table, 'setting');

            $table->text('description')->nullable();
            ...
        });
```
by moving the ```$table->text('title')->nullable();``` from the translation table to the main table.

I hope this helps.

ifox:
Hi @Equruler!

Thanks for answering here @sauron :)

This only part I think is missing is that when a module is translatable the views of that module need to know that the title is not to be translated by using `"translateTitle" => false` from formData or the Blade `@extend` directive in the module's `form.blade.php` file.

ifox:
#532 which was merged earlier today makes this automatically handled for you, check out @antonioribeiro's PR description.

PR #532, opened 2020-02-09
Feature/disable translations

antonioribeiro:
This PR is doing three different things:

### Automatically disable translations for a module when:

1. The repository trait for translations is not present (this was the original behaviour)
2. The model trait for translations is not present (I first tried to remove it and got an error because I had to do more to disable translations)
3. If the model's `$translatedAttributes` is empty, it probably means that nothing is translatable on that module. If you don't have the attribute on it and put it on fillable (as it is on the main module table and not on the translation) the application will raise an exception because it tries to save the attribute as an array (of translations), and something like that is only allowed for json columns.

### Automatically disable translations for the title

If the title column is not listed  in `$translatedAttributes`, it's probably not translatable, right? It's basically automating what we see on https://github.com/area17/twill/issues/184, no need to set `translateTitle` on your blade template anymore.

### Reduce maintenance repetition between `Model` and `TranslationModel`

Having to maintain the `fillable` array on the translation model and the `$translatedAttributes` on the base model looks like redundant (it may not be, but I would like to hear some use cases), so if the fillable is empty it will not try to fill it up with `$translatedAttributes`. Then I also changed the stub to create a `TranslationModel` like this, referring to te base model:

``` php
<?php

namespace App\Models\Translations;

use A17\Twill\Models\Model;
use App\Models\Photographer;

class PhotographerTranslation extends Model
{
    protected $baseModuleModel = Photographer::class;
}
```  

It's not a breaking change because it first tries to use the available `fillable` we already have on most of our modules today.

So now if you created a module with translations, but wants to do some playing without them, just change your migrations accordingly and comment the attributes that are not translated anymore, and all should work without issues. And if you remove/commend all attributes, translation for that module is entirely disabled.

I'm still not sure if we need to display the locale selector on the modal, when creating a model that has no translation for the title, but still has translation for any other (not shown on the modal) attribute:

image

I know that's for displaying/switching between the permalink translations, and I know we have the locale on the labels, but it may not be relevant for some users, and it can be confusing, like people switching to french and expecting to type a french translation for the name "Antonio Ribeiro", when they shouldn't.

Maybe we could (when having only one translatable field and not really relevant) move the language switcher to the label?:

image


ifox:
This has been reviewed and merged in https://github.com/area17/twill/commit/957702ac0ecd84e541293d7790e96d6200358412. Such a great change, thank you so much @antonioribeiro!

PR #540, opened 2020-02-12
Fix/get slug infinite requests

gilbo75:
Fix getSlug() function when locale fallback is activated that’s create an infinite requests


ifox:
Merged in da1865b7e5382b63e194933c335e616ed8ecba22. Thank you!

PR #543, opened 2020-02-19
Fix config key

antonioribeiro:
Key name was different from twill.php's.

Maybe we should default that to `false`, but maybe only for people upgrading from 1.2 (somehow)?

### DIFF
--- a/src/TwillServiceProvider.php
+++ b/src/TwillServiceProvider.php
@@ -213,7 +213,7 @@ class TwillServiceProvider extends ServiceProvider
 
     private function publishMigrations()
     {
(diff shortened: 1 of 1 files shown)


ifox:
This has been fixed in https://github.com/area17/twill/commit/41378f9dbf4e508ab3432c232106422bf54167ca. I'm open to discussing this today before the release, but with our migrations being very defensive, as in they check for tables and fields existence in your database before doing anything, I think we should keep it `true`. 2.0 has very few breaking changes, so I will make this a big callout in our upgrade guide.

PR #545, opened 2020-02-21
Allow firstOrCreate to have only attributes

antonioribeiro:
This will allow us to first or create models that only have, for example, a name. It's also more compatible with Eloquent's `firstCreate()`: https://github.com/laravel/framework/blob/9f0af8433eaae7a59e297229ca84d998e0e56edf/src/Illuminate/Database/Eloquent/Builder.php#L431


ifox:
Merged in https://github.com/area17/twill/commit/0141579230e6a08063a94f841180aaa30cc1fc70. Thank you so much @antonioribeiro!

PR #546, opened 2020-02-21
fix composer classname collisions with already published migrations

shad0w-rep:
if you already published the twill migrations in an existing twill project, you will get composer classname collisions.
`PHP Fatal error:  Cannot declare class AddLocaleColumnToTwillMediables, because the name is already in use`

this PR applies TwillDefault to the class and filename.


philippdaun:
Can we please have this merged? The bug this is supposed to fix is currently breaking deployments, requiring ugly hacks in production.

ifox:
Fixed in https://github.com/area17/twill/commit/74173acd0b750243c2d3bab382ed842f36efdce9.

Issue #549, opened 2020-02-25
Feature request: allow a filter option to act on a model on initial load of the index page


arkid:
This doesn't seem to be possible out of the box.   So it seems that when a user interacts with the filter drop down, they always have to apply something as default for the interface to work properly.  The assumption seems to be that you have no filter applied as default.   

I've found for a few models that would otherwise return way too many records that applying a filter as default is really useful.  

So when you define in the controller

protected $filters = [];

It would be great if there was something like a $filtersDefaultIndex that would let you specify either an index number or a string value (if you know what is going to be populated into your select list). 
 
Defining this would do 3 simple things: 

1) actively enable that filter by default when the page loads
2) update the 'filter' url parameter appropriately on initial loading of the index page
3) update the filter visual interface so that when a user first goes into it they see the option set as the default already selected there.

PR #551, opened 2020-02-25
Display and apply default filters in listings

philippdaun:
This does two things in the listing view:

1. Mark the currently selected filter in the `v-select` component on page load (so it applies after page refresh as well)
2. Add the ability to specify default filter values per filter

This is a simplified example of setting the default value for a filter — display upcoming events only unless a different value for the `state` filter is selected from the dropdown.

It becomes clear why marking the selected filter becomes necessary: setting a default filter requires marking the current filter on page load.

```php
class EventController extends ModuleController
{
    protected $filters = [
        'state' => 'eventState',
    ];

    protected $filtersDefaultOptions = [
        'state' => 'upcoming',
    ];

    protected function indexData($request)
    {
        return [
            'stateList' => collect([
                'upcoming' => 'Upcoming',
                'past' => 'Past',
                '_' => 'All',
            ]),
        ];
    }
}
```

This PR would close issue #549.


arkid:
Thanks @philippdaun this is great

ifox:
Thank you so much for this @philippdaun, it makes a lot of sense!

PR #556, opened 2020-02-26
Duplicate entities directly from listing

ferpetrelli:
RFC for entities duplication:

This is the first working version with an experimental approach on how to achieve content duplication.

1. Create a new item using the bare minimum attributes
2. Update this item using the latest revision from the original item
3. Trigger a notification and a redirect to the form page for the new element

Please feel free to shoot ideas and comments as this PR is still subject to change.

Requirements:

- Model should implement revisions
- Activate duplication on your Twill Controller for your model

`protected $indexOptions = [
        'duplicate' => true
    ];`

### DIFF
--- a/frontend/js/components/table/TableRow.vue
+++ b/frontend/js/components/table/TableRow.vue
@@ -11,7 +11,7 @@
     </td>
     <td class="tablecell tablecell--spacer">&nbsp;</td>
     <td class="tablecell tablecell--sticky">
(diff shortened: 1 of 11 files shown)


ifox:
Such a pristine and thorough contribution, thank you so much @ferpetrelli! This is going into Twill 2.0 and has been merged in master.

marinaglancy:
Hello! Thanks for this feature, there are couple of problems with it:
1. The "Duplicate" appears in the actions list even when indexOption 'duplicate' is set to false.
2. Default implementation of duplication does not work if the model does not have revisions.

at the same time the feature is great, thank you!

marinaglancy:
in fact I created an issue https://github.com/area17/twill/issues/673 , there is one more problem with duplicating "Trashed" entities

PR #566, opened 2020-03-04
Add portuguese (brazil and portugal) translations

antonioribeiro:


ifox:
Hmm, totally unrelated, but it seems like GitHub changed the way squash and merge works on pull requests. It used to keep authorship from the PR creator in the squashed commit, but in my last 2 merge, I'm the only author. I like to avoid the merge commits, but I might have to do that instead from now on. Not sure if it's a bug or a feature 🤔.

antonioribeiro:
If that's true, you will be the only contributor from now on: https://github.com/area17/twill/commits?author=antonioribeiro. 😆

But I'm wondering why you are rebasing/squashing a PR with only one commit? 
Shouldn't you just merge it?

PR #572, opened 2020-03-06
Fix fields not rendering after switching locales

LiquidPL:
Noticed this happening with WYSIWYG fields, like in here: https://streamable.com/hdjc9.

It turned out that the field attributes were shallowly copied, which caused them to be overwritten when rendering the field for subsequent locales.


ifox:
Good catch! This is a regression introduced by a new feature we added to Quill which made the toolbar options object deeper than it used to be.

PR #573, opened 2020-03-08
I18n improvements

LiquidPL:
* configured frontend libraries (date-fns, vue-timeago, flatpickr) to respect the user locale preferences,
* internationalized some more strings,
* added a Polish translation.


LiquidPL:
Now that I'm thinking about it, we should also add parameter replacing to the Vue `$trans` helper for parity with Laravel's, since I've seen a few strings that could make use of it.

CHANGELOG.md
```diff
@@ -43,7 +43,7 @@ All notable changes to `twill` will be documented in this file.
 - Rename default migrations to prevent conflicts in existing codebases [`74173acd`](https://github.com/area17/twill/commit/74173acd0b750243c2d3bab382ed842f36efdce9)
 
 #### Frontend build
-- *@antonin and @ifox*
```
yanhao-li, line 46: Let's remove this commit `c35a20346e240415d9589fceb14fa7df9165d6c6` from the pull request.

lang/en/lang.php
```diff
@@ -63,25 +117,39 @@
             'create-another' => 'Create and add another',
         ],
         'update' => [
-            'title' => 'Update',
+            'title' => 'Update item',
```
yanhao-li, line 120: I think we may prefer a concise name, and keep it clean and simple as `Update`

lang/en/lang.php
```diff
@@ -145,6 +257,16 @@
     ],
     'publisher' => [
         'switcher-title' => 'Status',
+        'languages' => 'Languages',
+        'langages-published' => 'Live',
```
yanhao-li, line 261: typo here `langages` => `languages`

frontend/js/utils/locale.js
```diff
@@ -0,0 +1,44 @@
...
+    'date-fns': require('date-fns/locale/fr'),
+    flatpickr: French
+  },
+  pl: {
+    'date-fns': require('date-fns/locale/pl'),
+    flatpickr: Polish
+  }
+}
+
+const twelveHourFormatLocales = ['en']
```
yanhao-li, line 28: This - I guess its currently fine, but I guess in the future we should add another prop to the `_date_picker` for user to override.

frontend/js/components/modals/ModalCreate.vue
```diff
@@ -53,7 +53,7 @@
         return this.createMode ? this.formCreate : this.action
       },
       modalTitle: function () {
-        return this.createMode ? this.$trans('modal.create.title', 'Add new') : this.$trans('modal.update.title', 'Update')
+        return this.createMode ? this.$trans('modal.create.title', 'Add new') : this.$trans('modal.update.title', 'Update item')
```
yanhao-li, line 56: let's keep the label of `Add new` and `Update` consistent.

yanhao-li:
Hi @LiquidPL ,

Since I'm working on a new approach which will impact the way how we contribute translations in the future, I have included your commits in my PR: #624

Let me know if you have any comments.

Thanks for your contribution!

PR #576, opened 2020-03-11
Fix date_picker and color form fields not working in settings

joyceverheije:


Stiropor:
This won't work properly when using translations (date field is not set to be translatable but Twill saves settings in all locales I see). I get:
`htmlspecialchars() expects parameter 1 to be string, array given (View: /Users/mbratkovic/Sites/dsrc/resources/views/vendor/twill/partials/form/_date_picker.blade.php) `

PR #589, opened 2020-03-14
Fix Blade component name spelling

philippdaun:


philippdaun:
The commit history is a bit wonky after merging 2.0 from upstream, sorry. Probably easier if you just correct the typo yourselves instead of merging this.

PR #642, opened 2020-05-03
Skip modal to create record

mazeeblanke:
This update adds an option for skipping the modal to create a record this allows developers support complex form fields when creating a record such as blocks, repeaters etc.

To use set `skipCreateModal` to true in the `$indexOption` of the admin Module Controller

       protected $indexOptions = [
          'skipCreateModal' => true
       ];
        

A new route is added to a module for showing the create form i.e `/module/create.` 
If a user visits `/module/create` when  `skipCreateModal` is set to false, the user is redirected to the listing page with the `openCreate=1`

### DIFF
--- a/frontend/js/components/TitleEditor.vue
+++ b/frontend/js/components/TitleEditor.vue
@@ -7,9 +7,10 @@
         </a>
         <span v-else>{{ customTitle ? customTitle : title }}</span>
       </h2>
-      <a v-if="permalink || customPermalink" :href="fullUrl" target="_blank" class="titleEditor__permalink f--small">
+      <a v-if="(permalink || customPermalink) && !showModal" :href="fullUrl" target="_blank" class="titleEditor__permalink f--small">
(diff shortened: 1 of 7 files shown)


arkid:
@ifox any chance this can make the next release?  It would really make Twill a lot more flexible to not have this initial popup forced at users always for certain kinds of data entry.

marinaglancy:
This change has broke settings forms. I found that if I add this to the `settings.blade.php` it works again:

`window['{{ config('twill.js_namespace') }}'].STORE.publication.createWithoutModal = false`

jrusinowicz:
I see another problem with this change. When I'm setting skipCreateModal to true, I'm redirected to /create route but on this route modal is opening and i'm not able to close it...

ifox:
You're not supposed to be able to until you provide a title. The difference is that instead of creating a record it closes the modal and let's you fill in the rest of the form before creating anything in database.

Issue #626, opened 2020-04-08
[Feature Request] Update to vue-select 3.x — Add the option to paste several tags at once


tchapi:
Hi

In the tags field, you cannot at the moment paste a list of comma-separated (or _anything_-separated) tags. The whole string is parsed as a single tag.

It may be opinionated but I think that tags do not usually contain spaces (or commas, FWIW), and thus the behaviour is rather unusual. When creating, say, an article and inputting tags from another source ('marketing' gave you a list of keywords/tags to put in this article), you want to be able to copy and paste it like you did for the content, the title, or any other field.

If Twill updates its dependency to `vue-select` > **3.3**, it should be possible to add this behavior quite easily by adding a handler to "separate" tags if the value submitted contains a comma (or `; `, or a space, or anything — that could be a parameter of the field tag)

This would **also** allow to tag by entering a comma or a space, which is generally the default behaviour on this kind of field in forms.

See https://github.com/sagalbot/vue-select/pull/971 for reference on the `handlers` method.
The changelog between 2.x and 3.x seems quite reasonable : https://vue-select.org/guide/upgrading.html#events-instead-of-callbacks

Would that be something interesting ? Happy to help if needed
Thanks !

ifox:
Planned for our next minor release with an update of #674, thanks for the suggestion @tchapi!

PR #674, opened 2020-06-08
[3.x] Update vueselect dependency to v3

mazeeblanke:
Update vue select to version 3.

1) Replaced the onSearch prop with the search event has using the same parameters in it's place.
2) import the css file for the vue-select


tchapi:
That would close https://github.com/area17/twill/issues/626 I think, that would be really nice !

mazeeblanke:
@ifox thank you for the feedback. I will go through it again and fix it

haringsrob:
I think that since we are on this we can also add support for the `selectable` https://vue-select.org/guide/selectable.html#selectable-prop option.

We could take a similar array as the `options`.

haringsrob:
Moving this onto 3.x

PR #685, opened 2020-06-27
Add documentation with some examples for the MultiBrowser feature

sauron:
Please use the attached image as screenshot.
multibrowser

### DIFF
--- a/docs/.sections/form-fields.md
+++ b/docs/.sections/form-fields.md
@@ -776,6 +776,137 @@ Browser fields can be used to save a `belongsToMany` relationship outside of the
 Checkout this [Spectrum tutorial](https://spectrum.chat/twill/tips-and-tricks/step-by-step-ii-creating-a-twill-app~37c36601-1198-4c53-857a-a2b47c6d11aa) until we update this section to get more info on setting things up.
 When using inside of the block editor, no migration is needed.
 
+### Browser with multiple endpoints / Browser for different entities / MultiBrowser
+![screenshot](/docs/_media/multibrowser.png)
+
+```php
+@formField('browser', [
+    'endpoints' => [
+        [
+            'label' => 'Projects',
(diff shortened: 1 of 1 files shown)


docs/.sections/form-fields.md
```diff
@@ -774,6 +774,137 @@ Browser fields can be used to save a `belongsToMany` relationship outside of the
...
 When using inside of the block editor, no migration is needed.
 
+### Browser with multiple endpoints / Browser for different entities / MultiBrowser
+![screenshot](/docs/_media/multibrowser.png)
+
+```php
+@formField('browser', [
+    'endpoints' => [
+        [
+            'label' => 'Projects',
```
splatEric, line 784: I've just been working on this in my twill experiments this week. A point that might be worth highlighting here is that the endpoint values depend on whether or not there is a base path for your admin route, and whether your module is part of a group (and therefore has its own prefix).

I used the following, which accounts for the admin base path, and I think makes clear whether adjustments for grouping to the module path is necessary:

```
'endpoints' => [
            [
                'label' => 'Blogs',
                'value' => moduleRoute('blogs', null, 'browser', [], false),
            ],
            [
                'label' => 'Roles',
                'value' => moduleRoute('roles', 'role', 'browser', [], false),
            ]
        ]
```

ifox: Hi @splatEric, totally!

The browser field actually has another feature that will do this for you if you provide a `modules` array: https://github.com/area17/twill/blob/9ead6d5c588243f2727906799e1d36c22d5c7c3c/views/partials/form/_browser.blade.php#L5-L12 

In your case, that would mean:
```
'modules' => [
    [
        'label' => 'Blogs',
        'name' => 'blogs',
    ],
    [
        'label' => 'Roles',
        'name' => 'roles',
        'routePrefix' => 'role',
    ],
]
```

PR #686, opened 2020-06-29
Select field - fix vue.js error 

### LINKED ISSUE #175: Use getFormFields to render select field will cause Vue.js error
For example, if there is a select field named with "select_option_1" on the form, and when I'm trying to  render the form using getFormFields as following

```php
    public function getFormFields($user) {
        $fields = parent::getFormFields($user);
        $fields["select_option_1"] = "option_value";
        return $fields;
    }
```

This will trigger the javascript error: 

> edit:337 Uncaught ReferenceError: option_value is not defined

I noticed that "option_value" doesn't be treated as a string type when it was being pushed in Vuex.

```javascript
    window.STORE.form.fields.push({
    name: 'select_option_1',
    value:  option_value       **// WRONG, should be "option_value"**
    })
```

I believe the problem is caused by the logic existed in `_selector_input_store.blade.php` file.

```php
window.STORE.form.fields.push({
    name: '{{ $name }}',
    value: @if(isset($item) && is_numeric($item->$name)) {{ $item->$name }}
           @elseif(isset($item->$name)) {!! json_encode($item->$name) !!}
           @elseif(isset($formFieldsValue))
                @if (is_array($formFieldsValue))
                    @php
                        $formFieldsValue = array_first($formFieldsValue, null, '');
                    @endphp
                @endif
                @if (is_numeric($formFieldsValue)) {{ $formFieldsValue }}
                @else {!! $formFieldsValue !!}
                @endif
           @else
            ''
           @endif
})
```

The $formFieldsValue has been unescaped in the end, is this a bug or a feature?

A temporary possible workaround for this issue might be:
    public function getFormFields($user) {
        $fields = parent::getFormFields($user);
        $fields['select_option_1'] = "'" . $option_value . "'";
        return $fields;
    }

Insperio:
Check if $formFieldsValue is a string, if so we escape it correctly. Didn't find any conflicts in my app but let me know. 
Fixes #175

### DIFF
--- a/views/partials/form/utils/_selector_input_store.blade.php
+++ b/views/partials/form/utils/_selector_input_store.blade.php
@@ -9,6 +9,7 @@ window['{{ config('twill.js_namespace') }}'].STORE.form.fields.push({
                     @endphp
                 @endif
                 @if (is_numeric($formFieldsValue)) {{ $formFieldsValue }}
+                @elseif(is_string($formFieldsValue)) '{{ $formFieldsValue }}'
                 @else {!! $formFieldsValue !!}
                 @endif
            @else


ifox:
Merged with #653. Thanks for your contribution!

PR #691, opened 2020-07-11
List all blocks including repeaters

talvbansal:
Fixes a bug where blocks wouldn't render previously stored repeater fields even though they were in the database


ifox:
Our tests are missing assertions about repeaters in blocks so I didn't catch this either, sorry about that. Thank you so much for your time.

talvbansal:
No problem at all, thanks for building such a fantastic product! I'm pretty new to using it but can see huge potential for it and I hope to be able to contribute new features and fixes along the way. If theres anything I can help out with please let me know i'd be more than happy to help!

Issue #693, opened 2020-07-13
[Bug?] saved non-empty "checkboxes" value caused "Invalid argument supplied for foreach()" on next edit


ResetButton:
I'm trying to implement "checkboxes" in module.

When I'm creating new module item all is fine, but after I save item and checking some checkboxes  - I get "Invalid argument supplied for foreach()" when I try to edit this page next time or just refresh it.

Steps to reproduce
1. create a module with checkboxes and add demo checkboxes to form (https://twill.io/docs/#checkboxes)
2, in admin area create new module item, then edit this item checking some checkboxes and save it., 
3. check database - you'll see something like ["finance","arts","civic"] in database record.
4. reopen page for editing or just refresh it.

ResetButton:
Problem is here
https://github.com/area17/twill/blob/2.x/views/partials/form/_checkboxes.blade.php
line 42 Arr::pluck($item->$name, 'id')
but value passed from database is like ['1','2', '3'] - not array

ResetButton:
Here is quickfix
replace line 42 with
`value: {!! json_encode(isset($item) && isset($item->$name) ? json_decode($item->$name) : $formFieldsValue) !!}`

antonioribeiro:
Are you casting it? And creating the "id" for it?

``` php
protected $casts = ['mission' => 'array'];

public function getMissionAttribute($value)
{
    return collect(json_decode($value, true))
        ->filter()
        ->mapWithKeys(fn($value) => [$value => ['id' => $value]])
        ->toArray();
}
```

ifox:
Fixed in https://github.com/area17/twill/commit/736a064262bafd4c47b0562136b8a2d859c00ca0.

PR #694, opened 2020-07-14
Fix issue #693

ResetButton:
Details are here https://github.com/area17/twill/issues/693

### DIFF
--- a/views/partials/form/_checkboxes.blade.php
+++ b/views/partials/form/_checkboxes.blade.php
@@ -39,7 +39,7 @@
 @push('vuexStore')
     window['{{ config('twill.js_namespace') }}'].STORE.form.fields.push({
         name: '{{ $name }}',
-        value: {!! json_encode(isset($item) && isset($item->$name) ? Arr::pluck($item->$name, 'id') : $formFieldsValue) !!}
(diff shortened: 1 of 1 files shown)


ifox:
Thank you so much for your contribution @ResetButton!

I'm afraid this change would be breaking depending on how this type of field has been managed in the module repository or model. Can you share how you are saving and retrieving that field?

ResetButton:
Absolutely similar to https://twill.io/docs/#checkboxes  I touched nothing more so module repository or model is default, I just tried to use diffent MSQL field types - JSON or string and all seems to be working.
This change is only affect "checkboxes" form field.

ifox:
> I just tried to use different MSQL field types - JSON or string and all seems to be working.

That's the thing. The intention embedded in the current implementation is that you'd use the checkboxes field with a relationship. That's why it's trying to pull a list of ids.

ResetButton:
I rechecked docs and my project - it will always save list of ids, because we have "value" embedded in blade and we need only "keys" saved in DB to check corresponding fields.

ifox:
Fixed in https://github.com/area17/twill/commit/736a064262bafd4c47b0562136b8a2d859c00ca0. The checkboxes field will now work with a simple text/json column and no casting, as well as with a model relationship with another model.

PR #704, opened 2020-07-21
Allow files in s3 buckets with root directories to not have their root directory paths duplicated

talvbansal:
If an s3 disk has a root directory that isn't / the root directory will be reapplied when calculating the file path. So in those instances we need to strip the duplicated disk root.


talvbansal:
Sorry made a fundamental mistconfiguration which lead me down a long rabbit hole but finally i can see what i was doing wrong now! :)

PR #721, opened 2020-08-03
Add ability to store more data from the Google Maps API

sauron:
Add ability to store more data from the Google Maps API form field.

- Add documentation for the saveExtendedData parameter.
- Add examples of the returned and stored data.


daun:
That's a great idea! How do you feel about saving the zoom level as well? That's usually enough for re-creating the map on the frontend later. That's of course assuming you're actually using Google Maps and not some other provider.

sauron:
Hey @daun, the data stored for the Bounding Box, it is exactly what is displayed in the viewport.
That should be enough for basic calculations like detecting if a given lat, long is included within the selected location.
But, there is definitively room for improvement in the future.

PR #730, opened 2020-08-24
Add Glide cloud sources to allow Glide and S3 integration

antonioribeiro:
To make Glide and S3 and Azure (not implemented yet) work together we just need to add this to `config/twill.php`:

```php
'glide' => [
    'source' => app(A17\Twill\Services\Cloud\Glide::class)->makeCloudSource(
        env('GLIDE_SOURCE', 's3')
    ),
],
```

The parameter is a drive, so if a drive is already configured it will use the drive configuration, otherwise it will use `S3_*` or `AWS_*` from the `.env` file.

The tricky thing here is that Laravel `config` is not fully available when `config/twill.php` is being loaded. So, it will usually get info from the `.env` file, unless developers use this to configure Glide dynamically, further in the application lifecycle.

### DIFF
--- a/src/Services/Cloud/Aws.php
+++ b/src/Services/Cloud/Aws.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace A17\Twill\Services\Cloud;
+
+use Aws\S3\S3Client;
+use Illuminate\Support\Str;
+use League\Flysystem\Filesystem;
+use League\Flysystem\AwsS3v3\AwsS3Adapter;
+
+class Aws
+{
+    public function filesystemFactory($source)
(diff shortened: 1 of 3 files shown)


talvbansal:
Can this be serialized when caching the config? I'd imagine the object returned makeCloudSource would cause an issue. I was doing something similar to solve the same problem but had to remove cache:config from my deployment step for that reason.

ifox (approved):

ifox:
I share the same concern but this can be instantiated in a provider dynamically.

ifox (approved):

PR #731, opened 2020-08-30
[2.x] Create items within a browser field

mazeeblanke:
This Feature allows the developer to create new items within a browser field.

To enable the 'Create new item' button/feature, set either ```$browsers``` or ```$browsersWithCreate```  properties on the module repository:

```
   protected $browsers = [
           'jobs',
           'dogs',
           'cats'
    ];
```

OR 

   ```
 protected $browsersWithCreate = [
        'jobs',
        'dogs',
        'cats'
    ];
```
The properties should contain an array of browser names

The developer can also decide to customize the 'create' form by adding a new file ```create.blade.php``` in the views directory e.g views/admin/{module name}/create.blade.php. 

```
// Sample create.blade.php

@formField('input', [
    'name' => 'title',
    'label' => $titleFormKey === 'title' ? twillTrans('twill::lang.modal.title-field') : ucfirst($titleFormKey),
    'required' => true,
    'translated' => true,
    "fieldsInModal" => $fieldsInModal ?? false,
    'onChange' => 'formatPermalink'
])

@formField('input', [
    'name' => 'description',
    'label' => 'description',
    'translated' => true,
    "fieldsInModal" => $fieldsInModal ?? false,
    'required' => true,
])

@formField('input', [
    'name' => 'slug',
    'label' => twillTrans('twill::lang.modal.permalink-field'),
    'translated' => true,
    'ref' => 'permalink',
    'prefix' => $permalinkPrefix ?? ''
])
```

### DIFF
--- a/src/Repositories/Behaviors/HandleBrowsers.php
+++ b/src/Repositories/Behaviors/HandleBrowsers.php
@@ -3,6 +3,7 @@
 namespace A17\Twill\Repositories\Behaviors;
 
 use A17\Twill\Models\Behaviors\HasMedias;
+use Exception;
 use Illuminate\Support\Str;
 
 trait HandleBrowsers
@@ -27,6 +28,18 @@ trait HandleBrowsers
      * @var string|array(array)|array(mix(string|array))
      */
     protected $browsers = [];
+
+    /**
+     *  Adding browsers names activates the create option on the browser
+     *
+     * [
+     *   'books',
+     *   'publications'
+     * ]
+     *
+     *
+     */
+    protected $browsersWithCreate = [];
 
     /**
      * @param \A17\Twill\Models\Model $object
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: src/Repositories/Behaviors/HandleBrowsers.php [hunk 1]
```diff
@@ -28,6 +29,18 @@ trait HandleBrowsers
...
+     *  Adding browsers names activates the create option on the browser
+     *
+     * [
+     *   'books',
+     *   'publications'
+     * ]
+     *
+     *
+     */
+    protected $browsersWithCreate = [];
```
pboivin, line 42: All browsers will have the create option activated by default, unless some browsers are specified in `$browsersWithCreate`, am I understanding this correctly?

I'm curious to know, have you explored other ways to pass this configuration, like on the browser form field itself, or as an option of the "array" mode of `$browsers`?

views/layouts/form.blade.php
```diff
@@ -110,6 +110,27 @@
     <a17-modal class="modal--browser" ref="browserWide" mode="wide" :force-close="true">
         <a17-browser></a17-browser>
     </a17-modal>
+    @foreach ($browsersWithCreate as $browser)
+        @php
+          $fieldsInModal = $moduleName != $browser['moduleName'];
+          $titleIsTranslatable = count($browser['languages']) > 1;
```
pboivin, line 116: I'm still testing this but I believe this will not work when translations in use, but only one language is configured. The `title` field is not saved correctly in this context.

This is a bit of an edge case but can happen on projects where translations arrive later in the process.

pboivin (commented):
Hi @mazeeblanke, apologies for the delay on this one. This looks good! I've rebased this on the `2.x` branch locally and started testing. I added a few questions and comments below.

haringsrob:
@ifox just a thought on this. But once https://github.com/area17/twill/pull/1431 is merged, we could add "pick existing" to repeaters if they reference a model.

Something like:
@formField('repeater', ['type' => 'video', 'allow_browse' => true]);

Which would solve the problem mentioned here.

haringsrob:
I think this pr is obsolete with https://github.com/area17/twill/pull/1550 which will allow to create or pick existing.

If anything, we could further integrate the browsers functionality into repeaters by for example not allowing to edit existing.

What do you think @ifox?

ifox:
I agree about further integrating browser functionality into the repeater, but I also think there was another interesting value proposition with this PR, which is the ability to create new records using the create modal quickly within a browser field, which would preserve the browser UI exposing a link to open the full form for further content entry.

ifox:
This will be revisited at some point, but won't be merged in 2.x.

PR #748, opened 2020-10-12
minor feature - add option to clone block

chasecrawford:
add option to duplicate block AND copy field values


ifox:
Hi @chasecrawford, this is great! Thank you.

I just pushed a small update to support repeaters inside of blocks, however, I noticed a bug with the current implementation. If you clone a block and edit a field in the newly created block before saving and reloading the page, it edits the field in both. The binding between the field and the store isn't cloned, if you see what I mean. The feature works great after a page reload, but I would be worried about users not realizing and saving the cloned block with invalid content because of this bug.

ifox:
This has been implemented in #918 which is about to be merged.

PR #821, opened 2021-04-14
Use vendor Twill manifest when public/ is missing

antonioribeiro:
This is just a tweak to use `twill-manifest.json` from the vendor folder (which is source of `twill:install` and `twill:update`). 

Had to do this in order to improve our deployments on Laravel Vapor, since it pushes all files from /public to S3, including json files and clears the public folder.

In the process I also created an `Assets` class and refactored the `twillAssets()` helper function a bit, since it was also growing in complexity.

### DIFF
--- a/src/Helpers/frontend_helpers.php
+++ b/src/Helpers/frontend_helpers.php
@@ -1,6 +1,7 @@
 <?php
 
 use Illuminate\Support\Facades\Cache;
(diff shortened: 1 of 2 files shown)


ifox (approved):

ifox:
Thank you so much for reporting @jfw7. We're looking into it.

PR #920, opened 2021-05-21
Fix issues with wrong view name in frontend helper

ptrckvzn:
## Description

View name in frontend helper was returning broken view path.

`twillViewName` was returning `"'admin.'.pages.'.create'"` instead of `"admin.pages.create"`.

## Related Issues

–

### DIFF
--- a/src/Helpers/frontend_helpers.php
+++ b/src/Helpers/frontend_helpers.php
@@ -61,7 +61,7 @@ if (!function_exists('icon')) {
 if (!function_exists('twillViewName')) {
     function twillViewName($module, $suffix)
     {
-        $view = "'admin.'.$module.'.{$suffix}'";
+        $view = "admin.{$module}.{$suffix}";
 
         if (view()->exists($view)) {
             return $view;


antonioribeiro (approved):

antonioribeiro:
Those quotes are there because the code will generate Blade compiled code (PHP).

So it generates something like

``` php
<php

echo twillViewName('admin.posts.form');
```

If we remove the quotes, the final code will be

``` php
<php

echo twillViewName(admin.posts.form);
```

Generating an execution error because PHP would try to find an `admin` constant that doesn't exists.

ptrckvzn:
@antonioribeiro The PHP generated by the directive looks like this.

```
if( view()->exists(twillViewName(($moduleName ?? null), 'create'))) {
```

It's the `($moduleName ?? null)` expression that if being output from the directive and passed to `twillViewName` when the view is called. So it will evaluate the string returned from `twillViewName` but won't interpolate variables in it.

`view()->exists(twillViewName(($moduleName ?? null), 'create'))` with the returned string -> `view()->exists('admin.pages.create')`

It stills feels correct to me to return a string from `twillViewName` without expecting variables interpolation.

antonioribeiro:
You are right, @ptrckvzn. The fist `view()->exists($view)` will never be able to find a view because of those quotes. Good catch!

antonioribeiro (approved):

ifox (approved):

PR #926, opened 2021-05-22
Fix for multi-country locales supported by Translatable

antonioribeiro:
## Description

If we install Laravel and Twill and don't configure Translatable right away, [this error will happen](https://flareapp.io/share/oPRbJ8kP):

Screenshot 2021-05-23 at 00 20 59

Because Twill doesn't support [multi-countries locales present on Translatable](https://github.com/Astrotomic/laravel-translatable/blame/master/src/config/translatable.php#L16):

``` php
'locales' => [
    'en',
    'fr',
    'es' => [
        'MX', // mexican spanish
         'CO', // colombian spanish
    ],
],
```

This PR basically fix this by changing the `i18_helpers->getLocales()` to support them and return:

``` php
[
    "en",
    "fr",
    "es-MX",
    "es-CO",
]
```

It also replaces some instances of `config('translatable.locales')` by the helper.

### DIFF
--- a/src/Helpers/helpers.php
+++ b/src/Helpers/helpers.php
@@ -245,7 +245,6 @@ if (!function_exists('capsule_namespace')) {
     }
 }
 
-
 if (!function_exists('capsule_namespace_to_path')) {
     function capsule_namespace_to_path($namespace, $capsuleNamespace, $rootPath) {
         return capsules()->capsuleNamespaceToPath($namespace, $capsuleNamespace, $rootPath);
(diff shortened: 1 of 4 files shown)


ifox:
Hi @antonioribeiro, that's great to support it! For context, Twill publishes its own version of the `translatable.php` config file with only `'en'` in the locales by default. Did you publish from the translatable package itself? Still I'm going to test this and merge it!

Heqma (approved):

ifox (approved):

PR #929, opened 2021-05-23
Add Twill API documentation generator

pboi20:
## Description

This adds a `docs-api/` directory with a [Doctum](https://github.com/code-lts/doctum) setup to generate static API docs for multiple versions of Twill.

The versions correspond to branches `1.1`, `1.2`, `2.x` and `master`. This can be adjusted more granularly (e.g. tags instead of branches). See `config.php`.

Doctum generates independent sites/bundles for each versions, without a main level `index.html`. I was hesitant to add one. A server-side redirect (e.g. from `twill.io/api` to `twill.io/api/2.x`) could be a simple solution but may require some maintenance over time.

## TODO

- Add a link from the main docs to the API docs
- Rework the default CSS theme to be more in tune with the main docs

## Related Issues

Related to https://github.com/area17/twill/discussions/921

### DIFF
--- a/docs-api/.gitignore
+++ b/docs-api/.gitignore
@@ -1,0 +1,4 @@
+cache/
+build/
+vendor/
+composer.lock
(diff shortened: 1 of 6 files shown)


pboi20:
Awesome! Thanks @ptrckvzn :)

ifox (approved):

PR #936, opened 2021-05-25
Solve Prettier/ESLint conflict and review other formatting features

ptrckvzn:
## Description

I noticed an indentation conflict with ESLint and Prettier when auto formatting in VS Code. This PR removes the conflicting rules by adding prettier to the eslint config.

A node task `npm run format` has been added to run Prettier from the terminal.

Also, suggesting using only `phpcsfixer` for PHP formatting instead of Prettier by removing options for PHP in `.prettierrc.js`, allowing now to have a tab width of 2 spaces for js, vue and scss and leave a tab width of 4 spaces for php.

It adds a PSR2 configuration file for PHP_CodeSniffer.

Also adds a documentation block in the `Contributing` section to setup corresponding plugins in VS Code.

## Related Issues

—


haringsrob:
Should we maybe take this opportunity to set **psr12** as the standard going forward? Then I think that we can clear out the phpcs.xml config and just stick with the default psr12 rules.

I personally do not like this idea of using prettier (js) to do the formatting for PHP. But if it works well, why not.

haringsrob:
Maybe while we are at it, we can add static code analysis with Phan or phpstan to catch bugs early as well.

haringsrob:
Hey @ptrckvzn and @ifox,

I had another look at prettier PHP as it seemed to do a good job, however it was a bit too aggressive on some fronts so I changed it to use php-cs-fixer directly so we have more control.

I found a ruleset that leans quite well to the Laravel standards.

I also added a command `php artisan twill:setup-devtools` to copy these configs to the Laravel install you run it from.

Phan I will setup in a separate pr.

ptrckvzn:
Looks amazing! Thanks @haringsrob !

Issue #839, opened 2021-05-07
WYSIWYG Source code not saved


akizor:
Hello,

I'm encountering a wild bug when using `editSource => true` on a WYSIWYG field.

The code
```
    @formField('wysiwyg', [
        'name' => 'description',
        'translated' => true,
        'label' => 'Short description',
        'toolbarOptions' => [
            ['header' => [2, 3, 4, 5, 6, false]],
            'bold',
            'italic',
            'underline',
            'strike',
            ["script" => "super"],
            ["script" => "sub"],
            "blockquote",
            "code-block",
            ['list' => 'ordered'],
            ['list' => 'bullet'],
            ['indent' => '-1'],
            ['indent' => '+1'],
            ["align" => []],
            ["direction" => "rtl"],
            'link',
            "clean",
        ],
        'placeholder' => 'Short description here',
        'maxlength' => 400,
        'editSource' => true,
        'note' => 'Will show as an excerpt in listings',
    ])
```

The above code renders something like this.
<img width="965" alt="Screen Shot 2021-05-07 at 5 15 12 PM" src="https://user-images.githubusercontent.com/1052439/117462967-d70e8280-af57-11eb-9e46-13154229f958.png">

When I edit in source code view, everything added or changed is not saved. However, if I edit using source code and then revert to WYSIWYG mode and after that, I save the model, everything works.

Somehow, the changes added when editing in source code mode are not saved on the first try.

Any help is appreciated. Thank you

LE: I'm using all the traits (translation, revisions, etc)

tchapi:
Do you have an error in your console by any chance ? I think this is due to a bug in Quill directly

pboi20:
Hey @akizor,

I can confirm this bug, following the steps you described :

1. Click the Source code button to edit the markup
2. Add some text
3. Click directly on the Update button
4. Fully refresh the page

The field still contains the old value, without the added text from step 2.


No errors in the console on my end. I'll do some more testing but so far, I don't think it's a Quill bug. Updating the field in "source mode" does reflect in the DOM, on the corresponding input field :

![wysiwyg-source](https://user-images.githubusercontent.com/7805679/121573503-5843b400-c9f3-11eb-84d0-0f5245fc8189.png)

Does any of this make sense? I should say that I'm not that familiar with how Quill actually operates under the hood :)

ifox:
@pboi20 there are few different bugs being discussed here, and I'll come back to provide more insights. The one you identified specifically is due to the absence of blur event before you click update.

[...]

PR #970, opened 2021-06-11
Save WYSIWYG field value while in "source mode"

pboi20:
## Description

An attempt to fix the issue of the WYSIWYG field not updating while "source mode" is active.

An `updateSourcecode` method is extracted and called:
- when the "source mode" is toggled (original behavior)
- when the textarea triggers a `change` event (new behavior)

Original steps to reproduce the bug:
1. Click the Source code button to edit the markup
2. Add some text
3. Click directly on the Update button
4. Fully refresh the page

With this fix, the new value is pushed into the store as part of step 3, before the AJAX request.

## Related Issues

https://github.com/area17/twill/issues/839


pboi20:
Update — The tiptap version of the field also had the same bug. Nothing to do with quill or tiptap actually, but the separate textarea used for editing the source.

ifox (approved):

PR #972, opened 2021-06-15
Added option to make multi select searchable

domihagen:
## Description

Added Option to make the Multi Select searchable without an Endpoint. Keept it backward compatible and updated the docs.


pboi20:
Hey @domihagen,

It could be useful to mention that this only applies when `unpack` is set to `false` (eg. Filter the field values while typing. Used with `'unpack' => false`.)

There are a few field settings here and there that apply conditionally like this. I wonder what would be the best way to reflect that in the docs... 🤔

domihagen:
Hey @pboi20,
i fully agree with you. The docs should be clearer. There are also some options still missing.
But i think that it should be clear to every developer that a search on Checkboxes (`'unpack' => true`) doesn´t make sense.

Thats why i just adopted the docs from the single select und put them into the multi select.

pboi20:
Thanks for the feedback @domihagen and for the patch, most importantly :)

ifox (approved):

PR #984, opened 2021-06-18
fix belongsTo error

Bwat-dev:
Fix preview error when using a belongsTo relationship in browser field

### DIFF
--- a/src/Repositories/Behaviors/HandleRevisions.php
+++ b/src/Repositories/Behaviors/HandleRevisions.php
@@ -2,7 +2,7 @@
 
 namespace A17\Twill\Repositories\Behaviors;
 
-use Illuminate\Database\Eloquent\Relations\BelongsToMany;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\Auth;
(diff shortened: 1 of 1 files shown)


ifox (changes requested):
Thanks @Bwat-dev! It seems like repeaters won't be hydrated anymore with this change. The change regarding browsers makes sense, but we need to keep the code for repeaters, unless I'm missing something.

src/Repositories/Behaviors/HandleRevisions.php
```diff
@@ -15,8 +16,10 @@ public function hydrateHandleRevisions($object, $fields)
             $this->hydrateRepeater($object, $fields, $repeater['relation'], $repeater['model']);
         }
 
+
```
ifox, line 19:
```suggestion
```

src/Repositories/Behaviors/HandleRevisions.php
```diff
@@ -15,8 +16,10 @@ public function hydrateHandleRevisions($object, $fields)
...
         // HandleBrowers trait => getBrowsers
         foreach($this->getBrowsers() as $browser) {
+            //hydrate only if BelongsToMany
```
ifox, line 22:
```suggestion
```

ifox (changes requested):

ifox (approved):

PR #990, opened 2021-06-22
Move svg icons to Blade partials

antonioribeiro:
## Problem

While trying to deploy a Twill app on Laravel Vapor we found an issue related to the icons, which are SVG stuffed directly into the HTML, at Blade compile time:

``` blade
@if(config('twill.dev_mode', false))
    {!! file_get_contents(twillAsset('icons.svg')) !!}
    {!! file_get_contents(twillAsset('icons-files.svg')) !!}
    {!! file_get_contents(twillAsset('icons-wysiwyg.svg')) !!}
@else
    {!! File::exists(public_path(twillAsset('icons.svg'))) ? File::get(public_path(twillAsset('icons.svg'))) : '' !!}
    {!! File::exists(public_path(twillAsset('icons-files.svg'))) ? File::get(public_path(twillAsset('icons-files.svg'))) : '' !!}
    {!! File::exists(public_path(twillAsset('icons-wysiwyg.svg'))) ? File::get(public_path(twillAsset('icons-wysiwyg.svg'))) : '' !!}
@endif
```

The problem is that Vapor deletes all files from the public directory (they are uploaded to S3 during deployment) and we don't see a way of telling it to not do it unless we fork the vapor-cli script and change it ourselves.

The current implementation of Twill expects the SVG files to be present on `/public`. Even if we overload `twillAsset()` to load those files from a different directory (or on S3), Twill is still expecting them to be inside `/public`, as per the snippet above.

## Description

This PR moves all `.svg` icon files to the folder `views/partials/icons` and import them as normal blade templates:

``` blade
<div class="svg-sprite">
    @include('twill::partials.icons.icons-svg')

    @include('twill::partials.icons.icons-files-svg')

    @include('twill::partials.icons.icons-wysiwyg-svg')
</div>
```

I removed the creation of hashed (versioned) files, because blade files will be recompiled automatically by Laravel when they change.


antonioribeiro:
@ifox , can you do a quick review? No need to merge it now, just to check the solution, because we are probably using this (as a fork) on a project on Vapor today, but if this implementation is not approved, we may be stuck with the fork forever, or we will have to get back to this to use a new implementation or do something different to get back to Twill 2.x.

ifox:
This solution looks good to me and I will approve and merge as soon as I can make sure this has no impact elsewhere.

antonioribeiro:
Awesome, thanks @ifox !!

ifox (approved):

PR #993, opened 2021-06-24
Register capsules routes before Twill internal routes

joyceverheije:
## Description

While trying to set up a capsule under the `settings` navigation namespace, I wasn't able to register the module's routes under a group with the `settings` prefix. The index route was redirecting to the CMS index because Twill did not find a settings Blade file for that route. Registering the capsules routes before Twill's internal routes is making that possible.

### DIFF
--- a/src/RouteServiceProvider.php
+++ b/src/RouteServiceProvider.php
@@ -48,6 +48,8 @@ class RouteServiceProvider extends ServiceProvider
             $this->getRouteMiddleware(),
             $this->supportSubdomainRouting()
         );
+        
+        $this->registerCapsulesRoutes($router);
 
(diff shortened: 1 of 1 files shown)


antonioribeiro (approved):
Yes, settings routes are a bit generic (`admin/settings/{section}`), so if they are in front of others (like `admin/settings/posts`) they will catch them first and then trigger a `redirect()->back()`.

I was intrigued by this, so I decided to test it and ended up finding that Laravel routes cannot be listed in resolution order, so I ended up pushing a [PR to Laravel](https://github.com/laravel/framework/compare/8.x...antonioribeiro:patch-9) too :)   

Good catch and thank you!

ifox (approved):

Issue #604, opened 2020-03-23
CLI generated $mediasParams config creates confusion


mzalevski:
**Problem:** When I've generated a model with Twill CLI the $mediasParams were set to "desktop" and "mobile" configs. The blade render function "{!! $item->image('cover') !!}" expects a second argument if you want to set any config other than "default".

**Solution:** Change the Twill CLI to generate "default" instead of "desktop" or provide info in the docs that one needs to pass the second argument to blade render function.

PR #1027, opened 2021-06-30
Update model stub default crops

pboi20:
## Description

Small update to `model.stub` to use `default` instead of `desktop` as first crop name. This makes generated models compatible out of the box with the various methods from `HasMedias` that use `default` as implicit crop name.

## Related Issues

Fixes https://github.com/area17/twill/issues/604

### DIFF
--- a/src/Commands/stubs/model.stub
+++ b/src/Commands/stubs/model.stub
@@ -28,7 +28,7 @@ class {{modelClassName}} extends Model {{modelImplements}}
     {{/hasSlug}}{{hasMedias}}
     public $mediasParams = [
         'cover' => [
-            'desktop' => [
+            'default' => [
                 [
                     'name' => 'desktop',
                     'ratio' => 16 / 9,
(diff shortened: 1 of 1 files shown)


src/Commands/stubs/model.stub
```diff
@@ -28,7 +28,7 @@ class {{modelClassName}} extends Model {{modelImplements}}
     {{/hasSlug}}{{hasMedias}}
     public $mediasParams = [
         'cover' => [
-            'desktop' => [
+            'default' => [
                 [
                     'name' => 'desktop',
```
ifox, line 33:
```suggestion
                    'name' => 'default',
```

ifox (changes requested):
Definitely a welcome change for newcomers. I even wonder if it's too opinionated to ship that whole configuration by default, but that's another consideration.

pboi20:
@ifox That's a good point! I personally enjoy when generated code is closer to a full example than to a simple skeleton. But looking at `flexible` here, I see how it can be confusing for someone who has yet to explore more elaborate crop configurations.

ifox (approved):

Issue #434, opened 2019-10-22
Bug - Add new row multi click Create issue


Equruler:
When trying to add new item,
if multi click the "Create" button,
multi records will be created.

Suggest to have js click block on the "Create" button.

yanhao-li:
Confirmed this is a bug, thanks for reporting!

Issue #777, opened 2021-01-06
[Bug] WYSIWYG not saving correctly in draft mode


adronitis:
- Laravel 8.21
- Twill 2.1.1

Translated WYSIWYG is not saving when clicking "Save as draft". 

The only way to work around this is to double save with "Save as draft" or use the other save button or edit any other input field before clicking "Save as draft".

pboi20:
Hey @adronitis,

I think I've been able to replicate this bug on Twill 2.1.1, as well as 2.3.1, but somewhat unreliably...

On my end, the bug happens regardless of draft or live status.

Steps to reproduce:

1. Fully refresh the page with the WYSIWYG field
2. Click directly in the field to bring focus to it and add some text
3. Click directly on the Update button
4. Fully refresh the page again

I get a standard browser warning about unsaved changes, but not reliably. Sometimes the save actually works and the refresh follows through normally.

Does that make any sense? Is this different to what you are describing?

pboi20:
Thanks for taking some time to test! I'll poke around the WYSIWYG source code over the next few days and report here :)

ifox:
@pboi20 thanks for digging into this! The main issue seems similar #839. The issue mentioned with pasted content from Google Docs is more complex. We ran into it to, a few times. It produces an error in console from Quill which is basically not booting correctly on load, or crashing at runtime, preventing any further update to the field content. We've reported the issue: https://github.com/quilljs/parchment/issues/87. There are very specific invisible characters causing this. We've went as far as adding a dynamic "fix WYSIWYG fields" button to the form sidebar for some of our clients that would sanitize that faulty content on demand from the backend and reload the page. Definitely not the solution we want to implement in Twill core.

antonioribeiro:
@ifox , someone created a fix for it [here](https://github.com/decidim/decidim/pull/7999), should we fork and apply it until Quill developers are able to fix it? That issue is sitting there for one year.

ifox:
Oh, awesome! We don't need to fork Quill to leverage this, I believe.

pboi20:
@ifox Thanks for the feedback! I see that the copy/pasting issue has been covered extensively in other threads. I'll focus on the main issue — WYSIWYG not saving reliably.

I'll take some more time to test but so far, I think this could be caused by the 600ms debounce on `textUpdate`: https://github.com/area17/twill/blob/2.x/frontend/js/components/Wysiwyg.vue#L236-L238

Looking at the vuex mutations, there is a difference between:
[...]

PR #1029, opened 2021-07-01
Prevent multiple submit events in add and create modals

pboi20:
## Description

This add a simple guard clause in the `submit` handler of `ModalAdd` and `ModalCreate` components. It prevents the creation of duplicate records when double-clicking any submit button.

## Related Issues

Fixes https://github.com/area17/twill/issues/434


ifox:
Great stuff @pboi20. A somewhat related issue in the creation process is the following: if you type very quickly and press enter immediately, the title (or whatever is the first field in the create modal) might not have had the time to be updated in the store before being sent through the backend.

pboi20:
@ifox Thanks, I'll look into the title issue! Actually, I get a feeling that it could be a more generalized issue and also affect fields in the form/edit view.

ifox:
Right, it is related to what we discussed in #777.

pboi20:
Also https://github.com/area17/twill/issues/777, which is a different issue but effectively the same problem.

ifox (approved):

PR #1036, opened 2021-07-04
Restore ability to create multiple records when using editInModal

ifox:
## Description

Following #1029, when using `editInModal`, it was impossible to submit more than once since the page does not reload.


frontend/js/components/modals/ModalCreate.vue
```diff
@@ -98,6 +98,7 @@
             if (self.$refs.modal) self.$refs.modal.close()
 
             self.$nextTick(function () {
+              this._isSubmitting = false
```
ifox, line 101: Actually, this needs to happen in case an an error as well.

pboi20: Super good catch! Do you think this could be handled in `finally`? eg.
```
 .finally(() => {
    self.$nextTick(function () {
      this._isSubmitting = false
    })
  })
```

ifox: Yes, this looks like the best option, thanks!

pboi20:
```suggestion
```

pboi20 (approved):

PR #1042, opened 2021-07-07
Allow Google2FA-QRCode 2.0 to add support for chillerlan php-qrcode

antonioribeiro:
## Description

BaconQR code requires (required?) the Imagick extension, which poses some problems on certain environments. This PR allows Composer to install Google2FA-QRCode 2.x, if possible, which support both BaconQRCode and [Chillerlan PHP QRCode](https://github.com/chillerlan/php-qrcode).

This is probably a breaking change, because people will have to choose and explicitly install one of those: 

``` sh
composer require chillerlan/php-qrcode
```

``` sh
composer require bacon/bacon-qr-code
```

This is the endresult on Twill with Chillerlan:

Screenshot 2021-07-07 at 19 09 04

### DIFF
--- a/src/TwillServiceProvider.php
+++ b/src/TwillServiceProvider.php
@@ -2,6 +2,7 @@
 
 namespace A17\Twill;
 
+use Exception;
 use A17\Twill\Commands\BlockMake;
 use A17\Twill\Commands\Build;
 use A17\Twill\Commands\CapsuleInstall;
@@ -86,6 +87,8 @@ class TwillServiceProvider extends ServiceProvider
 
         $this->extendBlade();
         $this->addViewComposers();
+
+        $this->check2FA();
     }
 
     /**
@@ -477,5 +480,24 @@ class TwillServiceProvider extends ServiceProvider
     public function version()
     {
         return static::VERSION;
+    }
+
+    /**
+     * In case 2FA is enabled, we need to check if a QRCode compatible package is
+     * installed.
+     */
+    public function check2FA()
+    {
+        if (!config('twill.enabled.users-2fa')) {
+            return;
+        }
+
+        try {
+            (new User())->get2faQrCode();
+        } catch (Exception $e) {
+            throw new Exception(
+                "Twill ERROR: As you have 2FA enabled, you also need to install a QRCode service package, please check https://github.com/antonioribeiro/google2fa-qrcode#built-in-qrcode-rendering-services"
+            );
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


antonioribeiro:
I just pushed a commit that will make the Service Provider to raise an exception in case the a QRCode service class is not available, when 2FA is enabled. We can then say this is not a breaking change anymore? 🤔

Screenshot 2021-07-08 at 09 52 36

src/TwillServiceProvider.php
```diff
@@ -478,4 +481,23 @@ public function version()
...
+     * installed.
+     */
+    public function check2FA()
+    {
+        if (!config('twill.enabled.users-2fa')) {
+            return;
+        }
+
+        try {
+            (new User())->get2faQrCode();
```
ifox, line 496: Generating a QR code during every single request when the feature is enabled doesn't sound great to me, am I ready this incorrectly?

ifox: I do like the idea of letting the developer know like that, but we don't have to avoid the breaking change if we can't, this could wait for Twill 3, right?

antonioribeiro: Horrible actually! I didn't thought it could take too much time, but it's actually too slow, so I just pushed a better implementation.

Timed, running it 100 times:

```
Generating QRCode: 5433ms
Instantiating the class: 1.91ms
Doing nothing: 0.13ms (just commented the line)
```

Can probably be better if I don't instantiate it, but this would mean to make Twill know which classes are necessary for the QRCode to work fine.

antonioribeiro: This is what it does internally: 

Screenshot 2021-07-09 at 01 06 07

ifox: That's great @antonioribeiro! I wonder if we should also scope this check to the console, or to the CMS routes. 1.91ms there is totally fine, but I'd like to avoid unnecessarily affecting the frontend when we're looking for every ms we can get there nowadays.

antonioribeiro: It's actually 0,0191ms/request 😆. But I agree that it's still 2ms and summing to all other ms we have, it's not great.

Moving to check only the existence of the classes (no instantiation) we get down to 0,0082ms/request (1ms).

A call to `App::runningInConsole()` runs in 0,0043ms.

A call to `config('twill.enabled.users-2fa')` runs in 0.0178ms (this one is also execute every time at the beginning and the class check is ran only if 2FA is enabled).

So, yeah, it's not great anyway, and maybe we should just make it a breaking change and only deploy this on the next major. Remembering this part is there just for helping developers not being caught by a broken production because they didn't tested 2FA locally before deploying their apps.

antonioribeiro: A call to `app()->runningInConsole()` runs in 0,0025ms.

A call to `$this->app->runningInConsole()` runs in 0,0016ms. (this is the current implementation)

ifox (approved):

PR #1048, opened 2021-07-13
Add table aliases

antonioribeiro:
## Description

When adding some joins I always get into the situation where the internal Twill scopes are not aliased and Laravel raises the exception saying that the `published` column is ambiguous. I believe this is a behaviour can be found on other classes (like the scopePublished on the User class) and probably needs a larger review by someone with more global knowledge of queries in Twill. But this one fixed query joins done directly between module models.

### DIFF
--- a/src/Models/Behaviors/HasPosition.php
+++ b/src/Models/Behaviors/HasPosition.php
@@ -19,7 +19,7 @@ trait HasPosition
 
     protected function getCurrentLastPosition()
     {
-        return ((int) static::max('position'));
+        return ((int) static::max("{$this->getTable()}.position"));
     }
 
     /**
@@ -28,7 +28,7 @@ trait HasPosition
      */
     public function scopeOrdered($query)
     {
-        return $query->orderBy('position');
+        return $query->orderBy("{$this->getTable()}.position");
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Models/Behaviors/HasPosition.php
```diff
@@ -19,7 +19,7 @@ protected function setToLastPosition()
 
     protected function getCurrentLastPosition()
     {
-        return ((int) static::max('position'));
+        return ((int) static::max("{$this->getTable()}.position"));
```
ifox, line 22: @antonioribeiro I was about to approve this but I realized this isn't a scope and shouldn't be necessary here, correct?

ifox (changes requested):

src/Models/Behaviors/HasPosition.php
antonioribeiro: Hey @ifox , this isn't a scope, yes, but it's a query, and if we have a global scope on this model joining a bunch of other tables, when we try to get the last position, the `position` column will have no alias and may clash wish some other table. I added it later because it happened to me during a save event. I also prevented by allowing the scope only if the application is running on frontend, but other people might not be able to do this.

ifox (approved):

PR #1049, opened 2021-07-13
[3.x] Feature: Permissions Management

pboivin:
## Description

A draft PR continuing the good work of #994. The original branch is being manually rebased on `2.x` with a focus on maximizing backward compatibility. This is a work in progress :)


migrations/default/2021_07_08_000001_update_twill_default_users_table.php
```diff
@@ -0,0 +1,66 @@
...
+    private function seedNewFields()
+    {
+        User::chunk(100, function ($users) {
+            foreach ($users as $user) {
+                if (!empty($user->password)) {
+                    $user->activated = true;
+                }
+
+                if (!empty($user->created_at)) {
+                    $user->registered_at = $user->created_at;
```
ifox, line 59: @pboi20 i wonder if this should be set in the above condition too. If the user currently doesn't have a password, it is not considered registered. It makes me question the need for the activated boolean, do we actually need it on top of the registration date?

pboi20: Makes sense! Good question about the `activated` column, I will dig a bit deeper and confirm if it's actually needed. At a glance, it does feel somewhat redundant...

PR #1076, opened 2021-08-05
fix: Disable permalink option in controller stub if module has no slugs

pboi20:
## Description

Reported earlier today on Discord by `@leftblank`:
> I have created a module which wasn't specified to use slugs, and clicking the Edit icon next to the title also shows an editable Permalink field, which doesn't seem to update after a save. 

This adds a check to ensure that the permalink field is hidden if the module has no slugs.

## Related Issues

—


pboi20:
Not sure if this is cleaner — Another option is to handle this in the controller stub:
If the module was generated without slugs, prefill the `$indexOptions` with `'permalink' => false`

ifox:
Would using `||` instead of `&&` be a good option?

ifox:
Unless I'm not properly awake, I think that would allow existing users that have leveraged the permalink field without slugs enabled to continue doing so by setting their indexOptions to true. I'm afraid both approaches are somewhat of a breaking change.

pboi20:
I believe using `||` would negate the check for `moduleHas('slugs')`, because of the default value of `'permalink' => true` in `$defaultIndexOptions`.

Curious to know what you think of handling this in the controller stub? This would make it a non-breaking change for existing projects.

ifox:
That makes sense with our default, I thought it'd be false without looking. Let's move forward with the stub approach.

ifox (approved):

PR #1080, opened 2021-08-09
Add Ukrainian lang

Dmi3yy:
Add Ukrainian language

### DIFF
--- a/lang/uk/lang.php
+++ b/lang/uk/lang.php
@@ -1,15 +1,439 @@
 <?php
-    /*
-    |--------------------------------------------------------------------------
-    | 5 Steps to Contribute a New Twill Localization at Ease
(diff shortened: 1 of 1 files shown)


EPInformatyka:
I propose to rename dirname and use the official Ukrainian shortcode (ua)
I think, "uk" belongs to the English language group

Dmi3yy:
> I propose to rename dirname and use the official Ukrainian shortcode (ua)
> I think, "uk" belongs to the English language group

No, need. Correct is UK, by standards. 

I know we all time inside Ukraine use ua, but this wrong.

layout-lab:
> I think, "uk" belongs to the English language group

GB is the locale for English in UK.

PR #1092, opened 2021-08-20
Feature/i18n permalinks

MamlukiSn:
## Description
This PR enables the translation of permalinks.
Useful when frontend URL's are also translated based on the locale.

## Usage

1. Allows localization of the permalinkBase property of the ModuleController
```
//return an array from the getLocalizedPermalinkBase controller method with the keys being locales
//the slug attribute will be appended to the value

/**
* @return array
*/
protected function getLocalizedPermalinkBase()
{
    return [
        'en' => 'collection/artists',
        'fr' => 'collection/artistes',
    ];
}
```

2. Allow localization of the customPermalink property by use of localizedCustomPermalink property instead

```
//pass localizedCustomPermalink to the form data which is an array with the keys being locales and values being the translated urls
protected function formData($request)
{
    return [
        'localizedCustomPermalink' => [
            'en' =>'https://website.com/en/news',
            'fr' =>'https://website.com/fr/nouvelles',
        ],
    ];
}
```

### DIFF
--- a/frontend/js/components/TitleEditor.vue
+++ b/frontend/js/components/TitleEditor.vue
@@ -120,7 +120,7 @@
       },
       permalink: function () {
         const localizedPermalinkbase = this.localizedPermalinkbase.length > 0 ? JSON.parse(this.localizedPermalinkbase) : {}
-        return Object.keys(localizedPermalinkbase).length > 0 ? ((this.currentLocale.value in localizedPermalinkbase) ? localizedPermalinkbase[this.currentLocale.value].concat('/',this.fieldValueByName('slug')[this.currentLocale.value]) : this.fieldValueByName('slug')[this.currentLocale.value] ): this.fieldValueByName('slug')[this.currentLocale.value]
+        return Object.keys(localizedPermalinkbase).length > 0 ? ((this.currentLocale.value in localizedPermalinkbase) ? localizedPermalinkbase[this.currentLocale.value].concat('/', this.fieldValueByName('slug')[this.currentLocale.value]) : this.fieldValueByName('slug')[this.currentLocale.value]) : this.fieldValueByName('slug')[this.currentLocale.value]
       },
       ...mapState({
         baseUrl: state => state.form.baseUrl,
--- a/frontend/js/components/TitleEditor.vue
+++ b/frontend/js/components/TitleEditor.vue
@@ -71,7 +71,7 @@
         type: String,
         default: ''
       },
-      customLocalizedPermalinkbase: {
+      localizedPermalinkbase: {
         type: String,
         default: ''
       },
@@ -119,8 +119,8 @@
         return Object.keys(localizedCustomPermalink).length > 0 ? localizedCustomPermalink[this.currentLocale.value] : (this.customPermalink ? this.customPermalink : false)
       },
       permalink: function () {
-        const customLocalizedPermalinkbase = this.customLocalizedPermalinkbase.length > 0 ? JSON.parse(this.customLocalizedPermalinkbase) : {}
-        return Object.keys(customLocalizedPermalinkbase).length > 0 ? customLocalizedPermalinkbase[this.currentLocale.value].concat(this.fieldValueByName('slug')[this.currentLocale.value]) : this.fieldValueByName('slug')[this.currentLocale.value]
+        const localizedPermalinkbase = this.localizedPermalinkbase.length > 0 ? JSON.parse(this.localizedPermalinkbase) : {}
+        return Object.keys(localizedPermalinkbase).length > 0 ? ((this.currentLocale.value in localizedPermalinkbase) ? localizedPermalinkbase[this.currentLocale.value].concat('/',this.fieldValueByName('slug')[this.currentLocale.value]) : this.fieldValueByName('slug')[this.currentLocale.value] ): this.fieldValueByName('slug')[this.currentLocale.value]
       },
       ...mapState({
         baseUrl: state => state.form.baseUrl,
--- a/frontend/js/components/TitleEditor.vue
+++ b/frontend/js/components/TitleEditor.vue
@@ -74,6 +74,10 @@
       customLocalizedPermalinkbase: {
         type: String,
         default: ''
+      },
+      localizedCustomPermalink: {
+        type: String,
+        default: ''
       }
     },
     data: function () {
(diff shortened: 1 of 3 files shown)


daun:
Nice to see somebody tackle this!

To follow the way other Twill features are handled, the localized permalinks should probably be supplied through a method on the controller, similar to the original `getPermalinkBaseUrl` method. E.g. `getLocalizedPermalinkBaseUrls`. Requiring the `json_encode` in every controller is error-prone and a separate method would make it easy to see what that code is doing. The array can be converted to json in blade.

src/Http/Controllers/Admin/ModuleController.php
```diff
@@ -1698,8 +1700,16 @@ protected function getPermalinkBaseUrl()
         return $appUrl . '/'
             . ($this->moduleHas('translations') ? '{language}/' : '')
             . ($this->moduleHas('revisions') ? '{preview}/' : '')
-            . ($this->permalinkBase ?? $this->getModulePermalinkBase())
-            . (isset($this->permalinkBase) && empty($this->permalinkBase) ? '' : '/');
+            . (empty($this->getLocalizedPermalinkBase()) ? ($this->permalinkBase ?? $this->getModulePermalinkBase()) : '')
+            . (((isset($this->permalinkBase) && empty($this->permalinkBase)) || !empty($this->getLocalizedPermalinkBase())) ? '' : '/');
```
pboivin, line 1704: Can this be updated to avoid having to provide the final `/` in `getLocalizedPermalinkBase()`? For example:
```
    protected function getLocalizedPermalinkBase()
    {
        return [
            'en' => 'content',
            'fr' => 'contenu',
        ];
    }
```
This would be more similar to the behavior of `$permalinkBase`.

frontend/js/components/TitleEditor.vue
```diff
@@ -106,8 +114,13 @@
...
         return titleValue || this.warningMessage
       },
+      customlink: function () {
+        const localizedCustomPermalink = this.localizedCustomPermalink.length > 0 ? JSON.parse(this.localizedCustomPermalink) : {}
+        return Object.keys(localizedCustomPermalink).length > 0 ? localizedCustomPermalink[this.currentLocale.value] : (this.customPermalink ? this.customPermalink : false)
+      },
       permalink: function () {
-        return this.fieldValueByName('slug')[this.currentLocale.value]
+        const customLocalizedPermalinkbase = this.customLocalizedPermalinkbase.length > 0 ? JSON.parse(this.customLocalizedPermalinkbase) : {}
+        return Object.keys(customLocalizedPermalinkbase).length > 0 ? customLocalizedPermalinkbase[this.currentLocale.value].concat(this.fieldValueByName('slug')[this.currentLocale.value]) : this.fieldValueByName('slug')[this.currentLocale.value]
```
pboivin, line 123: I think there is no fallback for undefined locales in`getLocalizedPermalinkBase()`. For example, this would crash on the frontend:
```
protected function getLocalizedPermalinkBase()
    {
        return [
            'en' => 'content',
        ];
    }
}
```
when a user selects `FR` in the CMS. However `localizedCustomPermalink` seems to fallback nicely to the basic value of `$permalinkBase` for missing locales.

pboivin (changes requested):
Hi @MamlukiSn, this is cool! I believe it's the final missing piece for fully configurable permalinks in `TitleEditor`.

I've added a few questions and comments below.

views/layouts/form.blade.php
```diff
@@ -47,6 +47,8 @@
                     :editable-title="{{ json_encode($editableTitle ?? true) }}"
                     custom-title="{{ $customTitle ?? '' }}"
                     custom-permalink="{{ $customPermalink ?? '' }}"
+                    custom-localized-permalinkbase="{{ json_encode($localizedPermalinkBase ?? '') }}"
```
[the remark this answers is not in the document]
MamlukiSn: @pboivin . Yes, makes sense to rename. I've pushed an update that refactors this.

src/Http/Controllers/Admin/ModuleController.php
MamlukiSn: @pboivin . I've pushed an update to factor this in. The trailing slash is added on the Frontent end
```
      permalink: function () {
        const localizedPermalinkbase = this.localizedPermalinkbase.length > 0 ? JSON.parse(this.localizedPermalinkbase) : {}
        return Object.keys(localizedPermalinkbase).length > 0 ? ((this.currentLocale.value in localizedPermalinkbase) ? localizedPermalinkbase[this.currentLocale.value].concat('/',this.fieldValueByName('slug')[this.currentLocale.value]) : this.fieldValueByName('slug')[this.currentLocale.value] ): this.fieldValueByName('slug')[this.currentLocale.value]
      },
```

Due to how the Frontend Works by replacing parts of the baseUrl with the current locale;
```
      fullUrl: function () {
        return this.customlink || this.baseUrl
          .replace('{language}', this.currentLocale.value)
          .replace('{preview}/', this.published ? '' : 'admin-preview/') + this.permalink
      },
```
The section referenced in the initial comment serves to not include the permalinkBase `$this->permalinkBase` as part of the base URL when `getLocalizedPermalinkBase` is not empty.

```
    protected function getPermalinkBaseUrl()
    {
        $appUrl = Config::get('app.url');

        if (blank(parse_url($appUrl)['scheme'] ?? null)) {
            $appUrl = $this->request->getScheme() . '://' . $appUrl;
        }

        return $appUrl . '/'
            . ($this->moduleHas('translations') ? '{language}/' : '')
            . ($this->moduleHas('revisions') ? '{preview}/' : '')
            . (empty($this->getLocalizedPermalinkBase()) ? ($this->permalinkBase ?? $this->getModulePermalinkBase()) : '')
            . (((isset($this->permalinkBase) && empty($this->permalinkBase)) || !empty($this->getLocalizedPermalinkBase())) ? '' : '/');
    }
```

frontend/js/components/TitleEditor.vue
MamlukiSn: @pboivin . Fixed to fall back to the records slug with the assumption that the permalinkBase for the undefined locale has not been set. This is also limited by how the FE constructs the full URL from the base url (Cannot fall back to the basic value of `$permalinkBase` since it would be missing from the base URL). See https://github.com/area17/twill/pull/1092#discussion_r741870901

src/Http/Controllers/Admin/ModuleController.php
pboivin: @MamlukiSn Thank you for taking the time to look into this! Your solution works perfectly.

frontend/js/components/TitleEditor.vue
pboivin: Thanks! I think The fallback to the bare slug is a good alternative.

pboivin (approved):

Issue #595, opened 2020-03-17
php artisan twill:dev not watching changes on custom components


zedee:
Hi, after a bit of poking, I successfully ran `php artisan twill:dev` from a docker container ( issue #584 )  the odd thing now is that despite it's compiling successfully all changes, it does not 'sense' any changes and I have to run again `php artisan twill:dev` in order to reload changes.

My work directory is `./resources/assets/js/components`

Is there anything I'm not doing right?

is-blackhole:
Same problem here, `php artisan twill:dev` doesnt detect files changes while running inside docker container

Issue #1014, opened 2021-06-28
twill:dev does not work(no docker)


ammma:
## Description
I'm developing custom components and i've following issues running twill:dev
- changing `TWILL_DEV_MODE_URL` env variable doesn't make any difference, it keeps serving `http://localhost:8080` (i just had other server running same port so i had to remove it.)
- when i run `php artisan twill:dev` with `TWILL_DEV_MODE=true` then `http://localhost:8080` brings me `Cannot GET /` and `http://localhost:8080/twill-manifest.json` is showing file content from `public/assets/admin/twill-manifest.json` Have no idea how to solve it. Google says it could be a webpack issue but i didn't try to downgrade it or something.
- `php artisan twill:dev` does not watch for changes in `./resources/assets/js/components`. But this could be a normal behavior, i don't know. If so then developing a custom component is turning into a terrible nightmare:)

ammma:
For those who are struggling with same thing: https://github.com/benblamey/when_changed
just use it to monitor folder and run twill:build

ptrckvzn:
Hi @ammma, I've been doing quite a bit of local development on Twill recently and maybe I can help you. Have you tried using your local network IP instead of localhost? I use Homestead (instead of Sail/Docker) and setting the env variables to the local IP that `php artisan twill:dev` outputs makes local development possible.

![image](https://user-images.githubusercontent.com/6365432/123645415-5308a680-d7f4-11eb-90ee-66935f65fc7f.png)

```
TWILL_DEV_MODE_URL=http://10.0.0.10:8080
TWILL_DEV_MODE=true
```

Please let me know if that works for you.

As for your third point, that `twill:dev` doesn't check `./resources/assets/js/components`, I will have a look, but I feel you might be right as the components are copied over to the twill folder on build. If that's the case, let's see if we can improve that!

pboi20:
Hey @ammma,

I like your solution to use `when_changed` to trigger `twill:build`! Although, compared to `twill:dev`, this probably rebuilds the full production bundles. It's considerably slower (~30s on my machine).

To echo what @ptrckvzn mentioned, your value for `TWILL_DEV_MODE_URL` will depend on your dev setup (eg. Homestead vs. WAMP or similar).

Also, when using `TWILL_DEV_MODE=true` and `TWILL_DEV_MODE_URL=http://localhost:8080`, you may access your site at the same URL as before (eg. http://example.test). I believe `TWILL_DEV_MODE_URL` is only serving your assets.

[...]

PR #1107, opened 2021-08-27
feat: Watch for custom blocks & components in development

pboivin:
## Description

This is a draft PR for a proposed improvement to the `twill:dev` workflow:

1. A `--copyOnly` option is added to the `twill:build` command

2. A check to find a project-level install of `chokidar-cli` is added. If found, a background process is started to trigger `twill:build --copyOnly` when custom blocks or components are modified. Otherwise, a message is displayed with instructions on how to install the package.

This works for me locally but I'd love to get some feedback to make it more robust.

Would this even be usable on Windows?

Should the `chokidar-cli` be added to Twill's internal devDependencies instead?

<br>
<br>

**Install instructions:**

01-warning

**Watcher started:**

02-watching

## Related Issues

Fixes https://github.com/area17/twill/issues/595
Related https://github.com/area17/twill/issues/1014

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@area17/twill",
-  "version": "2.3.0",
+  "version": "2.4.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


ifox:
This is awesome @pboivin, thank you! If we can make this happen without requiring the dependency at the project level, it'd be perfect, but I think it is totally acceptable if not.

Issue #1117, opened 2021-09-01
Browser fields implemented with HasRelated do not hydrate when previewing


joyceverheije:
## Description

When using the `HasRelated` trait to save a browser field, previewing a change does not work.

### Steps to reproduce

- Make changes to a browser field implemented with `HasRelated`, eg. reordering or changing selected records
- Preview without saving

### Expected result

The changes made to the browser field should be hydrated on the previewed model

### Actual result

The previewed model doesn't include the changes, like regular browser fields with a pivot table would

### Versions

Twill version:  2.x

pboivin:
Hi @joyceverheije, I can confirm this issue. I was able to replicate it on 2.4.0

Thanks for reporting!

poncianodiego:
I am having an issue trying to view saved items of custom pivot tables:

    public function afterSave(TwillModelContract $object, array $fields): void
    {
        $this->updateBrowser($object, $fields, 'artists');
        $this->updateBrowser($object, $fields, 'featured_artworks');
        parent::afterSave($object, $fields);
    }
    
    none are showing

PR #1087, opened 2021-08-12
fix: Support inverse relationships with belongsTo browsers

pboivin:
## Description

This adds support for inverse relationships with belongsTo browsers: hasOne & hasMany.

## Related Issues

Follows/depends on https://github.com/area17/twill/pull/1085

PR #1130, opened 2021-09-03
fix: Hydrate related browser fields for preview and restore

pboivin:
## Description

This adds the missing steps to hydrate browser fields implemented through `HasRelated`. Special attention is paid to the content of `relatedItems` when hydrating, in order to preserve any related item that is not directly tied to a known browser field.

The tests were updated to cover this new behavior. The tests for `belongsTo` and `hasMany` browsers were also updated along the same lines.

## Related Issues

Fixes https://github.com/area17/twill/issues/1117

Related to https://github.com/area17/twill/pull/1085 https://github.com/area17/twill/pull/1087

### DIFF
--- a/src/Repositories/Behaviors/HandleRevisions.php
+++ b/src/Repositories/Behaviors/HandleRevisions.php
@@ -4,7 +4,6 @@ namespace A17\Twill\Repositories\Behaviors;
 
 use A17\Twill\Models\Behaviors\HasRelated;
 use A17\Twill\Models\RelatedItem;
-use Illuminate\Database\Eloquent\Relations\BelongsTo;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\Auth;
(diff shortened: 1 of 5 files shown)


src/Repositories/Behaviors/HandleRevisions.php
```diff
@@ -138,7 +141,7 @@ public function hydrateBrowser($object, $fields, $relationship, $positionAttribu
      * @param \A17\Twill\Models\Model|null $model
      * @return void
      */
-    public function hydrateOrderedBelongsTomany($object, $fields, $relationship, $positionAttribute = 'position', $model = null)
+    public function hydrateOrderedBelongsToMany($object, $fields, $relationship, $positionAttribute = 'position', $model = null)
```
ifox, line 144: We can clean this up in Twill 3, but for now I think we need to keep the wrongly cased method available in the public API, for case sensitive environments. We can keep the fixed method name and provide the wrongly cased as a deprecated method that calls the new one.

ifox (changes requested):

src/Repositories/Behaviors/HandleRevisions.php
pboivin: @ifox My bad, I was not aware of this. I was under the impression that all methods and functions were case-insensitive in PHP. I will revert the change.

ifox: You're right, ignore me. PHP weirdness, ha. Variables are case sensitive, but classes and functions are not. I wonder however, not in this case specifically, if Laravel itself would be affected. I could swear I already ran into issues with class names and case sensitivity.

ifox (approved):

PR #1151, opened 2021-09-13
Handle JSON columns with Laravel casts in Model and its translations

agnonym:
## Description

When I wanted to use JSON columns in admin forms with the `HandleFieldsGroup`, I noticed a few things:

- The Trait handle JSON encode / decode by itself preventing the use of standard Laravel casts mechanism
- It doesn't work on Translation associated Model
- The fields in the admin form have to be named from the name of the fields of the `fieldsGroups` (the repercussion is you can't have a standard `title` field in your Model and a `title` field in a `fieldsGroups`, or many `title` in different `fieldsGroups`, they will be overwritten)

The purpose of this PR is to allow this, without breaking the current functionality:

- If you provide a cast as an array for your field, the JSON encode / decode is handled by Laravel (and the improvment), if not, it works like before
- You now can have translated JSON fields
- There is new `fieldsGroupsFormFieldNameSeparator` attribute on the `ModuleRepository` which is `null` by default (to keep backward compatibility), but you can provide a value that will be used to prefix the field names in the admin form:
    - $fieldsGroupsFormFieldNameSeparator = null => a `client_id` field of a `settings` `fieldsGroups` will have to be named `client_id`
    - $fieldsGroupsFormFieldNameSeparator = '_' => a `client_id` field of a `settings` `fieldsGroups` will have to be named `settings_client_id`

The code might probably be cleaned, it seems to work in all the cases

### DIFF
--- a/src/Repositories/Behaviors/HandleFieldsGroups.php
+++ b/src/Repositories/Behaviors/HandleFieldsGroups.php
@@ -3,10 +3,10 @@
 namespace A17\Twill\Repositories\Behaviors;
 
 use Illuminate\Support\Arr;
+use Illuminate\Support\Str;
 
 trait HandleFieldsGroups
 {
-
     /**
      * @param \A17\Twill\Models\Model $object
      * @param array|null $fields
@@ -16,7 +16,10 @@ trait HandleFieldsGroups
     {
         foreach ($this->fieldsGroups as $group => $groupFields) {
             if ($object->$group) {
-                $object->$group = json_encode($object->$group);
+                $casts = $this->getModelCasts($object);
+                if (!array_key_exists($group, $casts) || (array_key_exists($group, $casts) && $casts[$group] !== 'array')) {
+                    $object->$group = json_encode($object->$group);
+                }
             }
         }
 
@@ -51,10 +54,23 @@ trait HandleFieldsGroups
     {
         foreach ($this->fieldsGroups as $group => $groupFields) {
             if ($object->$group) {
-                $decoded_fields = json_decode($object->$group, true) ?? [];
+                $casts = $this->getModelCasts($object);
+                if (array_key_exists($group, $casts) && $casts[$group] === 'array') {
+                    $decoded_fields = $object->$group;
+                } else {
+                    $decoded_fields = json_decode($object->$group, true) ?? [];
+                }
+
                 // In case that some field read the value through $item->$name
-                foreach($decoded_fields as $field_name => $field_value) {
-                    $object->setAttribute($field_name, $field_value);
+                foreach ($decoded_fields as $field_name => $field_value) {
+                    if ($this->fieldsGroupsFormFieldNameSeparator !== null) {
+                        $decoded_fields[$group . $this->fieldsGroupsFormFieldNameSeparator . $field_name] = $field_value;
+                        unset($decoded_fields[$field_name]);
+
+                        $object->setAttribute($group . $this->fieldsGroupsFormFieldNameSeparator . $field_name, $field_value);
+                    } else {
+                        $object->setAttribute($field_name, $field_value);
+                    }
                 }
                 $fields = array_merge($fields, $decoded_fields);
             }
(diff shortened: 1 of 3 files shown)


pboivin:
Hi @agnonym,

This is cool! I took some time to test yesterday and it works quite well.

My only comment is about the implementation of `$fieldsGroupsFormFieldNameSeparator`. It took me a few reads to wrap my head around what it does and how to use it.

What do you think of using a boolean property to the same effect? Perhaps `$fieldsGroupsAutoPrefix`, which would be `false` by default. Then, setting it to `true` would mean that you can prefix your form field names with `group_` to avoid collisions, and the prefix is handled automatically.

I don't know if it's that much better... I guess I'm just thinking out loud :)

On a related note — I don't see the need to control just the separator symbol. I believe it is fairly standard to use snake_case for field names in this context.

pboivin:
> On a related note — I don't see the need to control just the separator symbol. I believe it is fairly standard to use snake_case for field names in this context.

Let's scratch this one... I think I'm being short-sighted.

src/Repositories/Behaviors/HandleFieldsGroups.php
```diff
@@ -63,12 +81,37 @@ public function getFormFieldsHandleFieldsGroups($object, $fields)
...
 
+            if ($this->fieldsGroupsFormFieldNameSeparator !== null) {
+                $fieldsGroupWithGroupSeparator = [];
+                foreach ($fields[$group] as $key => $value) {
+                    $fieldsGroupWithGroupSeparator[Str::replaceFirst($group . $this->fieldsGroupsFormFieldNameSeparator, '', $key)] = $value;
+                }
+                $fields[$group] = $fieldsGroupWithGroupSeparator;
+            }
+
+            if (in_array($group, $this->model->translatedAttributes) && is_array($fields[$group])) {
```
pboivin, line 105:
```suggestion
            if (in_array($group, $this->model->translatedAttributes ?? []) && is_array($fields[$group])) {
```
I believe `?? []` or a similar check is needed for models without translations.

agnonym:
> 
> 
> Hi @agnonym,
> 
> This is cool! I took some time to test yesterday and it works quite well.
> 
> My only comment is about the implementation of `$fieldsGroupsFormFieldNameSeparator`. It took me a few reads to wrap my head around what it does and how to use it.
> 
> What do you think of using a boolean property to the same effect? Perhaps `$fieldsGroupsAutoPrefix`, which would be `false` by default. Then, setting it to `true` would mean that you can prefix your form field names with `group_` (where `group` is the actual group name) to avoid collisions, and the prefix is handled automatically.
> 
> I don't know if it's that much better... I guess I'm just thinking out loud :)
> 
> ~On a related note — I don't see the need to control just the separator symbol. I believe it is fairly standard to use snake_case for field names in this context.~

Hi @pboivin, and thanks to have taken time to test!

For `$fieldsGroupsFormFieldNameSeparator`, I hesitated between using a boolean or directly the value of the separator and ended with the separator to be less opiniated. But you're right and I agree with you that prefixing the group name and snake_casing may probably the more natural way to do so.
Maybe a compromise could be to have to variable: as you suggest a boolean to prefix or not, and the prefix with default value `_` that you may change if you want (in rare cases, you mayba can have collision using the `_`).
What do you think of it?

src/Repositories/Behaviors/HandleFieldsGroups.php
agnonym: > 
> 
> I believe `?? []` or a similar check is needed for models without translations.

You're right, I basically used the way test is done in the `HandleTranslations` Trait. Digging deeper, I suggest calling a function on the Model that already do the existence test instead of using directly the attribute:

```
// A17/Twill/Models/Model.php

    public function getTranslatedAttributes()
    {
        return $this->translatedAttributes ?? [];
    }
```

Do you agree with this?

pboivin: I totally agree, very clean!

pboivin:
@agnonym This is a good solution! We'll retain the flexibility of a custom separator (or no separator, e.g. `groupfieldname`). But personnaly, I may always use the boolean. Thinking ahead, this could even be set to `true` as part of a next major release (a breaking change).

agnonym:
Finally got the time to make the discussed modifications:
- Use `$this->model->getTranslatedAttributes()` instead of `$this->model->translatedAttributes`
- Add `$fieldsGroupsFormFieldNamesAutoPrefix` ModuleRepository boolean attribute with default value to `false` (just set it to true in your Model Repository to activate form field names prefix)
- Keep `$fieldsGroupsFormFieldNameSeparator` ModuleRepository attribute but set to `_` instead of `null` which is used only if `$fieldsGroupsFormFieldNamesAutoPrefix` is set to `true`

Hope all this is fine, let me know if the new attributes names are not appropriate

pboivin (approved):
This looks great @agnonym, thank you!

Took some time to retest this on my end and all seems to be working working perfectly.

ifox (approved):

PR #1161, opened 2021-09-18
Avoid additional ajax calls

westberliner:
The app causes with UPDATE_FORM_FIELD and REMOVE_FORM_FIELD a lot of additional unnecessary ajax calls on newly created blocks.

These issues are occurring since 2.5.0.

## Description

I am just getting started to get familiar with the Editor. Not sure if this is the right solution. I am open to better solutions.

But still, non the less two issues:
The first is by creating a block. This will trigger via UPDATE_FORM_FIELD for every field in the block an ajax call. Not on all blocks but on Video-with-Text for example.
Similar issue on removing the same block before submitting. This triggers for every field in the block the REMOVE_FORM_FIELD event and with that an ajax call - if another block comes after the removed one and has now the index of the removed block.

Thx for your feedback upfront and nice project.

## Related Issues

No related issues were found.


pboivin:
Just merged https://github.com/area17/twill/pull/1282 to address this issue. Thanks for all the initial work @westberliner!

westberliner:
Nice, thank you too for the fix.

PR #1170, opened 2021-09-29
Detect when the user has entered a lat/lng into the location field and set the map to that position

mattdinthehouse:
## Description

I had a bunch of specific lat/lng's that didn't correspond to addresses so there was no autocomplete option, and I cbf punching each one into google maps so I could then click-to-place in the Location field's map.

This pull request uses regex to detect when the user has entered a lat/lng string like `-38.149926, 144.361808` and sets the map to there.

## Related Issues

None.

### DIFF
--- a/frontend/js/components/LocationField.vue
+++ b/frontend/js/components/LocationField.vue
@@ -172,6 +172,11 @@
 
         this.address = newValue
         this.$emit('change', newValue)
+
+        const latlng = newValue.match(/^(-?\d+(?:\.\d+)?),+ *(-?\d+(?:\.\d+)?)$/)
+        if (latlng) {
+          this.onLatLngEntered(latlng[1], latlng[2]);
+        }
       },
       onPlaceChanged: function () {
         const place = this.autocompletePlace.getPlace()
@@ -213,6 +218,23 @@
 
         if (this.map) {
           this.addMarker(latlng)
+        }
+
+        // see formStore mixin
+        this.saveIntoStore()
+      },
+      onLatLngEntered: function (lat, lng) {
+        const latlng = new google.maps.LatLng(lat, lng)
+
+        this.clearMarkers()
+        this.clearLatLng()
+
+        this.address = [latlng.lat(), latlng.lng()].join(',')
+        this.setLatLng(latlng)
+
+        if (this.map) {
+          this.addMarker(latlng)
+          this.map.setCenter(latlng)
         }
 
         // see formStore mixin
--- a/frontend/js/components/LocationField.vue
+++ b/frontend/js/components/LocationField.vue
@@ -175,7 +175,7 @@
 
         const latlng = newValue.match(/^(-?\d+(?:\.\d+)?),+ *(-?\d+(?:\.\d+)?)$/)
         if (latlng) {
-          this.onLatLngEntered(latlng[1], latlng[2]);
+          this.onLatLngEntered(latlng[1], latlng[2])
         }
       },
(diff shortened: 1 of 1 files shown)


frontend/js/components/LocationField.vue
```diff
@@ -172,6 +172,11 @@
 
         this.address = newValue
         this.$emit('change', newValue)
+
+        const latlng = newValue.match(/^(-?\d+(?:\.\d+)?),+ *(-?\d+(?:\.\d+)?)$/)
+        if (latlng) {
+          this.onLatLngEntered(latlng[1], latlng[2]);
+        }
```
pboivin, line 179: Nice regexp, it seems pretty robust.

What do you think of doing this check in `onBlur()` instead? I don't think there's an issue in terms of performance with `onInput()`, just asking out of curiosity.

pboivin (commented):
Hi @mattdinthehouse , sorry for the delay on this.

I just tested the PR locally and it works great! I've added a few questions and comments below.

pboivin:
I'm wondering if there's any reason to make this an opt-in rather that change the default behavior. Something like adding an `autoDetectLatLngValue` prop and a corresponding option on the [`map`](https://github.com/area17/twill/blob/2.x/views/partials/form/_map.blade.php) partial.

cc. @ifox

frontend/js/components/LocationField.vue
```diff
@@ -172,6 +172,11 @@
 
         this.address = newValue
         this.$emit('change', newValue)
+
+        const latlng = newValue.match(/^(-?\d+(?:\.\d+)?),+ *(-?\d+(?:\.\d+)?)$/)
+        if (latlng) {
+          this.onLatLngEntered(latlng[1], latlng[2]);
```
pboivin, line 178:
```suggestion
          this.onLatLngEntered(latlng[1], latlng[2])
```

Small fix for eslint

ifox:
Hi guys, this is a nice feature but I agree with @pboivin that it should probably be an opt-in. We never know what Twill users ended up doing with the map field and I wouldn't want to unexpectedly change the behavior for them.

frontend/js/components/LocationField.vue
mattdinthehouse: If it's a blur event then you have to change focus before it triggers and with that method I often find that I'll finish typing and have a confused moment where nothing happens until I remember I need to tab off the input. Using an input event just means it happens immediately without me having to take any extra steps.

What I've got here shouldn't impact performance but we could just debounce it so it only happens once you stop typing, but in my use case we were just copy-pasting lat/lng's into the field so it didn't matter.

mattdinthehouse:
@pboivin @ifox up to you but if it was in the next major release you could flag it as "potentially breaking"? I can't imagine too many users would have implemented a custom lat/lng detector on the frontend themselves.

frontend/js/components/LocationField.vue
pboivin: Thanks for the feedback on this. I think adding a light debounce makes a lot of sense.

pboivin:
I think this could be released sooner in 2.6 if we could add the debounce and the "feature flag". Do you have some time to look into this @mattdinthehouse?

mattdinthehouse:
@pboivin unfortunately I'm slammed for the next few weeks so I won't be able to commit to delivering it on time. I'll check back here once things calm down to see if it still needs doing though.

pboivin:
@mattdinthehouse No worries, thanks for the quick reply :)

PR #1171, opened 2021-09-29
Fix dynamic repeater title / prefix on json Repeaters

chrispymm:
Add in 2 missing lines from getFormFieldsForRepeaters that set the titleField and hideTitlePrefix values to support dynamic repeater titles when used outside the block editor with HandleJsonRepeaters trait


pboivin (approved):
Good catch, thank you @chrispymm !

ifox (approved):

PR #1172, opened 2021-09-30
Created TwillModel trait and TwillModelInterface

OhKannaDuh:
## Description

This PR creates a trait and interface for the twill model functionality. Why? To allow easier integration for more complex systems. For example, you may be using models provided by another third party package, if you wanted to manage these using Twill you'd have to copy/replicate all that functionality within your new models. That's because we can only inherit from one class, in this case either the Twill model or the third party model.

To rememedy this we can take advantage of a trait and interface. That should ensure we can easily integrate Twill when we encounter scenarios like above.

For backwards compatability I've kept everything wrapped within the base model (`A17\Twill\Models\Model`). So previous implementations should still work and those looking to take advantage of third party models can implement the interface and use the trait.

## More on this

Do I think this is a good idea? Honestly I'm not 100% on it, I'm hoping to get some feedback on this and see where it goes, maybe this is the type of thing that should be left for 3.x

Do I think it's useful? Yes, for the most part. I've implemented this into a work project, and why that almost made me feel this being in the core codebase was unnecessary, I ran into static analysis issues since the module repository (`A17\Twill\Repositories\ModuleRepository`) type hints at the base twill model. On this, just the interface if nothing else could also help here.

Let me know what you think 👌 (I've ommited some doc block changes from this PR to avoid drowning out the actual code changes for now)


src/Models/Behaviors/TwillModel.php
```diff
@@ -0,0 +1,166 @@
...
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Relations\MorphToMany;
+use Illuminate\Database\Eloquent\SoftDeletes;
+use Illuminate\Support\Str;
+
+trait TwillModel
+{
+    use SoftDeletes;
+    use HasPresenter;
+    use SoftDeletes;
```
pboivin, line 20:
```suggestion
```
I think the duplicated trait is causing the tests to fail on PHP <= 7.2

pboivin:
Hi @OhKannaDuh, 

This looks good! I like the idea of introducing interfaces for core Twill concepts in `3.x` and beyond. With better use of the Laravel service container, perhaps this could be a good way to provide more extensibility for Twill.

I also agree that dockblocks and typehints can be improved overall.

I'm not 100% convinced about the `TwillModel` trait but it's nice to have this as a proof of concept. I'll play around with this on my end. Hopefully this PR can be a great place for further discussion. :)

OhKannaDuh:
Yes I'd absolutely love to see something like that implemented in the future.

Is there a roadmap for 3.0?

haringsrob:
Hey @OhKannaDuh,

I will be merging this pr into another working branch of mine. I have removed the trait and kept the abstract model implementing the interface.

I will further extend where needed in the other branch.

Thanks a lot for your contribution!

PR #1174, opened 2021-10-01
Handle many-to-many relations in index columns

mattdinthehouse:
## Description

I had a many-to-many relationship that I wanted to print as a column but only single `BelongsTo` relations were supported through the use of `Arr::get($item, "{$column['relationship']}.{$column['field']}");`.

This change adds a check for `BelongsToMany`, plucks the `$column['field']` values, and concatenates them together for printing.

## Related Issues

None


ifox (approved):
Thanks guys! Really appreciate the quick test as well @pboivin.

PR #1178, opened 2021-10-01
feat: Add twill:make:singleton command

pboivin:
## Description

This PR adds a new Artisan command to generate singleton modules.

Singletons are a type of module that are designed to only have a single record in the database. Unlike regular modules, they do not have an index page in the CMS. Typically, the menu entry for a singleton will lead directly to the form of its first and only record.

Singletons a great for one-off pages that require unique fields and/or custom logic (e.g. Home, About, Contact, etc.). An example from [demo.twill.io](https://demo.twill.io/about) : 

demo

Like `twill:make:capsule`, this new command is an extension of `twill:make:module`. Outside of the base command, some other modifications were made to support this feature:

- added a `SingletonModuleController` base class
- added a `Route::singleton()` macro
- added support for a `singleton` parameter for `twill-navigation.php` items
- added Seeder class stub for singleton modules
- added a few helper methods to handle stub file processing (this will be further improved in a future PR)

Some integration tests were also added along the way.

**The command in action:**

make-singleton

## Related Issues

—


ifox (dismissed):
This is amazing @pboivin! Very nice holistic solution for Twill singletons.

Minor nitpick: can the `contactpage` singleton in tests be renamed to `contactPage`?

ifox (approved):

PR #1226, opened 2021-10-30
Add doctrine/dbal 3.0 support

### LINKED ISSUE #818: doctrine/dbal ^3.0 is not supported
```
composer require area17/twill:"^2.0"
./composer.json has been updated
Running composer update area17/twill
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - area17/twill[2.0.0, ..., 2.x-dev] require doctrine/dbal ^2.9 -> found doctrine/dbal[v2.9.0, ..., 2.13.x-dev] but the package is fixed to 3.0.0 (lock file version) by a partial update and that version does not match. Make sure you list it as an argument for the update command.
    - Root composer.json requires area17/twill ^2.0 -> satisfiable by area17/twill[2.0.0, ..., 2.x-dev].

Use the option --with-all-dependencies (-W) to allow upgrades, downgrades and removals for packages currently locked to specific versions.
```

ifox:
## Description

Because [this migration](https://github.com/area17/twill/blob/2.x/migrations/default/2020_02_09_000012_change_locale_column_in_twill_default_fileables.php) uses the `->change()` [helper from Laravel](https://laravel.com/docs/8.x/migrations#prerequisites), Twill needs this dependency. In Twill 3.0 the migrations will be cleaned up to avoid this need, but for now we need to add support for installing `doctrine/dbal` 3+ in host applications.

## Related Issues

Fixes #818

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -69,7 +69,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          composer require "laravel/framework:${{ matrix.laravel }}" "orchestra/testbench:${{ matrix.testbench }}" --no-interaction --no-update
+          composer require "laravel/framework:${{ matrix.laravel }}" "orchestra/testbench:${{ matrix.testbench }}" "doctrine/dbal:${{ matrix.dbal }}" --no-interaction --no-update
           composer update --${{ matrix.stability }} --prefer-dist --no-interaction
 
       - name: Execute tests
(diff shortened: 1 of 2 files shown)


pboivin (approved):
Nice! I agree that Twill 3.x is a good opportunity to refactor this.

Out of curiosity, I took a peak at the Actions tab to see if the change to the test matrix had any impact on the total run time. It doesn't :)

PR #1235, opened 2021-11-03
Add german translations

fabianfetting:
## Description

At the german `lang` file some texts are in english. I replaced it with the german translation.


pboivin (approved):
Thanks @fabianfetting!

Issue #520, opened 2020-01-29
Max chrs counter red by default on fields that have a low limit defined - not good 


arkid:
This is a tiny detail but its also something that would be very quick to change and I think should be.

Basically if I define fields with a max chr limit of say 5, when I create a new record for that module type the form loads and the fields with limits set show this little number on the far right of them.  For the ones with these very low max chr counts, its already showing in red.   

That to me is really negative.  if you think about it in terms of user psychology it's as if an error has already occured just loading the page before I've even typed something.     

Can the logic please be changed so that the red colour only appears when a user is actively typing with the focus inside the input box ?   Or maybe make a simple formula that says that this red starts to occur not on a specific chr count (which I assume is what it does across the board right now) but instead when a percentage figure of the total available chr count is left to type - ie 20/15%).

PR #1237, opened 2021-11-04
fix: Show red input count only if above 90% capacity

pboivin:
## Description

This updates `TextField` and both `WYSIWYG` to display the remaining characters indicator in red only when the count is above 90% of the allowed length.

## Related Issues

Fixes https://github.com/area17/twill/issues/520


ifox:
I see where the issue's OP is coming from however I think the initial thinking was that there was value in showing to the user that they were near the limit. I feel like a nice approach would be to consider the max length value to dynamically change the threshold. Something arbitrary like preserving the current behavior above 50, and this new behavior under that. What do you think?

pboivin:
Makes sense, thanks for the feedback! I'll admit I don't have a very strong opinion on this one. I agree with the OP that the use of red is not optimal if the limit is not exceeded. Perhaps we can take inspiration from Twitter and introduce a "warning" color?
twitter

ifox (approved):

PR #1242, opened 2021-11-06
Ensure autoloading on cached config.

haringsrob:
## Description

This is a bugfix that ensures bootstrapping processes such as autoloading and service provider registration are working when config is cached.

## Some things to consider

We might want to move the call to bootstrapCapsule to CapsuleServiceProvider because now there is a flag on the getCapsuleList function. But this makes less sense when reading the code.

### DIFF
--- a/src/CapsulesServiceProvider.php
+++ b/src/CapsulesServiceProvider.php
@@ -11,6 +11,8 @@ class CapsulesServiceProvider extends RouteServiceProvider
 {
     use HasRoutes, HasCapsules;
 
+    public static $capsulesBootstrapped = false;
+
     protected $manager;
 
     protected function mergeTwillConfig()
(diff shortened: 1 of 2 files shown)


pboivin (commented):
Hi @haringsrob, good catch!

I like the simplicity of your implementation. It makes sense to piggy back on `getCapsuleList()` for this. But I also agree that it would make sense to move some code to `CapsuleServiceProvider`, considering that the boostrapping responsibility belongs to this class.

What do you think of having an explicit `bootstrapCapsules()` step, as part of `register()`?

haringsrob:
Hey @pboivin,

Yes will move this around. 

Thanks for checking.

haringsrob:
Hey @pboivin I have updated the flow of code. Is it ok like this?

pboivin (approved):
Hi @haringsrob, thanks for taking the time to revisit this, it looks great! Over time (probably post 3.0), I'd love to push even more on the refactor of `HasCapsules` responsibilities.

ifox (approved):

PR #1245, opened 2021-11-08
Repeater feature to allow repeatable blocks of varying types

stiknoltz:
## Description

This feature adds the possibility of using an array of block options as repeatable blocks. 

Use in a form blade looks like this: 

```
@formField('repeater_array', [
    'name' => 'machine_name_for_field',
    'types' => [
        [
            'key' => 'machine_name_for_repeater_block_type',
            'name' => 'Name for Block Type',
        ],
        ...
    ],
])
```

## Related Issues

* Originally had a discussion with @ifox and @antonioribeiro around this feature about a year ago. Finally got around to updating the UI as I had a need for a longer list of block options.

### DIFF
--- a/frontend/js/components/RepeaterArray.vue
+++ b/frontend/js/components/RepeaterArray.vue
@@ -1,0 +1,209 @@
+<template>
+  <div class="content">
+    <draggable class="content__content" v-model="blocks" :options="dragOptions">
+      <transition-group name="draggable_list" tag='div'>
+        <div class="content__item" v-for="(block, index) in blocks" :key="block.id">
+          <a17-block :block="block" :index="index" :size="blockSize" :opened="opened" @open="setOpened">
+            <a17-button slot="block-actions" variant="icon" data-action @click="duplicateBlock(index)"  v-if="hasRemainingBlocks"><span v-svg symbol="add"></span></a17-button>
+            <div slot="dropdown-action">
+              <button type="button" @click="collapseAllBlocks()">Collapse All</button>
+              <button type="button" @click="deleteBlock(index)">Delete</button>
+              <button type="button" @click="duplicateBlock(index)" v-if="hasRemainingBlocks">Duplicate</button>
+            </div>
+          </a17-block>
+        </div>
+      </transition-group>
+    </draggable>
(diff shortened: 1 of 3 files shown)


pboivin:
Hi @stiknoltz,

This is cool! I did a quick test on my end and it seems to work well inside the block editor. Am I right in thinking that this can't currently be used outside of the block editor, like a regular repeater?

From this [original Spectrum discussion](https://spectrum.chat/twill/feature-requests/repeater-with-several-type-of-blocks~63cb18c6-df37-400f-87bf-025f6a3cef51), I'd say the primary use case seems equivalent to nested block editors. Does this make sense? I'm curious to know if nested `block_editor` fields could cover this use case. (Not supported at the moment, just thinking out loud)

stiknoltz:
Thanks @pboivin. You're correct that this was intended for use in the block editor context. This feature is probably roughly equivalent to a nested block editor. 

I'm sure thinking about it more broadly could yield some interesting additional possibilities, but for my current needs (and the needs discussed in that Spectrum thread) it seemed to fit the bill. 

I'd be happy to take a look at expanding its application, but I figured since I'd wrapped up the feature for my current needs, I'd submit it back since it seemed to be useful as-built.

haringsrob:
This is a great feature.

I was wondering, can't we make the repeater array the default (so no new type, just update the existing one with the option of having multiple types).

This would be a breaking change but would work well for 3.x?

So it becomes:
```
@formField('repeater', [
    'name' => 'machine_name_for_field',
    'types' => [
        [
            'key' => 'machine_name_for_repeater_block_type',
            'name' => 'Name for Block Type',
        ],
        ...
    ],
])
```

Or we could go for nested block editors and then a repeater becomes obsolete for blocks but can still be used for relational data?

If we do nested block editors then we also have to add some options like a limit of the amount of blocks.

haringsrob:
Hey @stiknoltz,

Thanks again for taking the time to contribute!

After some discussion, we found that a nested block editor would provide the same functionality but with more control already built into it.

There is a wip pr for this here: https://github.com/area17/twill/pull/1397 and we will further focus on improving that.

Kind regards,

PR #1251, opened 2021-11-17
Fixes nested module count pluralisation

freshleafmedia:
## Description

Simplifies the pluralisation of the nested module count. It also fixes an edge case: When the count is 0 the name should be pluralised: `0 things` rather than `0 thing`


ifox (approved):

pboivin:
Thanks @freshleafmedia!

PR #1270, opened 2021-11-24
Fix edit link feature regression in Quill WYSIWYG

joyceverheije:
The combination of 51302a151fdb22e3f5bf0fb99ee21c07c72273af and f8276462fdfb32fe0d425cff7536991078c3b12b created an issue where links could not be edited anymore. It appears that this code is not necessary now that the scroll container is using `html`.


pboivin (approved):
Thanks @joyceverheije, just did a quick test and it works great.

PR #933, opened 2021-05-25
Add dependency resolver to twill:capsule:install

antonioribeiro:
## Description

This PR brings a simple dependency resolver to Capsules installer. It's already possible to test it by running:

``` bash
php artisan twill:capsule:install --copy cities
```

It will install, along with the main Capsule ([Cities](https://github.com/area17/twill-capsule-cities)), also the [Base](https://github.com/area17/twill-capsule-base) and [Countries](https://github.com/area17/twill-capsule-countries) Capsules.

It gets information from a `twill.json` file (not sure if this the best name/format for it) which should be present in the root of a Capsule that has dependencies:

``` json
{
    "dependencies": {
        "capsules": [
            {
                "capsule": "Base"
            },
            {
                "capsule": "Countries",
                "branch": "stable"
            }
        ]
    }
}
```

PR #1291, opened 2021-12-03
Composer Installable Capsules

freshleafmedia:
## Description

This PR adds the ability to install Capsules via composer (eg `composer require freshleafmedia/test-twill-capsule`). It also adds some documentation around capsules.

## Requirements

- The packages `composer.json` needs to declare it's type as `twill-capsule`
- The package must declare a PSR-4 autoloader - This is so we know what namespace the package is using
- The package name must match the Capsule name
- Composer v2

## How it works

Composer is used to install the package like normal to the vendor directory, we could get Composer to install it to the Capsule directory but as was discussed in #979 this is not east, instead we use composers internal `getInstalledPackagesByType()` method to get all installed packages which have been marked as a 'twill-capsule'.

This package is then loaded in almost the same way as a regular capsule, there is just some additional logic to set namespaces correctly etc. This means that manually created and Composer installed Capsules can co-exist

The developer is still required to update the `twill-navigation` config, run migrations etc.

## Example

To test this feature I created a dummy package (https://github.com/freshleafmedia/test-twill-capsule, https://packagist.org/packages/freshleafmedia/test-twill-capsule)

```
composer require freshleafmedia/test-twill-capsule:master-dev
```

## Compatibility Considerations

- I have had to change the signature of the `Manager->capsuleNamespace()` method to accept the whole `$capsule` context but this could possibly be refactored to a separate method `Manager->composerCapsuleNamespace()` if this is a feature we want to pursue

## Related Issues

#979
#933


pboivin:
Hi @freshleafmedia, this looks amazing! I'm trying to give it a spin locally with your freshleafmedia/test-twill-capsule package. I'm wondering, how should I register this in my `config/twill.php` and `twill-navigation.php`?

freshleafmedia:
Hi @pboivin Glad you like the look of it :smiley:

You shouldn't need to add it to the config at all, it's done automatically

src/Services/Capsules/HasCapsules.php
```diff
@@ -65,6 +80,12 @@ public function getCapsulesSubdir()
 
     public function makeCapsule($capsule, $basePath = null): array
     {
+        $capsule['composer'] = $capsule['composer'] ?? false;
+
+        if ($capsule['composer'] === true) {
+            $basePath =  __DIR__ . '/../../../vendor/' . $capsule['fullName'];
```
pboivin, line 86:
```suggestion
            $basePath = InstalledVersions::getInstallPath($capsule['fullName']);
```
If I'm not mistaken, this will be required to support Composer `'type' => 'path'` installations, that can be used in development setups. For example, my own development copy of Twill is installed via git submodule into a Laravel project, and then symlinked into `vendor/area17/twill`. The path used above fails to locate the actual install location of the `vendor` directory in this context.

src/Services/Capsules/HasCapsules.php
```diff
@@ -199,6 +222,10 @@ public function registerServiceProvider($capsule): void
 
     public function capsuleRootPath($capsule)
     {
+        if ($capsule['composer'] === true) {
+            return __DIR__ . '/../../../../../' . $capsule['fullName'];
```
pboivin, line 226:
```suggestion
            return InstalledVersions::getInstallPath($capsule['fullName']);
```
(see comment above)

src/Services/Capsules/Manager.php
```diff
@@ -6,12 +6,20 @@ class Manager
 {
     use HasCapsules;
 
-    function capsuleNamespace($capsuleName, $type = null)
+    function capsuleNamespace($capsule, $type = null)
     {
+        $type = config("twill.capsules.namespaces.{$type}");
+
+        if ($capsule['composer'] === true) {
+            $composer = json_decode(file_get_contents(__DIR__ . '/../../../../../' . $capsule['fullName'] . '/composer.json'), true);
```
pboivin, line 14:
```suggestion
            $installPath = InstalledVersions::getInstallPath($capsule['fullName']);
            $composer = json_decode(file_get_contents($installPath . '/composer.json'), true);
```
(see comment above)

src/Services/Capsules/HasCapsules.php
```diff
@@ -21,7 +22,21 @@ public function getCapsuleList()
...
         if (!config('twill.capsules.loaded')) {
+            $composerCapsules = collect(InstalledVersions::getInstalledPackagesByType('twill-capsule'))
+                ->unique()
+                ->map(function ($fullName): array {
+                    [$vendorName, $packageName] = explode('/', $fullName, 2);
+
+                    return [
+                        'name' => $packageName,
+                        'fullName' => $fullName,
+                        'enabled' => true,
```
pboivin, line 33: I personally like the auto-discovery feature here! Does this mean that a composer capsule can't be disabled manually (or programmatically) via configuration? Curious to know your thoughts on this.

pboivin (commented):
Still testing on my end but it looks pretty good! I've included a few comments below.

`getInstalledPackagesByType()` which you are using and `getInstallPath()` which I'm suggesting have been introduced in [Composer 2.1](https://getcomposer.org/changelog/2.1.0-RC1), which is fairly new (June 2021). Would it make sense to add a check for this and disable the feature if the version requirements are not met? Currently, it will simply fail during `composer install` with a method not found error.

src/Services/Capsules/HasCapsules.php
freshleafmedia: :+1: I wasn't aware of that method. That is much better, IMO lots of `../` is a code smell

src/Services/Capsules/HasCapsules.php
freshleafmedia: Yeah, that's right, they can't currently be disabled. I couldn't think why it would be useful to be able to disable a Capsule when you can just composer remove it.

What is the use-case for disabling capsules?

freshleafmedia:
I've added a check to ensure that Composer 2.1 is in use if you try and use the feature. I opted to throw an exception rather than fail silently so that the developer gets some feedback about why it's not working

Muffinman:
I happen to also be working on this exact requirement currently, but have managed to get it mostly working it without a PR using some inventive config and composer namespace auto-loading. Unfortunately my solution does require all capsules to be contained within a single composer package.

```php

return [
    'capsules' => [
        'namespaces' => [
            'base' => 'Orphans\Twill\Capsules',
            'models' => 'Models',
            'repositories' => 'Repositories',
            'controllers' => 'Http\\Controllers',
            'requests' => 'Http\\Requests',
        ],
        'path' => base_path('./vendor/orphans/orphans-commerce-capsules/src'),
        'list' => [
            [
                'name' => 'Categories',
                'enabled' => true,
            ],
            [
                'name' => 'Products',
                'enabled' => true,
            ],
        ],
    ],
];
```

Package structure:
```
- src/
    - Categories/
        - database 
        - Http
        - Models
        - Repositories
        - resources
        - routes
    - Products/
        - database 
        - Http
        - Models
        - Repositories
        - resources
        - routes
```

Example package composer.json
```json
{
  "name": "orphans/orphans-commerce-capsules",
  "description": "A Twill Capsule to create Products and Categories for your website",
  "autoload": {
    "psr-4": {
      "Orphans\\Twill\\Capsules\\": "src/"
    }
  }
}
```

src/Services/Capsules/HasCapsules.php
pboivin: Good point, `composer remove` is probably the most straightforward way to remove a capsule.

Some use case I can think of for disabling capsules via configuration: debugging, customizing which capsules are available for a given user or context (e.g in a multi-tenant application). There are probably more use cases to uncover. I guess the main point I'm thinking of is: should we provide a way to opt-out of capsule auto-discovery? Similar to what is possible with Laravel packages.

freshleafmedia: I think I agree. How about this, make autodiscovery optional (but enabled by default) and when it is enabled Composer installed capsules can not be disabled. This I believe will cover most use-cases.

Then if more control is needed autodiscovery can be disabled and then the developer is given full control over Composer capsules using the existing Capsule config, eg:

```
    'capsules' => [
        'auto_discover' => false, 
        'list' => [
            [
                'name' => 'test-twill-capsule',
                'composer' => true,
                'enabled' => false,
            ],
        ],
    ],
```

src/Services/Capsules/HasCapsules.php
```diff
@@ -65,6 +81,16 @@ public function getCapsulesSubdir()
 
     public function makeCapsule($capsule, $basePath = null): array
     {
+        $capsule['composer'] = $capsule['composer'] ?? false;
+
+        if ($capsule['composer'] === true) {
+            if (class_exists(InstalledVersions::class) === false) {
```
pboivin, line 87:
```suggestion
            if (class_exists('\Composer\InstalledVersions') === false) {
```
I'm not 100% sure, forgive me if I'm missing something. I'm wondering if the check on `InstalledVersions::class` actually makes sense if the class doesn't exist (ie. Composer < 2.0). Also, I think a check on `method_exists('\Composer\InstalledVersions', 'getInstallPath')` could also be needed to complete the check for Composer 2.1.

On a related note, should this check happen earlier in the process, perhaps in `getCapsuleList()`?

freshleafmedia:
I'm starting to question the wisdom of this approach. It feels like we're re-inventing the wheel, I propose that instead of creating our own custom capsule package type we make use of the Laravels standard for installing and discovering packages. We could make use of a service provider to then register a capsule. This side steps the need for a very modern version of Composer and embraces the mature standard for package installation that Laravel has used for ages.

Would be interested to explore this new approach first and if that fails fall back on this PR.

Thoughts?

Worth noting that @kirkbushell touched on this first: https://github.com/area17/twill/discussions/979#discussioncomment-1300308

ifox:
Hi @freshleafmedia,

first, thank you so much for helping with this, it is very exciting to work on it with the community.

I agree with you on both sides. If possible, I would much prefer a standard Laravel package approach. There's something I really liked about using a custom package type, so I wouldn't mind pursuing what you are doing here either. I am afraid however that, as you said, we'd be reinventing the wheel. Standard Laravel packages bring so many features like configuration and publishing, we should leverage that.

freshleafmedia:
Hi @ifox

No worries, it's awesome to be involved, the community is very easy to collaborate with :smiley:

I totally agree, there is definitely something cool about having your own custom package type but pragmatically it doesn't make sense.

src/Services/Capsules/HasCapsules.php
haringsrob: Perhaps we can reverse the process and introduce a "TwillServiceProvider" that can be published.

In there during the boot/register (honestly I don't know which has what purpose). We can do:

$this->registerCapsule(ExampleCapsuleServiceProvider::class);

haringsrob: This is also much easier to manage as a developer as opposed to registering capsules via config.

freshleafmedia: @haringsrob See the later replies below. We looking more at using the inbuilt Laravel auto discovery to register capsules automatically when they are installed:

> ... I propose that instead of creating our own custom capsule package type we make use of the Laravels standard for installing and discovering packages. We could make use of a service provider to then register a capsule...

haringsrob:
Hey @freshleafmedia,

Closing this in favor of https://github.com/area17/twill/pull/1446 but I will for sure try to cherry pick the documentation updates as they still make sense.

If you want could you have a look at #1446?

Thank you!

PR #1302, opened 2021-12-10
Adds relatedBrowser option to indexColumns

freshleafmedia:
## Description

This PR allows `$indexColumns` to reference a value selected from a browser and stored in the related table:

```
class PostController extends BaseModuleController
{
    protected $indexColumns = [
        'author' => [
            'title' => 'Author',
            'field' => 'name',
            'relatedBrowser' => 'authors',
        ],
    ];
}
```

While this was possible using the presenter that seemed overkill to just show a related value.

## Alternative approach

Instead of having a separate `relatedBrowser` key I considered adding additional logic to the `relationship` type as it is a relationship we're accessing, just a special. It could look something like this:

```
protected $indexColumns = [
    'author' => [
        'title' => 'Author',
        'field' => 'name',
        'relationship' => 'relatedAuthors',
    ],
];
```

### DIFF
--- a/docs/.sections/crud-modules.md
+++ b/docs/.sections/crud-modules.md
@@ -423,6 +423,11 @@ public function hydrate($object, $fields)
             'title' => 'Field title',
             'field' => 'presenterMethod',
             'present' => true,
+        ],
+        'relatedBrowserFieldName' => [ // related browser column
+            'title' => 'Field title',
+            'field' => 'relatedFieldToDisplay',
+            'relatedBrowser' => 'browserName',
         ]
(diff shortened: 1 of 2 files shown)


pboivin:
Hey @freshleafmedia, this is cool! I'll often setup an attribute accessor for this, but I know not everyone is a fan of that.

I like the idea of having a shortcut directly in `$indexColumns` and I like you current solution (`'relatedBrowser' => 'authors'`), more than the alternative approach.

pboivin (approved):

PR #1303, opened 2021-12-10
Fix by key function in settings repository

shawkiIIalasy:
## Description
`byKey` function in the setting repository return incorrect data when there are more setting page that has the same field name .

Problem senario: 
If I have there setting pages A,B, C and I have a field called `site_title` in A and B and C, then the `byKey`  function will return incorrect data.  

## Related Issues
None


pboivin (approved):
Good catch, thanks @shawkiIIalasy!

PR #1308, opened 2021-12-13
Add block getData via helper class.

haringsrob:
## Description

With this patch you can add more data to a block via a helper class.

Example given a view named: **hero_header_with_menu.blade.php**

```php
<?php

namespace App\Twill\Block;

use A17\Twill\Helpers\TwillBlock;

class HeroHeaderWithMenuBlock extends TwillBlock
{
    public function getData(): array
    {
        $image = $this->block->imageAsArray('heading', 'desktop');
        $ctaTarget = $this->block->getRelated('call_to_action_target')->first()->getSlug();

        return [
            parent::getData(),
            'image' => $image,
            'linkSlug' => $ctaTarget,
        ];
    }
}

```


haringsrob:
- [ ] Needs to be added to the previewer as well
- [ ] Configurable namespace?
- [ ] documentation

Some points needed if this moves forward.

ifox:
I love this @haringsrob! It reminds me of how blocks were rendered a long time ago in Twill and I think we were all missing something like this when not working with Transformers. A lot of blocks tend to have a bit of PHP at the top to pull data out of the block model and this feels like a great fit to clean things up moving forward.

haringsrob:
I added this to the preview function as well. Also while in dev mode, now the exception message will be shown.

Screenshot 2021-12-17 at 11 19 21

haringsrob:
Also added some basic documentation

Screenshot 2022-01-03 at 09 07 33

docs/src/block-editor/block-classes.md
```diff
@@ -0,0 +1,27 @@
+---
+pageClass: twill-doc
+---
+
+# Block classes
+
+If you need more control over blocks, their validation or data for rendering you can use a block class.
+
+To do this create a file named after your block. (ex. for `hero_header_with_menu.blade.php` your class will be
```
ifox, line 9: This is really nitpicking but I feel like that example might not be the most relevant as blocks are not necessarily the right tool to build a hero header with a menu. Let's use something like `images_grid` and `ImagesGrid` instead?

frontend/js/mixins/inputFrame.js
```diff
@@ -37,7 +37,10 @@ export default {
 
       Object.keys(this.$store.state.form.errors).forEach((error) => {
         if (error.substr(0, error.indexOf('.')) === errorKeyWithoutLocale) {
-          locales.push(error.substr(error.indexOf('.') + 1, error.length))
+          const cleaned = error.substr(error.indexOf('.') + 1, error.length)
+          if (cleaned.length === 2) {
```
ifox, line 41: Could this be a problem with regional locales like `fr_be`?

src/Http/Controllers/Admin/ModuleController.php
```diff
@@ -1895,8 +1896,8 @@ protected function fireEvent($input = [])
      */
     public function getRepeaterList()
     {
-        return app(BlockCollection::class)->getRepeaterList()->mapWithKeys(function ($repeater) {
-            return [$repeater['name'] => $repeater];
+        return app(BlockCollection::class)->getRepeaterList()->mapWithKeys(function (Block $repeater) {
+            return [$repeater->name => $repeater->toList()];
```
ifox, line 1900: I'm not sure I understand this change. Was the list of repeaters serialized automatically without this and this makes sure we control what ends up in the view through the the `toJson()` above in this file?

frontend/js/mixins/inputFrame.js
haringsrob: True, Fixed this.

src/Http/Controllers/Admin/ModuleController.php
haringsrob: Before this the $repeaters and $blocks were arrays, I changed this to pass around the Block objects instead.

haringsrob:
Closed for https://github.com/area17/twill/pull/1421

PR #1311, opened 2021-12-13
Delete ContactpageRevision.php

liamjcooper:
## Description
Remove duplicate test stub file that has different casing, mitigating filesystem errors


pboivin (approved):
Good catch, thanks @liamjcooper!

PR #1312, opened 2021-12-13
Updated Italian language

AleksanderHideolab:
Additional Italian translation.


lang/it/lang.php
```diff
@@ -10,14 +10,24 @@
     | 5. Run the command "php artisan twill:lang" to sync all lang files.
     */
 
-
 return [
+    'pages' => 'pagina',
+    'posts' => 'posti',
+    'template-settings' => 'Impostazioni',
```
pboivin, line 16: It seems like this is specific to your site/application. If so, it should probably go in your `resources/lang/`. I know that `pages` and `posts` are kind of common but they are not specific to Twill.

AleksanderHideolab: Ok, no problem on my side.

ifox (approved):

PR #1325, opened 2021-12-20
Fix singleton routing for primary and secondary navigation support

ifox:
## Description

When using the new `singleton` route macro, it was not possible to prefix it in a router group so that it could be configured to be accessible from the primary or secondary level of navigation. This PR fixes the issue by refactoring some of the logic in the `module` macro into helpers.

## Related Issues

Related to #1178

### DIFF
--- a/src/RouteServiceProvider.php
+++ b/src/RouteServiceProvider.php
@@ -290,18 +290,12 @@ class RouteServiceProvider extends ServiceProvider
                 );
             }
 
-            $lastRouteGroupName = RouteServiceProvider::lastRouteGroupName();
+            $lastRouteGroupName = RouteServiceProvider::getLastRouteGroupName();
 
-            $groupPrefix = RouteServiceProvider::twillRouteGroupPrefix();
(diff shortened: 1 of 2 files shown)


src/RouteServiceProvider.php
```diff
@@ -29,6 +29,7 @@ class RouteServiceProvider extends ServiceProvider
      */
     public function boot()
     {
+        require_once __DIR__ . '/Helpers/routes_helpers.php';
```
pboivin, line 32: Maybe as part of `3.x`, we could revisit which helpers can be loader earlier in the process (thinking of `register()` instead of `boot()` in `TwillServiceProvider`)

pboivin (dismissed):
Good catch! This works perfectly.

src/RouteServiceProvider.php
ifox: Totally. I decided to create helpers because inside a macro we can't reference class methods from the provider, but maybe a static reference would be best?

pboivin (approved):

PR #1326, opened 2021-12-20
Add `twill:capsule:rename` command

ifox:
## Description

WIP


haringsrob:
Hey @ifox 

Do we still want/need this? I guess it can be useful, but on the other hand, I do not know how often this is needed and it is additional code that needs maintenance as capsules evolve.

PR #1341, opened 2022-01-04
#1324 - Ensure button state changes on toggle when skipCreateModal is…

### LINKED ISSUE #1324: Save Button Text Doesn't Update When skipCreateModal Is Enabled
## Description

When creating a new model the label of the submit button isn't correctly updated when the status toggle is switched if `skipCreateModal` is set to `false`

### Steps to reproduce

1. Set `skipCreateModal` to `false`
2. Click Add new
3. Click the Status toggle changing it from from Draft to Live

### Expected result

The label of the submit button changes from Save as draft to Publish

### Actual result

The button label doesn't change

### Versions

Twill version:  2.x-dev (2523336)
Laravel version: 8.76.1
PHP version: 8.1.0
Database engine: MySQL 5.7.34

haringsrob:
## Description

I don't really understand why that piece of code was there in the first place. But I might be totally missing something.

I have checked both ways of creating and I could not find any issue after my change.

## Related Issues

Fixes #1324


haringsrob:
@ifox check if you can find why the check was there.

ifox:
Checked with @mazeeblanke, there's no apparent reason for this to be there.

PR #1349, opened 2022-01-05
1307 content scheduling improvements

### LINKED ISSUE #1307: Custom language and date formatting
Is it currently not possible to adjust the format of a date when listing the models? For example, the publish date is displayed as "Month Day, Year" (Dec 10, 2021), but I would rather have the German notation "dd.mm.yyyy H:i:s" (10.12.2021 10:00:00).
I am aware that it is possible via an additional function in the model (e.g. getFormattedPublishStartDateAttribute). But unfortunately the publish date is formatted wrong in several areas, for example also when displaying the publish status when editing an entry of the model.

Also, I have not found a way to translate the term "Scheduled" (which appears in various areas in twill) into German. I saw this word is output 1:1 in the Vue app and is not overridable with a custom language, is that correct? 

Bildschirmfoto 2021-12-13 um 09 38 29

haringsrob:
## Description

In this pr:
- We now use the date format defined in twill config for displaying the scheduled publication dates in the content tables
- By default, under the same conditions as we allow to schedule we display the schedule start date (and end date if set)
- Adds some label translations that were not yet translatable

There are some questions I left in code and also wondering if this is generally the correct approach.
@ifox could you have a look?

## Related Issues

Fixes #1307


src/Http/Controllers/Admin/ModuleController.php
```diff
@@ -1182,6 +1182,22 @@ protected function getIndexTableColumns($items)
                 'html' => $column['html'] ?? false,
             ]);
         }
+
+        // @todo: This feels weird, must be a better way? This is copied from how it is approached in the blade file.
+        if ($this->repository->createForPreview([])->isFillable('publish_start_date')) {
+            // @todo: BEFORE MERGE: Is there any reason to use array_push over this?
```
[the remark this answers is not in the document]
haringsrob: while we are on 2.x we can introduce a flag: $showScheduledDatesColumn that we remove in 3.x as it becomes default.

src/Http/Controllers/Admin/ModuleController.php
```diff
@@ -1182,6 +1182,22 @@ protected function getIndexTableColumns($items)
                 'html' => $column['html'] ?? false,
             ]);
         }
+
+        // @todo: This feels weird, must be a better way? This is copied from how it is approached in the blade file.
+        if ($this->repository->createForPreview([])->isFillable('publish_start_date')) {
```
haringsrob, line 1187: You can use `$this->model->newInstance`

haringsrob: or even try $this->isFillable;

views/layouts/main.blade.php
```diff
@@ -82,6 +82,10 @@
             window['{{ config('twill.js_namespace') }}'].twillLocalization = {!! json_encode($twillLocalization) !!};
             window['{{ config('twill.js_namespace') }}'].STORE = {};
             window['{{ config('twill.js_namespace') }}'].STORE.form = {};
+            window['{{ config('twill.js_namespace') }}'].STORE.options = {
```
haringsrob, line 85: Move this to a more generic "globalConfig" store that is always available.

haringsrob: Move this then to main-form.js and main-listing.js

haringsrob: Call this "twillConfig" so we can use this for anything in the future.

haringsrob:
I made it so it is opt-in instead of default to have the column:

```php
class ScheduledContentController extends BaseModuleController
{
    protected $moduleName = 'scheduledContents';

    protected $indexOptions = [
        'includeScheduledInList' => true,
    ];
```

views/layouts/main.blade.php
```diff
@@ -82,6 +82,10 @@
             window['{{ config('twill.js_namespace') }}'].twillLocalization = {!! json_encode($twillLocalization) !!};
             window['{{ config('twill.js_namespace') }}'].STORE = {};
             window['{{ config('twill.js_namespace') }}'].STORE.form = {};
+            window['{{ config('twill.js_namespace') }}'].STORE.config = {
+                publishDate24Hr: {{config('twill.publish_date_24h') ? 'true' : 'false'}},
```
haringsrob, line 86: @ifox as it is in main.blade.php we do not have to set this up in the form or listing as this is globally available all the time.

haringsrob:
This should be ready now.

lang/nl/lang.php
```diff
@@ -127,6 +127,9 @@
     'publisher' => [
         'cancel' => 'Annuleren',
         'publish' => 'Publiceren',
+        'scheduled' => 'Gepland',
+        'expired' => 'Vervallen',
+        'publish-on' => 'Publiceer op',
```
ifox, line 132:
```suggestion
        'publish-on' => 'Publiceer op',
```
```suggestion
```

ifox (changes requested):
Thanks @haringsrob! Is the `publish-on` key used anywhere I missed?

haringsrob:
@ifox looks like it is not :D I will remove it

haringsrob:
@ifox how do you usually deal with lang.csv merge conflicts? 😅

ifox:
I would rebase the branch ignoring it and then regen the CSV before force pushing.

haringsrob:
regen the CSV -> the command can only convert to the files from the csv or am I missing something?

ifox:
There's a `toCsv` flag on the command :)

haringsrob:
@ifox great will try that next time.

Did it manual for now so should be fine now

config/twill.php
```diff
@@ -156,6 +156,16 @@
      */
     'load_default_migrations' => true,
 
+    /*
+    |--------------------------------------------------------------------------
+    | Twill Publish date formatting configuration
```
ifox, line 161: After a more thorough review, I realized those configurations value were purposefully not set because the Vue components are defaulting to a computed format that takes the user locale into account.

frontend/js/store/modules/config.js
```diff
@@ -0,0 +1,17 @@
+const state = {
+  publishDate24Hr: window[process.env.VUE_APP_NAME].STORE.config.publishDate24Hr,
```
ifox, line 2: It doesn't look like this isn't being used in Vue components, so we should be able to remove it.

views/layouts/main.blade.php
ifox, line 86: It doesn't look like this isn't being used in Vue components, so we should be able to remove it.

ifox (changes requested):

config/twill.php
haringsrob: Removed it.

ifox (approved):

PR #1350, opened 2022-01-05
#1332: Performance improvements.

### LINKED ISSUE #1332: Frontend admin panel working slowly with blocks (app has more 20 languages)
I have overloaded in my admin panel when I try to focus on input or change language. 
My usual page has 7-8 blocks. Blocks are different and have 10-40 fields for each block. The application has 24 languages now.  

### Steps to reproduce
1. Create admin panel with 20-25 languages.  
2. Create a page with a block editor and put 10 blocks (each block just has 30 fields). 
3.  Try to switch languages.

### Expected result
Switch works, after a click

### Actual result
Switch works, after clicking with 7-10 **seconds** delay

### Notes from my side
I found Vue's strange component LocaleField. That's components render HTML input for each input. If you put on $formField and your app has 20 languages on the frontend you will have 20 inputs in HTML. 
It's not a big now, but for me it's - 10 blocks * 30 fields * 24 languages = 7200 inputs in HTML handled by Vue's. Of course, my browser will overload. 
From my side, usually, the value of inputs is stored in "State" and changing only the value of input without cloning this input, but I havn't Vue experience and maybe I am wrong. 

### Versions
Twill version: 2.5.2
PHP version: 7.4

haringsrob:
## Description

This pr changes a bit when the frontend renders fields.

For block editor: It will only render the fields if the item is expanded. Freeing up dom space.
For fields: It will only render the active language fields, Freeing up dom space.

This was tested on a page with **100** blocks, each having **102** fields and **8** languages.

After this pr:
Time to interactive (via lighthouse): 3.2 s
Dom size: nodes: 2972, depth: 17, length: 188737

Before this pr (had to reduce the languages to 2 to make it work): 
Time to interactive (via lighthouse): 8.3 s
Dom size: nodes: 45331, depth: 31, length: 410479

So this pr dramatically reduces the size of the page and improved the performance by quite a lot on bigger blocks editors/language combinations.

Please do have a look at this as I don't know why it is there or what side effects it may generate:

https://github.com/area17/twill/compare/2.x...haringsrob:1332-dom-performance?expand=1#diff-4d78aaef72f437ac2e1e4230b2bf68cc575aa52a7ec4798ecf659ef89768f71f

## Related Issues

Fixes #1332


ifox:
Amazing @haringsrob!

Thanks for flagging what to look for. Regarding the `beforeDestroy`, it was meant to remove deleted fields values from the vuex store. Say the user deletes a block or repeater, we want to make sure that the fields it contained are not going to be sent to the backend on save.

It might be that this is not necessary anymore, but worth testing further.

haringsrob:
Ok makes sense, I will test this further.

frontend/js/components/blocks/BlockEditorItem.vue
```diff
@@ -71,7 +71,7 @@
     },
     data () {
       return {
-        visible: true,
+        visible: false,
```
haringsrob, line 74: If less then 4 expandables, we can show them

haringsrob: mounted () {
      // if there are blocks, these should be all collapse by default
      this.$nextTick(function () {
        if (this.blocks(this.editorName) && this.blocks(this.editorName).length > 3) this.collapseAllBlocks()
      })
    }

in blocks.vue

frontend/js/mixins/formStore.js
```diff
@@ -103,12 +103,5 @@ export default {
       // init value with the one present into the component itself
       this.saveIntoStore()
     }
-  },
-  beforeDestroy: function () {
```
pboivin, line 107: Just wanted to echo the comment from @ifox. With this change, if I add a block, fill some fields, then delete the block, is this content still sent to the backend on save?

pboivin (approved):
This is awesome @haringsrob, quite a big gain from a relatively small fix!

One quick question below re. `beforeDestroy()` to make sure I understand correctly.

frontend/js/mixins/formStore.js
haringsrob: Yes I checked, but I will test it once more tomorrow to be sure.

Add/remove edit all should work fine in all languages.

haringsrob:
Having another look at this today it seems to break reference fields, media fields

haringsrob:
I double checked this and resolved some issues.

For all destroy hooks I checked and the db is clean after removing an item.

PR #1359, opened 2022-01-07
#1305: Do not show dialog when content is not changed + ...

### LINKED ISSUE #1305: Submitting via enter doesn't reset the unsaved change warning
## Description

The warning that shows when you attempt to leave the edit screen with unsaved changes will show up if you just focus on a field without changing it.

Suspect all that is needed is to move an event listener from the submit button to the form submit event but haven't looked into this

### Steps to reproduce

1. Edit a model which has regular text inputs
2. Focus on a field
3. Attempt to leave the page

### Expected result

No 'are you sure you want to leave' confirmation

### Actual result

The 'are you sure you want to leave' confirmation shows

### Versions

Twill version: 2.x-dev (9b276c5)
Laravel version: 8.75.0
PHP version: 8.1.0
Database engine: MySQL 5.7.34

haringsrob:
## Description

Currently on any state change there would be a trigger that would make the popup show on page refresh or navigation.

This change is a bit more specific about what changes to look for.

By doing this, we also now know when content is actually changed, and we can use this to display a "content changed" message.

We could also use this for autosaving content if we add this in the future.

## Related Issues

Fixes #1305 

# Todo

If the approach is ok, I still need to make sure the original state is updated when the form is actually submitted.

### DIFF
--- a/frontend/js/components/Publisher.vue
+++ b/frontend/js/components/Publisher.vue
@@ -11,9 +11,6 @@
       <div class="publisher__item" v-if="revisions.length">
         <a href="#" class="publisher__link" @click.prevent="openPreview"><span v-svg symbol="preview"></span><span class="f--link-underlined--o">{{ $trans('publisher.preview') }}</span></a>
       </div>
-      <div class="publisher__item publisher__unsaved-changes" v-if="hasUnsavedChanges">
-        <span v-svg symbol="edit"></span> {{ $trans('publisher.unsaved-changes') }}
-      </div>
       <div class="publisher__item publisher__item--btns">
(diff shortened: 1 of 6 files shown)


frontend/js/components/Publisher.vue
```diff
@@ -11,6 +11,9 @@
       <div class="publisher__item" v-if="revisions.length">
         <a href="#" class="publisher__link" @click.prevent="openPreview"><span v-svg symbol="preview"></span><span class="f--link-underlined--o">{{ $trans('publisher.preview') }}</span></a>
       </div>
+      <div class="publisher__item publisher__unsaved-changes" v-if="hasUnsavedChanges">
```
haringsrob, line 14: Move to separate pr so the base can already move on.

haringsrob:
@ifox  I removed the visual indicator to another pr so this can move forward #1378

PR #1362, opened 2022-01-10
#1203: Add typecasting to select dropdown

### LINKED ISSUE #1203: Select dropdown does not pre-select existing value in Settings when string is used
## Description

When a string is used as the value parameter of a dropdown select box in a settings section, its current value does not get selected on page reload. When using an integer as value, it DOES get selected correctly.

### Steps to reproduce

Create a dropdown select box with string values in a settings section.

### Versions

Twill: 2.5.2
Laravel: 8.64.0

haringsrob:
## Description

This pr adds typecasting to the select options based on the value coming from the form.

Also, it unescapes using lodash so that special characters are allowed. -> This I am not 100% sure is a way to go. Special characters in option values should maybe just be forbidden security-wise? Or we specifically only fix &amp; for ampersands.

What is your view @ifox?

## Related Issues

Fixes #1203

### DIFF
--- a/frontend/js/components/VSelect.vue
+++ b/frontend/js/components/VSelect.vue
@@ -1,25 +1,26 @@
 <template>
   <div class="vselectOuter">
-    <a17-inputframe :error="error" :label="label" :note="note" :size="size" :name="name" :label-for="uniqId" :required="required" :add-new="addNew">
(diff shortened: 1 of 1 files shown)


haringsrob:
Removed so should be fine now.

ifox (approved):

PR #1363, opened 2022-01-10
#1154: Improve preg replace pattern.

### LINKED ISSUE #1077: Repeater title is not parsed when using double quotes
Hi, I am new to Twill and trying to learn. I came across something peculiar. Trying to use a Repeater without the (soon to be deprecated) twill.php registration. views/admin/repeaters/role.blade.php is as follows:

```
@twillRepeaterTitle('Role')

@formField('input', [
    'name' => 'title',
    'label' => 'Title',
    'note' => "Single noun.",
    'maxlength' => 200
])
```

Works. However, when instead using `@twillRepeaterTitle("Role")`, Block.php does not want to parse the 'title' property, it seems. I get Exception: Block role does not exists or the mandatory property 'title' was not found on this block. If you are still using blocks on the twill.php file, please check if the block is present and properly configured.

I presume this is a simple oversight in the parsing method.

### LINKED ISSUE #1154: Block definition with ")" gives extra unwanted characters
When using a block name with a ")" character in it, there is an extra ') printed in the block:

For example, a block definition with:

@twillBlockTitle('Html **(code)**')

results in this:

afbeelding

Renaming the block to e.g @twillBlockTitle('Html **[code]**') is a workaround for now.

Or is it an easter egg you can play around with? 

afbeelding

:)

Versions:

Twill version: 2.51
Laravel version: 8
PHP version: 7

haringsrob:
## Description

This pr improves the preg matching so that the removal of Twill specific blade tags is more consistent.

## Related Issues

fixes #1154
fixes #1077

### DIFF
--- a/src/Services/Blocks/Block.php
+++ b/src/Services/Blocks/Block.php
@@ -231,7 +231,7 @@ class Block
         $bladeProperty = ucfirst($property);
 
         foreach (['twillProp', 'twillBlock', 'twillRepeater'] as $pattern) {
-            preg_match("/@{$pattern}{$bladeProperty}\('(.*)'\)/", $block, $matches);
+            preg_match("/@{$pattern}{$bladeProperty}\((?:'|\")(.*)(?:'|\")\)/", $block, $matches);
 
             if (filled($matches)) {
                 return $matches[1];
--- a/src/Services/Blocks/Block.php
+++ b/src/Services/Blocks/Block.php
@@ -276,7 +276,7 @@ class Block
 
                 return $callback($value, $options);
             }
-        };
+        }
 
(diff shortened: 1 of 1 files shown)


src/Services/Blocks/Block.php
```diff
@@ -378,9 +378,9 @@ public function render()
     public static function removeSpecialBladeTags($contents)
     {
         return preg_replace([
-            "/@twillProp.*\((.*)\)/sU",
```
haringsrob, line 381: Put the pattern part in a const

haringsrob:
Should be fine now.

ifox (approved):

PR #1364, opened 2022-01-10
#1100: Add defaults to store if new.

### LINKED ISSUE #1100: Connected Fields ignored if default is boolean
## Description

When using a radio form field with a boolean as value, and conditionally showing a form field via the connected fields utility, it doesn't show the connected field when adding the block, containing the form fields. Only when selecting the other radio option and then selecting the default again, the connected fields load as expected. If using a string value, this works as expected without having to switch value first.

### Steps to reproduce

Take the following example:

```blade.php
@formField('radios', [
	...
	'default' => true,
	'options' => [
		[
			'value' => true,
			'label' => 'Yes'
		],
		...
	]
])

@component('twill::partials.form.utils._connected_fields', [
	...
	'fieldValues' => true,
])
	@formField('input', [
		...
	])
@endcomponent
```

### Expected results

Our `input` form field should show up when adding the block.

### Workaround

Since we're dealing with booleans, as long as we only need a connected field for the `true` state, we can change the `true` value to any truthy string. This workaround, however, does not work for the `false` state.

### Versions

Twill version: 2.4.0

haringsrob:
## Description

This pr will add the defaults to the store for radios if there is no value yet from the db.

## Related Issues

fixes #1100

### DIFF
--- a/views/partials/form/_radios.blade.php
+++ b/views/partials/form/_radios.blade.php
@@ -43,13 +43,12 @@
 >
     @if($addNew)
         <div slot="addModal">
-            {{-- unset($note, $options, $required, $default, $inline, $addNew, $inModal); --}}
             @partialView(($moduleName ?? null), 'create', ['renderForModal' => true, 'fieldsInModal' => true])
         </div>
     @endif
 </a17-singleselect>
 
-@unless($renderForBlocks || $renderForModal)
+@unless($renderForBlocks || $renderForModal || (!isset($item->$name) && null == $formFieldsValue = getFormFieldsValue($form_fields, $name, $default)))
 @push('vuexStore')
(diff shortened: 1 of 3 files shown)


views/partials/form/_radios.blade.php
```diff
@@ -49,7 +49,7 @@
     @endif
 </a17-singleselect>
 
-@unless($renderForBlocks || $renderForModal || (!isset($item->$name) && null == $formFieldsValue = getFormFieldsValue($form_fields, $name)))
+@unless($renderForBlocks || $renderForModal)
```
haringsrob, line 52: Double check this removal

haringsrob:
@ifox I have reinserted the old behaviour but added an option for the default value.

ifox (approved):

Issue #1160, opened 2021-09-18
Media File Could not be able to deleted


venkataadithan:
## Description

Media File Could not be able to deleted, when with an polymorphic relation association with the soft deleted records..

### Steps to reproduce
1) Add media file to media library
2) Create a new record in any module ex:Blog
3) Associate the media file to the blog.
4) Soft Delete the blog with deleted_at column on.
5) Now try to delete the media file, delete option is not enabled.

### Expected result

Media file should be deleted after checking the soft deleted colum is null or not.

### Actual result

Media File Could not be deleted. 

### Versions

Twill version:  2.3.1
Laravel version: 6.20.34
PHP version: 7.4.23
Database engine: mySql

pboivin:
Hi @venkataadithan, I can confirm this issue, thanks for reporting!

Just to clarify, it seems normal to me that I can't delete an image associated with a soft delete record. I may restore this record eventually and the image should be restored with it. Although the problem also happens when I permanently delete the record. Then, the image also can't be deleted, which is not desirable.

Does this make sense? Is this what you are experiencing?

ifox:
Hi both. This behavior is definitely on purpose, for the reason you identified @pboivin. However, after destroying a soft deleted record from the UI, the relationship to the media should be destroyed as well which would let the user delete the image in the media library.

pboivin:
@ifox I ran a few more tests on my end. When destroying a soft deleted record from the UI, it looks like the corresponding records from the `mediables` table are not deleted with it. Am I missing something obvious?

venkataadithan:
This is the issue here @pboivin both softDelete and Destroy function checks only canDeleteSafely method on mediables or fileables.

This **canDeleteSafely** method didn't check mediable against the deleted records.

PR #1365, opened 2022-01-10
#1160: Allow media to be deleted.

haringsrob:
## Description

This adds a test + force delete check for media.

I wonder:
- [x] Should we avoid $media->delete() when it is still in use?
- [x] the admin.media-library.medias.destroy API returns 200 even when it fails, should we change that?

## Related Issues

Fixes: #1160

### DIFF
--- a/src/Models/Behaviors/HasMedias.php
+++ b/src/Models/Behaviors/HasMedias.php
@@ -3,6 +3,7 @@
 namespace A17\Twill\Models\Behaviors;
 
 use A17\Twill\Models\Media;
+use A17\Twill\Models\Model;
 use Illuminate\Support\Arr;
 use ImageService;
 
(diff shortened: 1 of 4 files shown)


haringsrob:
Added the can delete check in the delete method to prevent code level deletion if in use.

src/Models/Behaviors/HasMedias.php
```diff
@@ -15,6 +16,16 @@ trait HasMedias
         'crop_h',
     ];
 
+    public static function bootHasMedias(): void
```
ifox, line 19: @haringsrob this totally works and is standard in Laravel however I'm wondering if this is the right location in Twill, knowing we have repository level hooks. What do you think?

haringsrob: This is just here to avoid media from being unable to delete when it is force deleted using some other call. (outside of twill).

If I run $myModel->forceDelete(); via my own API or controller. I might not use the repository to do so.

ifox (approved):

PR #1366, opened 2022-01-10
[3.x][POC] #987: Add support for dropdowns in the global navigation.

### LINKED ISSUE #987: Dropdowns in Menu
### Discussed in https://github.com/area17/twill/discussions/980

I think it should be possible to create dropdowns in the admin menu

<div type='discussions-op-text'>

<sup>Originally posted by **marlonbasten** June 16, 2021</sup>
Is there any way to create dropdowns for the admin interface?
</div>

haringsrob:
## Description

This pr introduces the option to make a primary navigation show as a dropdown instead of a bar.

- [ ] The menu blade files are pretty big. We could convert these into blade-x components to remove most of the logic from the blade file into the component class file.

## Related Issues

fixes #987 

Screenshot 2022-01-10 at 15 04 54
Screenshot 2022-01-10 at 15 04 44

### DIFF
--- a/docs/src/getting-started/navigation.md
+++ b/docs/src/getting-started/navigation.md
@@ -69,3 +69,10 @@ Route::group(['prefix' => 'work'], function () {
     Route::module('studios');
 });
 ```
+
+## Submenu
+
+Optionally you can add `'submenu' => true` above a `primary_navigation` which will remove the items from the primary
+navigation bar and make it into a dropdown instead.
+
+Note: When using submenu the top level item will not be clickable.
--- a/frontend/js/main.js
+++ b/frontend/js/main.js
@@ -21,6 +21,7 @@ if (!window[process.env.VUE_APP_NAME]) {
   window[process.env.VUE_APP_NAME] = {}
 }
 window[process.env.VUE_APP_NAME].vheader = new Vue({ el: '#headerUser' })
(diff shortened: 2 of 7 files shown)


ifox:
Sounds great! Can you share a quick screenshot/video of this? It looks like it currently replaces the bar, I wonder if an option to keep both at the same time would be interesting, so that you can quickly jump to another section using the drop-down, but can still use the bar for single click navigation inside the same section.

ifox:
Oh well, you beat me to it for the screenshot, thanks!

haringsrob:
Hey @ifox, yea we can simply keep both as well.

haringsrob:
Screenshot 2022-01-11 at 08 44 48

Made it so that both remain visible now.
Also fixed some bugs :)

haringsrob:
Thinking a bit about menu's in general. Now that we have packages that register menu items as well (but without control). It could be interesting to also make this a bootable thing.

Implementors could say, from their own code:

`TwillMenu::addLink(ModuleName::class)->children([TwillMenu::addLink()...]);`

While packages can do: `$this->addTwillMenuLink()...`

We also introduce a new config `disable_package_menu_registration = false` so that if your want more control over where a package places its menu you can add that to your own code.

Thoughts?

haringsrob:
Closing this to be reimplemented (which will be easier) in 3.x (using alpinejs)

PR #1371, opened 2022-01-13
Support stdclass objects by avoiding array access + small refactor

ifox:


haringsrob:
Hey @ifox was this solving a particular problem? 

I checked and found no regressions so I think this is mergable.

PR #1375, opened 2022-01-14
Update docs

haringsrob:
## Description

While making a blog installation option I am documenting things I run into that might be/are missing.


haringsrob:
@ifox very minor improvements here. Was tags not documented before?

ifox:
Thanks @haringsrob! Surprising, but indeed, the tags field was not documented. It is somewhat limited and suffers from the same issues you identified in a recent issue, but it can still be helpful.

haringsrob:
Ok then all fine :)

ifox (approved):

PR #1376, opened 2022-01-14
[3.x] Blog example

haringsrob:
## Description

This pr adds a basic blog example as an install option.

It can be installed to have a starting point for a site or to play around with twill. It is made so it is extensible in the future with other examples.

It can be installed using `php artisan twill:install blog`


haringsrob:
I added a command to put back the example. The workflow to update examples would be like this:

`php artisan twill:install blog` -> Copies files to the project

-> Change files

`php artisan twill:update-example blog` -> Copies back the files copied earlier to the twill package

New files still need to be manually copied the first time (or create blank placeholders) as we don't know which files belong to the example.

haringsrob:
lets:

- [ ] add global search

haringsrob:
Hey @ifox,

In this commit the browser tests have been refactored to use an installable example:

https://github.com/area17/twill/pull/1376/commits/346d363a606608de776a0ea60ea8731c77967379

if the approach is ok I will do the same for the other test stubs.

haringsrob:
- [ ] Fix this onto 3.x

haringsrob:
All test stubs are now replaced with installable examples.

haringsrob:
Should be ready!

docs/src/getting-started/installation.md
```diff
@@ -33,6 +33,24 @@ Twill's `install` command consists of:
...
+
+If this is your first time using Twill or you just want to experiment with a Twill installation you can use the demo
+setup to quickly get started.
+
+The demo setup is a blog. It comes with:
+
+- A blog module
+- A categories module
+- A blade based frontend
+- 2 example blocks to use in the block builder
```
ifox, line 46:
```suggestion
- 2 example blocks to use in the block editor
```

haringsrob:
> This is a fantastic addition, thanks for the refactoring on top of that @haringsrob!
> 
> Could presets be installed after the twill:install command? Say you already created your superadmin user but now want to install the blog exemple.

No the example install only works when installing fresh as it overwrites files.

PR #1377, opened 2022-01-14
#1356: Add question to generate preview files for modules and blocks.

### LINKED ISSUE #1356: Allow to generate a preview blade file when generating blocks
## Summary

When generating a block using the command `php artisan twill:make:block` add an option parameter or prompt to also generate the preview blade file.

The preview blade file could also contain the demo output of the fields that are in the default block.

Bonus: Do the same for modules?

haringsrob:
## Description

This adds a question to generate a module preview file and a block rendering file:

Screenshot 2022-01-14 at 13 34 29

Screenshot 2022-01-14 at 13 34 54

## Related Issues

fixes #1356

### DIFF
--- a/src/Commands/BlockMake.php
+++ b/src/Commands/BlockMake.php
@@ -54,12 +54,15 @@ class BlockMake extends Command
      */
     public function handle()
     {
+        $generateView = $this->confirm('Should we also generate a view file for rendering the block?');
+
         $this->blockMaker
             ->setCommand($this)
             ->make(
                 $this->argument('name'),
                 $this->argument('base') ?? 'text',
-                $this->argument('icon') ?? 'text'
+                $this->argument('icon') ?? 'text',
+                $generateView
             );
 
         return parent::handle();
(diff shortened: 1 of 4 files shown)


haringsrob:
If we keep this for 3.x we do not have to deal with Laravel < 6 errors.

ifox:
> If we keep this for 3.x we do not have to deal with Laravel < 6 errors.

works for me!

src/Commands/BlockMake.php
```diff
@@ -54,12 +54,15 @@ public function getBlockMaker(): \A17\Twill\Services\Blocks\BlockMaker
      */
     public function handle()
     {
+        $generateView = $this->confirm('Should we also generate a view file for rendering the block?');
```
haringsrob, line 57: @ifox Thinking since this is 3.x only, should we swap this with an option like `--noPreview` to skip preview generation and do it by default if no option is provided?

haringsrob: Let's not do this as decoupled frontends have no use for this.

haringsrob:
Hey @ifox,

Thinking, shouldn't we generate a `site/layouts/block.blade.php` file by default on install (3.x) as you effectively always need it?

PR #1390, opened 2022-01-20
#1381: Add test to check unique + fix.

haringsrob:
## Description

I am not sure if this still solves the problem originally reported in #1309.

But it does fix the regression and #1381.

## Related Issues

fixes #1381


ifox:
I do believe this should still work for @liamjcooper since the need for him was to avoid using the DB facade in favor of the Eloquent model.

ifox (approved):

PR #1399, opened 2022-01-24
Add browser field support to connected fields utility

joyceverheije:
## Description

- When `isBrowser` is true, the connected fields are displayed if the connected browser has selected items
- When `matchEmptyBrowser` is true,  the connected fields are displayed if the browser is empty


haringsrob:
Hey @joyceverheije ,

I found a small issue when the X is clicked and it removed the last item in the browser.

Rest is all fine and thanks for updating docs!

haringsrob (approved):

PR #1405, opened 2022-01-26
Fix exception for missing repository in related browser

### LINKED ISSUE #1404: Empty related browser field in blocks breaks admin UI
## Description

Related browser fields in blocks break when they are empty, causing an exception: `Repository class not found for model 'Related'`.

### Steps to reproduce

- Add related browser field to block

```blade
    @formField('browser', [
        'endpoints' => $linkEndpoints,
        'name' => 'related',
        'label' => 'Page',
    ])
```

- Leave browser field empty
- Save page in admin
- Reload edit screen in admin

### Expected result

Empty browser fields should be handled gracefully.

### Actual result

Exception is thrown: `Repository class not found for model 'Related'`.

### Versions

Twill version:  2.6.0
Laravel version: 8.69
PHP version: 8.0.8
Database engine: MySQL

daun:
## Description

Catch any errors for unfound model repositories in empty block browsers. Alternatively, one could disable the `$block->getRelated($relation)->isNotEmpty()` check — not sure what that does or why it's there.

## Related Issues

Fixes #1404

### DIFF
--- a/src/Repositories/Behaviors/HandleBlocks.php
+++ b/src/Repositories/Behaviors/HandleBlocks.php
@@ -306,8 +306,12 @@ trait HandleBlocks
                     }
                 }
             } else {
-                $relationRepository = $this->getModelRepository($relation);
-                $relatedItems = $relationRepository->get([], ['id' => $ids], [], -1);
+                try {
+                    $relationRepository = $this->getModelRepository($relation);
+                    $relatedItems = $relationRepository->get([], ['id' => $ids], [], -1);
+                } catch (\Throwable $th) {
+                    $relatedItems = collect();
+                }
                 $sortedRelatedItems = array_flip($ids);
 
                 foreach ($relatedItems as $item) {


ifox:
Thanks for this @daun!

The check is most likely there from a backwards compatibility intention. When we introduced this feature, there might have been use cases where the table was created but the data was not yet stored in it, so its fallbacking to a regular module/capsule repository resolution, which obviously doesn't work if the relation name doesn't match a repository in your codebase. Was that your case?

daun:
@ifox It's a new(ish) installation that had the related table from the start. The problem is that the `isNotEmpty` check means it will fall back to the regular repository resolution even though the table exists and the query returns a result set (albeit empty). Which is fine except that the regular resolution mechanism expects the repository to actually exist. Catching the exception here bridges the gap between these two implementations and should solve this particular bug. Maybe there's a more elegant solution but none that I could see at a glance.

PR #1411, opened 2022-02-01
Refactored translatable rules functionality

domihagen:
## Description

I have refactored the translatable rules functionality to add some features and fix a bug.

- Not published locales now getting validated (required rules will be disabled)
- Now supports defining rules for fields as Array (like [Laravel](https://laravel.com/docs/8.x/validation#quick-writing-the-validation-logic) it does)
- Fixed bug in the If-Condition where it checks for `'required_'` (haystack and needle were interchanged)

We came across this when we received a database error because it tried to save a text that was too long. Unfortunately, the validation was not performed because the language was not (yet) published.

### DIFF
--- a/src/Http/Requests/Admin/Request.php
+++ b/src/Http/Requests/Admin/Request.php
@@ -47,7 +47,7 @@ abstract class Request extends FormRequest
         if ($this->request->has('languages')) {
             foreach ($locales as $locale) {
                 $language = Collection::make($this->request->get('languages'))->where('value', $locale)->first();
-                $currentLocaleActive = $language['published'];
+                $currentLocaleActive = $language['published'] ?? false;
                 $rules = $this->updateRules($rules, $fields, $locale, $currentLocaleActive);
 
                 if ($currentLocaleActive) {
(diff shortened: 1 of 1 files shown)


haringsrob:
Hey @domihagen,

Thank you for the pull request! I have checked it and it seems to work fine.

This would however introduce a backward compatibility issue. For this I think we should have a switch on the Request, something like:

```php
protected $validateInactiveLanguage = true; // True is the default as this is the current behavior.
```

What do you think? Also, what are your thoughts @ifox?

haringsrob:
Hey @domihagen,

Sure, and please correct me if I am wrong.

Currently, the validation will check for inactive languages as well, this is the current design and users might rely on this behavior. 

Once this pr is merged, that changes as the inactive languages are no longer validated. Which might be unwanted by users relying on that behavior.

haringsrob:
hey @domihagen,

I did a check on 2.x this morning, the video illustrates the current behavior:

Screen Recording 2022-02-11 at 08 31 42

Sorry if I am still missing the point.

haringsrob:
@domihagen Ok, I got it now completely! Thank you for your patience. Looks good to me.

ifox (approved):

PR #1412, opened 2022-02-02
fix: Get image size from uploaded file instead of stored file

pboivin:
## Description

This is very much an edge case... In the event that the Media Library is configured with a `local` endpoint and a remote disk (eg. `s3`), this ensures that the image size is retrieved from the temporary uploaded file, before it is stored on the remote location.

## Related Issues

—


pboivin:
Side note — I'm currently experimenting with this configuration to enable S3 uploads from an EC2 instance throught IAM role authentication, instead of using access keys.

ifox (approved):

PR #1413, opened 2022-02-02
[3.x] Setup psr2 + larastan

haringsrob:
## Description

- [x] Change this to phpstan only


haringsrob:
do we run phpstan from our pipeline? I guess we want to do that. Also I dont know how styleci works about issues that cannot be autofixed?

PR #593, opened 2020-03-17
provide env variable for configuring custom s3 hosts

schaschjan:

PR #1414, opened 2022-02-03
Fix custom S3 endpoints

ifox:
## Description

This PR is addressing the inconsistencies with Laravel's S3 disks configuration, like the support for path style endpoint as discussed in #890. It uses the configuration from the provided disk and instead of creating the endpoint string manually, it leverages the `PostObjectV4` class from the AWS SDK which happens to build exactly the URL we need in the case of uploads posted from the frontend.

This PR also prepares for Flysystem changes needed for Laravel 9 support (the client is not exposed by the adapter anymore in Flysystem 2+)

## Related Issues

#890 
#703 
#593 
https://github.com/area17/twill/commit/5894ccce4c4a8dd0e60b3583c6fa3069d1119640#r40789394


haringsrob:
Merged in the Laravel 9 branch

PR #1360, opened 2022-01-07
[3.x] Blade x components instead of custom directive

haringsrob:
## Description

This is not finished, backwards compatibility has not been implemented.

The idea here is to, in addition to @formField we also have the option to use blade-x components.

So this:

```
@formField('checkbox', [
    'name' => 'checkbox',
    'label' => 'Checkbox',
    'note' => 'I am a note',
])
```

would become this:

```
<x-twill-checkbox name="checkbox" :label="__('This is the label')" :note="__('Note')" :form="$form" />
```

Each form field would have its own blade-x component.

By doing this we can also free the partials from additional PHP.

(there is not much abstraction done yet so I think in a final product the component class will have fewer data.)

Todo:
- [x] Tags: examples/blog/resources/views/twill/blogs/form.blade.php
- [x] select_permissions: views/groups/form.blade.php
- [x] Nest the views in a "Fields" directory.

# Fixes

fixes #969

PR #1419, opened 2022-02-04
[3.x] allow super admins to reset 2fa for other stuck users

YasienDwieb:
## Description

Introduces the ability for super admins to reset 2FA for other users who may get stuck and unable to login in

Admins are required to solve a challenge before disabling 2FA for other users in order to avoid doing that by mistake


haringsrob:
Hey @YasienDwieb !

I checked and this works fine. However, I think we should maybe limit the ability to disable it to admins that have 2fa themself.

What are your thoughts @ifox?

YasienDwieb:
@haringsrob @ifox I have add privilege check and 2fa check to the request validation, please check now

Thanks

YasienDwieb:
@haringsrob @ifox I have add privilege check and 2fa check to the request validation, please check now

Thanks

AndreSchwarzer:
IMHO the admin should also confirm the action by reentering his own password since it's a very sensitive action.

https://laravel.com/docs/9.x/authentication#password-confirmation

https://laracasts.com/series/whats-new-in-laravel-6/episodes/5

YasienDwieb:
@AndreSchwarzer I don't think we would need this, thinking about doing the same on AWS for example you will not be required to do so as you have already passed through all required authentication/authorization steps

haringsrob:
I can finalize and merge this once #1360 is done

- [ ] Replace the @formField with the new components.

haringsrob:
Hey @YasienDwieb,

This is now fine, I will merge it soon.

I also included some small bug fixes that I myself introduced (just in case you wonder about the non related changes).

PR #1431, opened 2022-02-08
Update repeaters to work with active states.

haringsrob:
## Description

This pr introduces some changes:

- When a new repeater is saved multiple times, we keep track of frontend/backend ids to not delete/resave so the IDs do no longer pile up.
- The active state is taken from the parent model, this also automatically publishes/unpublishes the slugs
- Blocks and childblocks are now updating instead of removing/creating.

Repeaters will not work nicely if it would be a module itself. As the active/inactive state of a translation might "desync"
Repeater activates the translation, while it might be disabled from the module's page itself.

There is no easy fix for this but to have the active flag (like in the sidebar) on the repeater itself.

(this problem is not new with this pr but has always existed)

Below is something I found out as well, but not really an issue.

When you remove the repeater field from the model and save, the values are erased in the database. -> I guess we can fix this to leave them untouched.

### DIFF
--- a/src/Repositories/Behaviors/HandleRepeaters.php
+++ b/src/Repositories/Behaviors/HandleRepeaters.php
@@ -2,6 +2,7 @@
 
 namespace A17\Twill\Repositories\Behaviors;
 
+use A17\Twill\Facades\TwillUtil;
 use A17\Twill\Services\Blocks\BlockCollection;
 use Carbon\Carbon;
 use Illuminate\Support\Arr;
@@ -39,7 +40,13 @@ trait HandleRepeaters
     public function afterSaveHandleRepeaters($object, $fields)
     {
         foreach ($this->getRepeaters() as $repeater) {
-            $this->updateRepeater($object, $fields, $repeater['relation'], $repeater['model'], $repeater['repeaterName']);
+            $this->updateRepeater(
+                $object,
+                $fields,
+                $repeater['relation'],
+                $repeater['model'],
+                $repeater['repeaterName']
+            );
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/Repositories/Behaviors/HandleRepeaters.php
```diff
@@ -253,13 +301,17 @@ public function getFormFieldsForRepeater($object, $fields, $relation, $modelOrRe
...
             if (isset($relatedItemFormFields['medias'])) {
                 if (config('twill.media_library.translated_form_fields', false)) {
-                    Collection::make($relatedItemFormFields['medias'])->each(function ($rolesWithMedias, $locale) use (&$repeatersMedias, $relation, $relationItem) {
-                        $repeatersMedias[] = Collection::make($rolesWithMedias)->mapWithKeys(function ($medias, $role) use ($locale, $relation, $relationItem) {
-                            return [
-                                "blocks[$relation-$relationItem->id][$role][$locale]" => $medias,
-                            ];
-                        })->toArray();
-                    });
+                    Collection::make($relatedItemFormFields['medias'])->each(
```
haringsrob, line 304: nothing changed here but reformatting

src/Repositories/Behaviors/HandleRepeaters.php
```diff
@@ -268,13 +320,17 @@ public function getFormFieldsForRepeater($object, $fields, $relation, $modelOrRe
...
 
             if (isset($relatedItemFormFields['files'])) {
-                Collection::make($relatedItemFormFields['files'])->each(function ($rolesWithFiles, $locale) use (&$repeatersFiles, $relation, $relationItem) {
-                    $repeatersFiles[] = Collection::make($rolesWithFiles)->mapWithKeys(function ($files, $role) use ($locale, $relation, $relationItem) {
-                        return [
-                            "blocks[$relation-$relationItem->id][$role][$locale]" => $files,
-                        ];
-                    })->toArray();
-                });
+                Collection::make($relatedItemFormFields['files'])->each(
```
haringsrob, line 323: Nothing changed here but reformatting

haringsrob:
@ifox can we add this to 2.8 beta?

PR #1443, opened 2022-02-11
Added $note in _medias.blade.php, in the case of a translated field

joecritch:
## Description

Added `$note` to translated case inside of `_medias.blade.php` as it was missing.

## Related Issues

None


haringsrob:
Thanks a lot, @joecritch, this was indeed missing.

I checked the documentation and this is already documented so for that we are fine as well.

While checking I found that the fieldNote is not implemented either. So I added a commit for that as well.

ifox (approved):

PR #1449, opened 2022-02-15
Fix unauthenticated user 500 error

raymondtri:
## Description

Currently, Twill aliases the \Illuminate\Auth\Middleware\Authenticate::class as 'twill_auth'.

The issue with this is that \Illuminate\Auth\Middleware\Authenticate's redirectTo function returns nothing, so it defaults to the Laravel default of the "login" named route as where it should redirect.

The issue with that, is that there is no login named route in Twill admin dashboard, instead there is 'admin.login.form'.

This PR creates a new Authenticate middleware in Twill that extends \Illuminate\Auth\Middleware\Authenticate and specifies the redirectTo function route correctly as 'admin.login.form'. This PR also modifies the routeserviceprovider so that it uses the new extended Authenticate class instead of the core Illuminate authenticate middleware.

## Related Issues

None I am aware of

### DIFF
--- a/src/Exceptions/Handler.php
+++ b/src/Exceptions/Handler.php
@@ -21,9 +21,11 @@ class Handler extends ExceptionHandler
      */
     protected function unauthenticated($request, AuthenticationException $exception)
     {
-        return $request->expectsJson()
-            ? response()->json(['message' => $exception->getMessage()], 401)
-            : redirect()->guest($exception->redirectTo() ?? route('admin.login', Route::current()->parameters()));
+        if ($request->expectsJson()) {
+            return response()->json(['message' => $exception->getMessage()], 401);
+        }
+
+        return parent::unauthenticated($request, $exception);
     }
 
     /**
--- a/src/Exceptions/Handler.php
+++ b/src/Exceptions/Handler.php
@@ -13,22 +13,6 @@ use Symfony\Component\HttpKernel\Exception\HttpExceptionInterface;
 class Handler extends ExceptionHandler
 {
     /**
-     * Convert an authentication exception into a response.
-     *
-     * @param  \Illuminate\Http\Request  $request
-     * @param  \Illuminate\Auth\AuthenticationException  $exception
-     * @return \Symfony\Component\HttpFoundation\Response
-     */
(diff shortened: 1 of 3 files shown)


haringsrob:
Hey @raymondtri,

Are you using the twill auth guard outside of twill?

Could you provide steps to reproduce the error 500?

Kind regards,

raymondtri:
No I'm not.

To reproduce the error 500:

1. Be logged out
2. Navigate to a twill admin page (admin.yoursite.com/settings/basic) via direct link

haringsrob:
Hey @raymondtri,

On a clean Laravel installation, this seems to work just fine.

Is there anything specific about your setup? Have you tried to reproduce this on a fresh installation?

Kind regards

raymondtri:
Have you tried this on a fresh Laravel installation without the Auth::routes() being generated? I cannot understand how it would work given the fact that per [here](https://github.com/area17/twill/blob/2.x/routes/auth.php#L6) there is no GET route named login, so how would it know where to route?

haringsrob:
Hey @raymondtri,

Could you check if you have `bind_exception_handler` in the twill config set to false?

The Twill exception handler normally takes care of this redirect.

raymondtri:
Ahhhhhhhhhh. That would be where the issue is. We modified the twill exception handler. Why would the redirect code be buried in an exception handler instead of just redirecting to the correct path in the first place?

ifox:
@raymondtri you're right, I think your change may still be relevant. The handler is attached to all Twill controllers, and I think that if you went as far as disabling that, you can control your Twill middleware stack as well, but you still make a good point. We'll look into this.

haringsrob:
Hey @ifox and @raymondtri,

I have tested today and with the change proposed we can actually remove that unauthenticated exception handling entirely while still having the same result.

ifox:
Makes sense @haringsrob!

Can we make sure this is working fine with the `moduleShowWithPreview` router macro?

haringsrob:
@ifox I have tested and I did not run into any issues.

raymondtri:
Thank you all!

PR #1460, opened 2022-02-21
Fix morphMany repeater update

### LINKED ISSUE #1459: MorphMany repeaters not saving
## Description

The `updateRepeaterMorphMany` function does not work as expected. The `id` and `type` columns are not set during the creation thus throw an SQL error when saving the form.

### Steps to reproduce

I have an `Alias` model that is attached to multiple models through a polymorphic relationship, and an `Element` model that can have many aliases.

**Element model:**
```php
<?php

namespace App\Models;

use A17\Twill\Models\Model;

class Element extends Model
{
    public function aliases()
    {
        return $this->morphMany(Alias::class, 'aliasable')->orderBy('position');
    }
}
```

**Alias model:**
```php
<?php

namespace App\Models;

use A17\Twill\Models\Behaviors\HasTranslation;
use A17\Twill\Models\Model;

class Alias extends Model
{
    use HasTranslation;

    /**
     * The attributes that are mass assignable.
     *
     * @var array
     */
    protected $fillable = [
        'aliasable_id',
        'aliasable_type',
        'position',
    ];

    /**
     * The attributes that are translatable.
     *
     * @var array
     */
    public $translatedAttributes = [
        'name',
    ];
}
```

**Element form:**
```blade
@extends('twill::layouts.form')

@section('contentFields')
    ...

    @formField('repeater', ['type' => 'alias'])
@stop
```

**Alias repeater form:**
```blade
@twillRepeaterTitle('Alias')
@twillRepeaterTrigger('Add new alias')
@twillRepeaterTitleField('name', ['hidePrefix' => true])
@twillRepeaterGroup('app')

@formField('input', [
    'name' => 'name',
    'label' => 'Titre',
    'required' => true,
    'translated' => true,
])
```

**ElementRepository:**
```php
<?php

namespace App\Repositories;

use A17\Twill\Repositories\Behaviors\HandleRepeaters;
use A17\Twill\Repositories\Behaviors\HandleRevisions;
use A17\Twill\Repositories\Behaviors\HandleSlugs;
use A17\Twill\Repositories\Behaviors\HandleTranslations;
use A17\Twill\Repositories\ModuleRepository;
use App\Models\Element;

class ElementRepository extends ModuleRepository
{
    use HandleTranslations, HandleSlugs, HandleRevisions, HandleRepeaters;

    public function __construct(Element $model)
    {
        $this->model = $model;
    }

    public function afterSave($object, $fields)
    {
        $this->updateRepeaterMorphMany($object, $fields, 'aliases', 'aliasable', 'Alias', 'alias');
        parent::afterSave($object, $fields);
    }

    public function getFormFields($object)
    {
        $fields = parent::getFormFields($object);
        $fields = $this->getFormFieldsForRepeater($object, $fields, 'aliases', 'Alias', 'alias');

        return $fields;
    }
}
```

### Expected result

After hitting the Save button, the values of all aliases should be saved and attached to the Element model.

### Actual result

The following error is thrown:

```
[2022-02-21 17:44:39] local.ERROR: SQLSTATE[HY000]: General error: 1364 Field 'aliasable_id' doesn't have a default value (SQL: insert into `aliases` (`position`, `updated_at`, `created_at`) values (3, 2022-02-21 17:44:39, 2022-02-21 17:44:39))
```

### Versions

Twill version: 2.6.0
Laravel version: 8.83.1
PHP version: 8.0.15
Database engine: MySQL 8.0.27

marcbelletre:
## Description

This PR fixes a bug when saving repeaters using the `updateRepeaterMorphMany()` method.

## Related Issues

Fixes #1459


ifox:
Thanks for this @marcbelletre!

cc @daun, do you have any input on this? Asking because you contributed this originally. Thanks!

daun:
I haven't run into the mentioned issue so far. Do we have tests covering the method?

haringsrob:
@daun could it be that your morphs are nullable?

haringsrob:
So after testing a bit, the fix seems to make sense, however, we could then get rid of the line 
`$object->$relation()->save($newRelation);` as it is redundant.

The error only happens when the morph is mandatory.

ifox:
Thanks @daun! @haringsrob let's check if removing the relation save doesn't break the code following that that is soft deleting non existing repeaters by looking at the relation again.

haringsrob:
Hey @ifox ,

This will need a bit more attention, in the past I updated `updateRepeater` to not recreate.

I guess we have to do the same for `updateRepeaterMorphMany`

Also I do not really see how `updateRepeaterMany` would be used? Especially the keepexisting property.

ifox:
Hey @haringsrob, `updateRepeaterMany` is used with pivot tables and `belongsToMany` relationships. It serves the same use case as `updateRepeaterMorphMany` except it doesn't store the relationship in a single table, it needs a table for each module using it.

Example: projects, publications, contributors

With `updateRepeaterMany`, you would need `contributor_project` and `contributor_publication` tables, while with the morph version you can use a single table.

ifox:
I'll have to dig the use case for `keepExisting` but I believe it would be used by a caller wanting to add new records to the relationships while not being able to pass the existing down for some reason.

haringsrob:
I double-checked and all seems fine.

Perhaps we did not have this issue before as the morph columns were nullable.

PR #1461, opened 2022-02-22
Fix: repeater inside a block has wrong expanding behavior

amargoCactus:
fixed the bug that when the add item button or the expand item button is clicked, all the items are expanded, this is especially uncomfortable in repeaters with a lot of items since the end user has to scroll down to get to the item that was just added, this also caused lag for a brief moment


haringsrob (approved):

haringsrob:
Works perfectly! Thank you

Issue #89, opened 2018-11-19
Part of the Block Editor's dropdown menu will be hidden and unable to scroll in some context.


yanhao-li:
Under some context (for instance, very small height window), when clicked the Add Content button, the dropdown menu may be displayed underneath the button, there will be chances that part of the menu's options area hidden and unable to scroll to reveal them.

ifox:
@mrdoinel @m4n1ok very small height window is not necessarily something we need to solve, however this came from a client using Firefox 63 on Windows.

In the following screenshot, you'll notice the publisher module alignments are quite off, and the add content dropdown is not using available space on the top and prevents users from selecting blocks at the end of the list.

![firefox_63_windows_ui_issues](https://user-images.githubusercontent.com/708498/48738111-f1383700-ec1d-11e8-9410-eabca74a270c.png)

haringsrob:
Closing as per follow up of https://github.com/area17/twill/pull/1464

PR #1464, opened 2022-02-22
[2.x] Make 'Add Block' scrollable

domihagen:
## Description

This PR sets a max-height for the 'Add Block' Dropdown to make it scrollable.

Before:
add_block_before

After:
add_block_after


haringsrob:
Hey @ifox I think this is a great improvement. Does this need to go past design?

haringsrob:
This is related: https://github.com/area17/twill/issues/89

There were commits in the past as well

ifox:
I'm going to close this PR as we already have a 430px max height set on the scroller element of this dropdown. A default of 270px would be too small. If you are not able to scroll this dropdown, can you please submit an issue that describes your environment? Thank you!

ifox:
Oooh, I misinterpreted this, my bad. I would then recommend to port the configuration of the one that does scroll to the one that doesn't, not to introduce a smaller max height.

domihagen:
@ifox PR is updated. Changed the max-height to 430px like it is for the oder Dropdown.

PR #1479, opened 2022-02-27
[3.x] Limit revisions on model

### LINKED ISSUE #1478: Max revisions
## Summary

[Inspiration](https://github.com/area17/twill/discussions/917)

Like it was discussed problem happens when a lot of revisions happen, and eventually, DB fills up with old records and takes a lot of space.

Two ideas were presented by @domihagen and @ifox which were using Quoques and the new config option `max_revisions`.
Both ideas are implemented with the addition of custom revisions tables.

kallefrombosnia:
## Description

This PR gives the ability to remove a specific slice of revision records if they are not included in the count of `max_revisions`.
Basically, we keep an `x` number of new records on a specific module, older records get deleted.

### Configuration
```php
'max_revisions' => {type},
```

`max_revisions` have 3 expected values: `false`, `integer` and `true`.
If something else is given it throws `Exception`.

**false**
False is the default value in the config, and if present no actions are taken.

**number**
If number values are present, this int value is used for every module which has revisions on it (globally).

**true**
If this bool value is present, the job will look into other config value `revisions`.

Revisions are an array of key => value items where is `key` is table name and value is type integer.
If the value is a non-integer, an exception is thrown.

They are used to scope specific modules and make custom configurations for them.

Example:
```php
  'max_revisions' => true,
  'revisions' => [
      'categories' => 5, //keep last 5 records on 'categories'
      'articles' => 15, //keep last 15 records on 'articles'
      'dummymodule' => 1 //keep last record only on 'dummymodule'
  ]
```
## Related Issues

Fixes #1478


kallefrombosnia:
@ifox Like I have written to you on Discord, the problem here happens when Twill waits for the job to finish before it saves revisions. It's synchronous, but I wanted to make it asynchronous since it takes a visible time period before it returns a response to the user on the front end.
But, if we use `database`, the core thing won't work out of the box, additional setup will be required from the developer (setting ENV var and migration of jobs table).
Is that desirable for the project? (will need to document it if yes)

pboivin:
Hi @kalle,

Can you double-check the source & destination branch for this PR? I'm having trouble zooming in on the changes here.

I think this idea is a welcome improvement and encapsulating this in a Job seems like the right approach to me. Although, I believe the choice of sync/async should be left to the developper.

You can find an example of this in [`HandleNesting`](https://github.com/area17/twill/blob/2.x/src/Repositories/Behaviors/HandleNesting.php#L14). It's set to use Laravel's `default` queue out of the box, which is synchronous and does not need a database table. This can be customized as needed on the repository.

Not saying that this is the best implementation but I do think that worrying about sync/async may be too much too soon for a newcomer, especially someone learning Laravel _with_ Twill.

Cheers,

P.

kallefrombosnia:
Hi @pboivin, <strike>I will close this PR and reopen it with the source branch `3.x`</strike> since like ifox [recommended](https://discord.com/channels/811936425858695198/811986414152450108/947621290964033546) me to is to change the target branch, but this additional content is from `2.x` on which tests fail since its using some functionality from P8. My mistake.

Regarding this async/sync approach, yes this is something that I should not be *worried* about, but this could be a new thing in the Twill approach since in the future developers could probably have use of it. 
Where I see possible implementation with Twill (workers): scheduled content actions, chores on Twill app, etc
All of the mentioned usages are doable on the Laravel level without the need for Twill, but I'm basing this on the future of Twill since probably like all other CMS Twill will need content scheduling in the core (again my biased opinion).

Regarding `HandleNesting` job implementation, it's understandable that the task which was made by you is synchronous since that is expected to be resolved before returning a response to the frontend. In my usage it works that way also, but it creates not needed time frame since nothing will change if it's executed in default synchronous mode.
In fact, I followed your example on this to create this PR and your code from discussion #917 and other participants' ideas on the implementation of this job. Thank you for that.

Sorry if I sound off subject sometimes, but my background is mostly JS (nodejs) whose main concern is the execution of the application (async/sync), so I take some fears here. And don't take my intentions wrongfully, I just want to help. 😃  

PS. @ifox will squash commits and changes will be clearly visible

pboivin:
Hey @kallefrombosnia, I'm sorry, I think this came out wrong. I did not mean to direct the last paragraph at you! What I meant to say is — From the perspective of someone starting with Twill and Laravel, it should not be necessary to setup an async queue with workers. I believe "sync" is a good default.

Anyhow, this is good work! It makes a lot of sense to introduce Jobs into Twill 👍

config/twill.php
```diff
@@ -217,6 +217,14 @@
     */
     'auto_seed_singletons' => false,
 
+    /*
+    |--------------------------------------------------------------------------
+    | Revisions configuration
+    |--------------------------------------------------------------------------
+    */
+    'max_revisions' => false,
```
haringsrob, line 225: I am wondering if we can change this.

We could use:

```php
twill.revisions.global_limit = null|int;
twill.revisions.module_limit = [
  'module' => 'limit'
];
```

Where the more specified always takes over?

src/Jobs/RevisionsKeepLimit.php
```diff
@@ -0,0 +1,126 @@
...
+
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Bus\Dispatchable;
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Support\Facades\DB;
+use A17\Twill\Models\Model;
+use Exception;
+
+class RevisionsKeepLimit implements ShouldQueue
```
haringsrob, line 14: I think a better name is CleanupRevisions

src/Jobs/RevisionsKeepLimit.php
```diff
@@ -0,0 +1,126 @@
...
+    {
+        $this->modelClass = get_class($model);
+        $this->id = $id;
+    }
+
+    public function handle()
+    {
+
+        // Check if revision limits are enabled
+        if($this->checkIfLimitEnabled()){
```
haringsrob, line 32: I think you can remove some of your inline comments. The method names are often descriptive enough of what is happening.

src/Jobs/RevisionsKeepLimit.php
```diff
@@ -0,0 +1,126 @@
...
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Support\Facades\DB;
+use A17\Twill\Models\Model;
+use Exception;
+
+class RevisionsKeepLimit implements ShouldQueue
+{
+    use Dispatchable, InteractsWithQueue, Queueable;
+
+    protected $modelClass;
```
haringsrob, line 18: Since 3.x is on php 8 we can use typecasting everywhere.

src/Jobs/RevisionsKeepLimit.php
```diff
@@ -0,0 +1,126 @@
...
+                $table = $this->getTableName();
+                $revisions = config('twill.revisions', []);
+
+                // Check if this models table is used on revisions limit list
+                if(array_key_exists($table, $revisions)){
+
+                    
+                    // Check if revisions table name value contains valid integer type
+                    if(!is_int($revisions[$table])){
+                        throw new Exception("twill.revisions.{$table} accepts type integer. Provided: " . gettype($revisions[$table]));
```
haringsrob, line 63: Perhaps we can introduce a custom exception here

haringsrob: Actually wondering at all if we should care about this. We can just do nothing if it is not an integer? When this is documented it should be clear what values are accepted.

haringsrob:
Hey @kallefrombosnia, want to move this forward, so I left a few comments.

Would also like to add tests for this, if you feel comfortable doing that have a go at it, otherwise let me know then I continue this ;)

kallefrombosnia:
Hey @haringsrob, thanks for the detailed reviews. I find your reviews very informative and logical, but unfortunately, I'm abroad so I'm unable to resolve these issues.
If you really want you can finish it, but when I'm back (next week) I'll surely look into this PR.

haringsrob:
Hey @kallefrombosnia, if you feel up for it can you have another try/look at the code to see if I did not miss anything?

After that we can merge it into 3.x

kallefrombosnia:
Hey @haringsrob, I have looked into changes and they just represent my initial idea which was in config to model settings and own facade settings.
Great work, I think that's it. 😄

Issue #611, opened 2020-03-30
Storage folder symlink creation not documented


omzy:
When using the local storage endpoint we need to create the storage folder symlink using the command `php artisan storage:link`. This needs to be added to the documentation.

haringsrob:
This is added in https://github.com/area17/twill/pull/1587

PR #1484, opened 2022-03-01
[3.x] Better defaults

haringsrob:
- [x] media and files
- [x] Default crops removed from generated models
       @ifox should we maybe change this into a method? GetCrops? Then in config we have "modelCrops" that are defaulting. We can still keep the property, then take them in order: Property there then property, if not, config.
- [x] renderBlocks(false)
- [x] Autoseed default to true
- [x] Disable video URL on media


phpunit.xml
```diff
@@ -66,10 +66,7 @@
     <env name="DB_CONNECTION" value="sqlite"/>
     <env name="DB_DATABASE" value=":memory:"/>
     <env name="LOG_CHANNEL" value="single"/>
-    <env name="MEDIA_LIBRARY_IMAGE_SERVICE" value="A17\Twill\Services\MediaLibrary\Glide"/>
     <env name="MEDIA_LIBRARY_LOCAL_PATH" value="media-library"/>
-    <env name="MEDIA_LIBRARY_ENDPOINT_TYPE" value="local"/>
-    <env name="FILE_LIBRARY_FILE_SERVICE" value="A17\Twill\Services\MediaLibrary\Glide"/>
```
haringsrob, line 72: I am not sure why the file library service was set to glide?

ifox: Looks like an oversight, I don't think this was intentional.

haringsrob:
Screenshot 2022-03-01 at 11 40 33

I changed the output of testbench, also I commented what I think was a leftover, timeout from the test and now they all complete in ~20seconds!

ifox:
WAT?!

I'm not sure if I need more coffee, but I'm not seeing what you commented that changed the tests duration so drastically.

tests/integration/TestCase.php
```diff
@@ -566,9 +566,13 @@ public function copyFiles($files)
...
                 }
 
+                if ($this->files->exists($destination)) {
+                    $this->files->delete($destination);
+                }
+
                 $this->files->copy($source, $destination);
 
-                usleep(1000 * 100); // 100ms
+                /* usleep(1000 * 100); // 100ms */
```
haringsrob, line 575: @ifox here

ifox: 😱

haringsrob: I honestly was wondering why the tests took so long before. Guessing it had something to do with testbench.

kallefrombosnia:
Nice changes regarding documentation, very informative. 👍 
A few days ago I have opened small PR #1488 which I have closed since this PR will implement it.
PR was fixing #611 so if possible you can target that issue in this PR and fix it.

haringsrob:
Hey @ifox would appreciate a review on this. Then we can merge this in 3.x

haringsrob:
@ifox, can we remove the configurability of this to always use bigIncrements/bigInteger? Or are there drawbacks to this?

```php
if (!function_exists('twillIncrementsMethod')) {
    /**
     * @return string
     */
    function twillIncrementsMethod()
    {
        return config('twill.migrations_use_big_integers')
            ? 'bigIncrements'
            : 'increments';
    }
}

if (!function_exists('twillIntegerMethod')) {
    /**
     * @return string
     */
    function twillIntegerMethod()
    {
        return config('twill.migrations_use_big_integers')
            ? 'bigInteger'
            : 'integer';
    }
}

```

Issue #1502, opened 2022-03-09
make repeater title and trigger translatable


cpfpolkehn:
## Summary

custom directives like `@twillRepeaterTitle` or `@twillRepeaterTrigger` currently are _not_ translatable.

## Describe the solution you'd like

It should be possible to pass translation keys that get rendered in the repeater form(s).

PR #1503, opened 2022-03-09
refs #1502 - make custom twill blade directives translatable

cpfpolkehn:
## Description

this PR enables translations for blade directives like `@twillRepeaterTitle`

## Related Issues

Addresses #1502

### DIFF
--- a/src/Services/Blocks/Block.php
+++ b/src/Services/Blocks/Block.php
@@ -345,7 +345,7 @@ class Block
             preg_match("/@{$pattern}{$bladeProperty}\(" . self::PREG_REPLACE_INNER . "\)/", $block, $matches);
 
             if (filled($matches)) {
-                return $matches[1];
+                return twillTrans($matches[1]);
             }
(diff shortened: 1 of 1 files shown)


haringsrob:
Thank you for your contribution @cpfpolkehn.

While this for sure solves the issue, I am not sure it is the solution that we should take.

I feel it might be better to have the translatable string in the property something like:

```
@twillRepeaterTitle(__('Category'))
@twillRepeaterTrigger(__('Add cat'))
```

But this does not work yet at the moment.

What are your thoughts @ifox?

cpfpolkehn:
@haringsrob Hello, I think in your example it should rather be `@twillRepeaterTitle(twillTrans('Category'))` than `@twillRepeaterTitle(__('Category'))` because `__` would ignore the Twill user's language.

haringsrob:
Closing this as we refactor to use __ in https://github.com/area17/twill/pull/1523

PR #1504, opened 2022-03-10
set correct block class name

YasienDwieb:
## Description
Class name should be `ImagesGridBlock` instead of `ImagesGrid`. Even if it can be known from the code example but it's misleading to have it in 2 forms so it should be set clear that it must have `Block` suffix


haringsrob:
Thank you @YasienDwieb

This was also resolved by https://github.com/area17/twill/pull/1506 which included some other fixes as well.

PR #1513, opened 2022-03-11
Allows slugs to be empty

freshleafmedia:
## Description

This PR allows slugs to be empty. The use case being entities like a homepage.

I'm pretty new to the Twill Vue setup so not sure if the changes are how you would want to go about this.

I've marked this as a draft because I was unable to generate the dist assets. When I ran `yarn build` I got a large number of `Syntax Error: TypeError: token.type.endsWith is not a function` in files I haven't changed, what am I doing wrong?

### DIFF
--- a/frontend/js/components/TitleEditor.vue
+++ b/frontend/js/components/TitleEditor.vue
@@ -7,7 +7,7 @@
         </a>
         <span v-else>{{ customTitle ? customTitle : title }}</span>
       </h2>
-      <a v-if="(permalink || customPermalink) && !showModal" :href="fullUrl" target="_blank" class="titleEditor__permalink f--small">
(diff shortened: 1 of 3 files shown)


haringsrob:
Hey @freshleafmedia,

Forgive me if I am mistaken, but I believe this is not needed.

You can set 

```
class BlogController extends BaseModuleController
{
    protected $moduleName = 'blogs';

    protected $indexOptions = [
        'permalink' => false,
    ];
}
```

Then the field will not be shown.

freshleafmedia:
That would disable all permalinks for that model, what I want is to have permalinks which are optionally empty.

To expand on my use case a little, I have a Pages Capsule and each page has permalink except the homepage which I want to be just `/` aka an empty permalink

kallefrombosnia:
Interesant use case, but I think it's opinionated for the project.

I would create an additional checkbox with the 'Homepage' value, and in the repository, with the `prepareFieldsBeforeSave` method I would overwrite the `permalink` value to the `/` (quick fix).

It's really up to you to serve the homepage, Twill treats all slugs the same.

PR #1514, opened 2022-03-13
Adds URL Field Type

freshleafmedia:
## Description

Simply adds `url` as a field type.

As noted in a PR #1513 I am unable to build the dist assets due to errors in unrelated files

### DIFF
--- a/frontend/js/components/Textfield.vue
+++ b/frontend/js/components/Textfield.vue
@@ -87,6 +87,23 @@
         @blur="onBlur"
         @input="onInput"
       />
+      <input v-if="type === 'url'"
+        ref="input"
+        type="url"
+        :placeholder="placeholder"
+        :name="name"
+        :id="uniqId"
+        :disabled="disabled"
+        :maxlength="displayedMaxlength"
+        :required="required"
+        :readonly="readonly"
+        :autofocus="autofocus"
+        :autocomplete="autocomplete"
+        :value="value"
+        @focus="onFocus"
+        @blur="onBlur"
+        @input="onInput"
+      />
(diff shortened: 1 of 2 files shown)


frontend/js/components/Textfield.vue
```diff
@@ -87,6 +87,23 @@
         @blur="onBlur"
         @input="onInput"
       />
+      <input v-if="type === 'url'"
```
haringsrob, line 90: Thinking as most of these are the same we could just do:

```vue
<input
        ref="input"
        :type="type"
        :placeholder="placeholder"
        :name="name"
        :id="uniqId"
        :disabled="disabled"
        :maxlength="displayedMaxlength"
        :required="required"
        :readonly="readonly"
        :autofocus="autofocus"
        :autocomplete="autocomplete"
        :value="value"
        @focus="onFocus"
        @blur="onBlur"
        @input="onInput"
      />
```

If it is not the e-mail (or even then we can conditionally add the pattern.).

haringsrob:
Hey @freshleafmedia!

Thanks for the contribution, I had a quick test and it builds fine here. Maybe it is a difference in node version?

I also left a suggestion, what do you think about that?

frontend/js/components/Textfield.vue
freshleafmedia: I had the same thought, I can create a separate PR for that :+1:

freshleafmedia:
Hi @haringsrob

Oh possibly, I'm using Node v16.14.0 what are you using?

haringsrob:
I am on v17.6.0

Issue #1520, opened 2022-03-15
indexTableColumn lables are not translated by default


crxiu:
## Summary

indexTableColumn lables are not translated by default

ModuleController->getIndexTableColumns does only use the indexColumns array item title filde directly without translation.

While the only possibility to set this indexColumns is to override the protected member var $indexColumns in my Controller, there is no possibility to access a translation function here.

`   protected $indexColumns = [
        'mycolumnname' => [
         'title' => twillTrans('admin.mycolumnname') // <-- can't be done here(!) twillTrans is not callable yet in the property declaration area of a class
...
`

The workarraound right now is I have to override the ModuleController method in each Controller I want it to happen. (Or create an inbetween Parent class inheritations like: ModuleController < MyAppModuleController < MyController)

For example the workarround could like this.
`    protected function getIndexTableColumns($items)
    {
        $columns = parent::getIndexTableColumns($items);
        array_walk($columns, function (&$column) {
            $column['label'] = twillTrans($column['label']);`
    });
      return $columns;

## Describe the solution you'd like

I would suggest to just use a twillTrans() call when setting the tableColumns['lable'] value in the Module Controller

`            $tableColumns[] = [
                'name' => $columnName,
                'label' => twilTrans($column['title']),...
`

## Additional context
none

haringsrob:
This functionality is improved with Twill 3.x using the new tablebuilder.

PR #1521, opened 2022-03-15
#1520 translate indexColumns labels

crxiu:
## Description

This will enable the possibility to just add a translation key in the $indexColumns

e.g. 
    protected $indexColumns = [
        'mycolumname' => [
            'title' => 'admin.mycolumname',

## Related Issues
#1520 
suggestion for https://github.com/area17/twill/issues/1520

### DIFF
--- a/src/Http/Controllers/Admin/ModuleController.php
+++ b/src/Http/Controllers/Admin/ModuleController.php
@@ -1188,7 +1188,7 @@ abstract class ModuleController extends Controller
 
             $tableColumns[] = [
                 'name' => $columnName,
-                'label' => $column['title'],
+                'label' => twillTrans($column['title']),
                 'visible' => $visibleColumns ? in_array($columnName, $visibleColumns) : ($column['visible'] ?? true),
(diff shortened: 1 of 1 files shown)


haringsrob:
Hey @crxiu, while I am confident this works, I feel it is not the correct approach.

I think that for this sort of data we have to move away from using properties and convert this into a method instead.

We can also utilize that strategy to add some OOP to this.

Example:
```php
<?PHP
public function getIndexTableColumns(): TwillTableColumns {
  $columns = new TwillTableColumns();
  $columns->addNew(new Column(
    name: 'columnName',
    label: __('Label'),
    visible: true,
    ....
  ));
  ...
  return $columns;
}

```

@ifox maybe we can merge this pr, but change the approach in 3.x or 4.x?

haringsrob:
Hello @crxiu,

For now we will not merge this pr. In 3.x we will add new ways to define TableColumns.

What you can do for now is extend/overwrite getIndexTableColumns and adjust it from there.

PR #1523, opened 2022-03-16
[3.x] #1502: Allow to use __( in twill properties.

### LINKED ISSUE #1502: make repeater title and trigger translatable

haringsrob:
## Description

This pr aims to add support for using __('text') in simple properties.

It also removes the cleanup regex and replace it with null blade properties to make it easier to maintain.

## Related Issues

fixes #1502


haringsrob:
Hey @ifox 

Can we set the laravel locale as well in our localization middleware?

What's the reason to have a custom locale implementation? Maybe you already explained before sorry.

```
    public function handle($request, Closure $next)
    {
        if ($request->user()->language) {
            config(['twill.locale' => $request->user()->language]);
        }

        return $next($request);
    }
```

ifox:
The idea of having a custom locale within Twill is so that you can set it to a different locale than your main Laravel locale. For example, if you're building a website in French, you will set your Laravel locale to French, but you might decide to use Twill I'm English. And then since Twill users can set their own locale, I guess it made sense to introduce a dedicated locale. I'm open to discuss it.

haringsrob:
Hey @ifox,

I believe with https://github.com/area17/twill/pull/1525 we can keep that behavior intact, but still use the standard local handling of Laravel.

We can of course keep twillTrans for compatability.

PR #1525, opened 2022-03-16
[3.x] Use default locale.

haringsrob:
## Description

Additional small pr to use laravel localization in the backend.

## Related Issues


haringsrob:
@ifox this pr changes to not use the custom locale, this seems to work fine as the locale is only set in the twill locale middleware.

But I might overlook some things. any input?

haringsrob:
Hey @ifox,

Maybe we can add this to 3.x? If you have any consideration before we do, I am happy to look into it.

PR #1537, opened 2022-03-21
Specify the config file's path

AndreSchwarzer:
## Description

For new developers it can be hard to get started with new packages. 

As a twill beginner myself, it took me a bit to figure out that the settings needed to be in the `config/twill.php` config file.


AndreSchwarzer:
> As a twill beginner myself, it took me a bit to figure out that the settings needed to be in the `config/twill.php` config file.

And also under the `block_editor` key.

haringsrob:
Thanks a lot @AndreSchwarzer !

PR #1550, opened 2022-03-28
[3.x] Pivot repeaters dev

haringsrob:
## Description

This adds support for using pivot fields in repeaters.

Todo:

- [x] https://github.com/area17/twill/pull/1460, I think we can remove the "keepExisting", there is no mention ever on the internet. And for such a specific case, the implementor can override the method in the future?


haringsrob:
This is now more complete. Basic instructions for now are:

Create a repeater as you would normally do, but this time with a many-many relation and pivot fields.

Example migration:
```
        Schema::create('blog_category', function (Blueprint $table) {
            $table->increments('id');
            $table->foreignIdFor(\App\Models\Blog::class);
            $table->foreignIdFor(\App\Models\Category::class);
            $table->string('pivot_field_1')->nullable();
            $table->string('pivot_field_2')->nullable();
            $table->integer('position')->default(999);
        });
```

And in your blog model:
```
    public function categories(): BelongsToMany
    {
        return $this->belongsToMany(Category::class)->orderByPivot('position');
    }
```

Then add the repeater to your form like this:
```
    @formField('repeater', [
        'type' => 'category',
        'allowCreate' => true, // Defaults to true, this you can use to disable creating new entries but still reference existing.
        'relation' => 'categories', // The relation method in your model.
        'browserModule' => [ // Same as a browser field
            'label' => 'Category',
            'name' => 'categories',
        ]
    ])
```

In the repository use the repeater `WithPivot` methods:

```php
    public function afterSave($object, $fields)
    {
        $this->updateRepeaterWithPivot(
            $object,
            $fields,
            'categories',
            ['pivot_field_1', 'pivot_field_2'],
            'Category',
            'category',
        );
        parent::afterSave($object, $fields);
    }

    public function getFormFields($object)
    {
        $fields = parent::getFormFields($object);
        return $this->getFormFieldForRepeaterWithPivot(
            $object,
            $fields,
            'categories',
            ['pivot_field_1', 'pivot_field_2'],
            'Category',
            'category'
        );
    }
```

haringsrob:
Demo:

demo-belongs-to-many-repeater

Docs:
screencapture-localhost-8081-docs-relations-belongs-to-many-html-2022-04-01-11_39_54

haringsrob:
Still have to do some cleanups of duplicated code

tests/integration/Repositories/RepositoryRepeatersTest.php
```diff
@@ -0,0 +1,217 @@
...
+    {
+        $fields = [
+            'repeaters' => [
+                'project_partners' => [
+                    [
+                        // @todo This should not be required to be here.
+                        // In prepareFieldsBeforeSaveHandleTranslations there is null-ing of fields that are
+                        // not in the data request, but that is not needed.
+                        // At least not for repeaters.
+                        'title' => ['en' => 'partner 1'],
```
haringsrob, line 48: @ifox this is the part that I feel should not be required.

PR #1561, opened 2022-04-04
[2.x] More permissive preview

m4n1ok:
## Description

This PR changes a bit the way how user can interact with CMS preview. 

Interaction with anchor links from current page are allowed as well as interactions with buttons. 
Form submission are disabled through the preview.


haringsrob:
Hey @m4n1ok,

Not sure what the exact use-case is for this?

But if you for example make a link `<a href="#">#</a>` and click it from the preview it will nest the editor.

Screenshot 2022-04-05 at 09 07 55

m4n1ok:
Hey @haringsrob, 

Thank you for the feedback, I will make an update for that.  

About the specific use case, the main idea is to give a preview working with javascript interactions and sometimes we may have behavior working with anchor links. 
For example, I'm working on a hidden footnote system and when user will click on a cta, the associated footnote will be shown. But the cta is here an `a` tag with an anchor.

mrdoinel:
I updated the code so all the links are disabled except this specific case :
Link is an anchor link (for example `href="#footer"`) AND the link have some JS attached to the click event (the click event have `event.preventDefault()` applied which overwrite the default behavior). Only this case will trigger the click.

And of course all the button elements.

mrdoinel (dismissed):

PR #1569, opened 2022-04-06
doc: minor changes to installation instructions

cambad:
## Description

- Make it clear you can install Twill on an existing application
- Make it clear that the artisan command is `twill:install` (for people that copy and paste directly)


ifox (approved):
Thanks, @cambad, great improvement!

PR #1579, opened 2022-04-11
[3.x] #375 file delete folder

### LINKED ISSUE #375: Delete media in media library not cascade delete the folder
when `MEDIA_LIBRARY_CASCADE_DELETE=true`
Delete the media from media library will delete the file itself, but not the folder containing it.
Testing it locally with Glide.

haringsrob:
## Description

Deletes the folders if the directory is empty after media deletion.

## Related Issues

Fixes #375


haringsrob:
Tested with aws a3 and also functions properly.

PR #1587, opened 2022-04-13
[3.x] #611: Document storage link.

### LINKED ISSUE #611: Storage folder symlink creation not documented

haringsrob:
## Description

Mentions to use storage link

## Related Issues

fixes #611

### DIFF
--- a/docs/src/getting-started/installation.md
+++ b/docs/src/getting-started/installation.md
@@ -35,6 +35,11 @@ Twill's `install` command consists of:
 - publishing Twill's assets for the admin console UI.
 - prompting you to create a superadmin user.
 
+## Storage
+
+If you have not yet done this following the Laravel installation guide, now would be a good time to run
+`php artisan storage:link` to setup the storage folders mapping to the public directory.
+
 ## Example setup
 
 If this is your first time using Twill or you just want to experiment with a Twill installation you can use the demo


rvyhvn:
Would be great if the `storage:link` also added on the [guide page](https://twillcms.com/guides/page-builder-with-blade/index.html).

Tofandel:
You can very easily create a PR to add it to this page

https://github.com/area17/twill/blob/3.x/docs/content/2_guides/1_page-builder-with-blade/2_installing-laravel.md?plain=1

rvyhvn:
@Tofandel, one more thing: When I was testing Twill using the guide, the image uploads weren't working properly. I found the fix: I had to change the `APP_URL` in my `.env` file from `http://localhost` to `http://127.0.0.1:8000`. Is this a known bug or just something the documentation missed? I’m thinking about adding a note about this to the guide so others don't get stuck.

Tofandel:
It's a common occurrence in laravel. Normally when you install and setup your env you would always be setting this even in localhost. Laravel doesn't assume your development env. Because you could be using a local apache and that would be localhost or you could be using the artisan serve command and then it's 127.0.0.1:8080.

You can add it to this doc, but I believe it's mentioned on the laravel doc itself

PR #1590, opened 2022-04-14
deleted items whose parent are not deleted should now be correctly displayed in trash list

YasienDwieb:
## Description
For a nested tree, If you deleted an item while its parent is not deleted, then it will not show in the trashed table. So while the count is correct, the table will be missing some items.

The fix for this, would be to display deleted items whose parents are not deleted as first level items without touching their tree-related attributes, this way when it gets restored again, it will be placed in the correct location

### Edit:
also when restoring an item whose parent is deleted, it will be set as a first level item


src/Models/NestedsetCollection.php
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * Build a tree from a list of nodes. Each item will have set children relation.
+     *
+     * To successfully build tree "id", "_lft" and "parent_id" keys must present.
+     *
+     * If `$root` is provided, the tree will contain only descendants of that node.
+     *
+     * @param mixed $root
+     *
+     * @return Collection
```
haringsrob, line 18: I believe this should be `@return NestedsetCollection`

src/Models/NestedsetCollection.php
```diff
@@ -0,0 +1,46 @@
...
+
+        $this->linkNodes();
+
+        $items = [];
+
+        $root = $this->getRootNodeId($root);
+
+        $ids = collect($this->items)->pluck('id')->toArray();
+
+        /** @var Model|NodeTrait $node */
```
haringsrob, line 34: NodeTrait is not imported so editors will not understand the typecast. It should be either imported or fully namespaced.

src/Repositories/Behaviors/HandleNesting.php
```diff
@@ -38,4 +38,17 @@ public function setNewOrder($ids)
...
+    /**
+     * @param \A17\Twill\Models\Model $object
+     * @return void
+     */
+    public function afterRestore($object)
+    {
+        if (!$object->parent) {
+            $object->parent_id = null;
+            $object->save();
+            return;
```
haringsrob, line 51: The return can be removed as there is nothing else after.

src/Models/Behaviors/HasNesting.php
```diff
@@ -93,4 +94,12 @@ public static function flattenTree(array $nodeTree, int $parentId = null)
 
         return $nodeArrays;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function newCollection(array $models = [])
+    {
+        return new Collection($models);
```
haringsrob, line 103: I think here we do not need to Alias it and we can just return new NestedsetCollection

haringsrob (changes requested):
Hey, thanks I tested and looks good. Just a few small nitpicks.

YasienDwieb:
Thank you @haringsrob for these helpful comments, they have all been addressed

src/Models/NestedsetCollection.php
YasienDwieb: @haringsrob  I have imported it

PR #1592, opened 2022-04-15
[2.x] Fix displayed language name in admin panel

Ser1ous:
## Description

When add to language list pt-BR and pt he displayed in language list same. 
https://monosnap.com/file/fXBhpEkkcIOUgL45iHK8lbPn25FYPl
If change function from getDisplayLanguage to getDisplayName. Language list be more correct. 
https://monosnap.com/file/e2ohyGD3ZU0ZmvLsrHaFab2c5Aw2xM
It's work for all language modifie, like Chines Simply and Chines Tradition

## Related Issues

No issue, just find and fix :-)

### DIFF
--- a/src/Helpers/i18n_helpers.php
+++ b/src/Helpers/i18n_helpers.php
@@ -83,9 +83,9 @@ if (!function_exists('getLanguageLabelFromLocaleCode')) {
     {
         if (class_exists(Locale::class)) {
             if ($native) {
-                return ucfirst(Locale::getDisplayLanguage($code, $code));
+                return ucfirst(Locale::getDisplayName($code, $code));
             } else {
-                return ucfirst(Locale::getDisplayLanguage($code, config('twill.locale', config('twill.fallback_locale', 'en'))));
+                return ucfirst(Locale::getDisplayName($code, config('twill.locale', config('twill.fallback_locale', 'en'))));
             }
         }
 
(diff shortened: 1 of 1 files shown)


haringsrob:
Thank you @Ser1ous,

@ifox I do not see any issue with merging this in 2.8.0 but just checking as you may?

src/Helpers/i18n_helpers.php
```diff
@@ -83,9 +83,9 @@ function getLanguageLabelFromLocaleCode($code, $native = false)
     {
         if (class_exists(Locale::class)) {
             if ($native) {
-                return ucfirst(Locale::getDisplayLanguage($code, $code));
+                return ucfirst(Locale::getDisplayName($code, $code));
```
haringsrob, line 86: I had a quick check and we can drop ucfirst here as getDisplayName is already properly capitalized. 

Also, you can remove the } else { as it is returned early.

haringsrob (changes requested):

src/Helpers/i18n_helpers.php
haringsrob: Same for the other call

Ser1ous: I think need } else { in reason, when need translate for not current language

haringsrob: Sorry, I should have been more clear as I did not mean the content of the else but rather to have it return properly.

```
if (true) {
   return 'bar';
} else {
   return 'foo';
}
```

is the same as:

```
if (true) {
   return 'bar';
}

return 'foo';
```

Ser1ous: Ok, my fail :-( . Fixed this.

Ser1ous: In this function has problem when class Locale not exists. But i check again on next weekend and if he be, then i fix too.

ifox (approved):

PR #1596, opened 2022-04-17
Update articles_index.blade.php

happytodev:
It misses an 's' to the model name


pboivin:
Hi @happytodev. In this example, the index route name is plural (`articles`) and the detail route name is singular (`article`). The change you're suggesting is on a link to a detail page. Singular is correct.

See: https://github.com/area17/twill/blob/0eacb70cf3b437c77f33a65dbd36fbf8a5dd198d/docs/src/guides/building_a_multilingual_site_with_twill_and_laravel_localization/routes_web.php#L58-L62

PR #1597, opened 2022-04-17
Update routes_web.php

happytodev:
In region newsrouteupdated : 
Remove double call to use LaravelLocalization and add use to the article's model


haringsrob:
Thank you @happytodev for your contribution!

PR #1606, opened 2022-04-19
[2.x] Custom JS function dynamic invoke in text component

kallefrombosnia:
## Description

This functionality adds a feature for developers to invoke a custom JS function which will transform the output of the input and insert event info into something new.

Since this made confusion on Discord I will explain by code details.

This PR adds 2 new options for text fields:
**callback** - this prop tells that we use dynamic invoke in the input component (bool) - default `false`
**execute** - name of the JS function (string) - default `''`

```php
@formField('input', [
    'name' => 'description',
    'label' => 'Description',
    'type' => 'text',
    'callback' => true,
    'execute' => 'handleStringChange'
])
```

After this, we need to define the JS function

```php
@push('extra_js')
  <script>
        function handleStringChange(string){
            return string.split('').join(',');
        }
  </script>
@endpush
```

Basically, this function will split this string into pieces and glue them by the comma.

Let's test it.

On load:
test1

After making some changes to the existing input:
test2

Notes: I'm aware of the existing `onChange` prop, but I didn't want to mix content here since `onChange` is used for static function calls in other parts of the website.

Also if the user doesn't return anything from the custom function, there is a check which if fails returns the original event target value as the output.


haringsrob:
Hey @kallefrombosnia,

2 things:

1. Can't we use a single parameter that if filled in will execute the callback, example: `execute-callback: 'my-callback'`
2. We needs some documentation before we can merge this.

@ifox anything else missing maybe? What do you think?

kallefrombosnia:
> Hey @kallefrombosnia,
> 
> 2 things:
> 
> 1. Can't we use a single parameter that if filled in will execute the callback, example: `execute-callback: 'my-callback'`
> 2. We needs some documentation before we can merge this.
> 
> @ifox anything else missing maybe? What do you think?

Hey @haringsrob, I agree with your statements.
1. Having 2 parameters for the callback is overhead, so I'll change it to one param.
2. Yup I'll document this with the changes from 1.

I'm open to new ideas on this.

kallefrombosnia:
I don't see a real purpose for this feature, so Ill close this PR.

PR #1612, opened 2022-04-23
Media note field proper wrap

### LINKED ISSUE #1604: Notes for some field types don't wrap properly
Very simple to reproduce.  Add a media field to a form and add a note of more than about 

image

Seen in latest Twill 2.8

kallefrombosnia:
## Description

It Fixes media note overflow when working with the long text strings.

Really long string:
test3

Long string:
test4

Short string:
test5

Tested in Chrome.

## Related Issues
Fixes #1604


haringsrob:
Hey @ifox, for me this is fine. But I personally would keep the button always top aligned.

Any chance someone can have a quick peek at this to suggest? This is an edge case as well.

haringsrob:
🤦‍♂️ I don't know why, maybe because it's not my pr, but I can't seem to diff this properly to remove the assets.

Might cherry-pick to a new branch.

kallefrombosnia:
I had issues with this also, but I resolved it by SOFT git reset where I go back to the latest fork 3.x branch commit, and from there I delete all of the assets from my side so they don't make a change.

adam-jones-net:
This doesn't seem to work in the current 2.x branch.  I still get the issue

image

PR #1613, opened 2022-04-23
Search input length trigger change

kallefrombosnia:
## Description

On Discord user, Gvido Blue Fox#3566 noticed that search won't be activated for 2 or fewer string lengths.
This creates problems for search fields that are shorter or equal to 2.

I have changed the limit to 0, which triggers a search on the first input change event with 300 delays from `debounce`.

### DIFF
--- a/frontend/js/components/Search.vue
+++ b/frontend/js/components/Search.vue
@@ -170,7 +170,7 @@
       },
       onSearchInput: debounce(function (event) {
         this.searchValue = event.target.value
-        if (this.searchValue && this.searchValue.length > 2) {
+        if (this.searchValue && this.searchValue.length > 0) {
           if (this.type === 'dashboard') {
             htmlSearchClasses.forEach((klass) => {
               html.classList.add(klass)


kallefrombosnia:
@haringsrob Just want to discuss with you since you wanted to remove the whole logic about the length check.
The only issue with that I see is handling add/remove of Search component CSS classes, which currently relies on this limit check and it's `else`.

```js
      onSearchInput: debounce(function (event) {
        this.searchValue = event.target.value
        if (this.searchValue && this.searchValue.length > 0) {
          if (this.type === 'dashboard') {
            htmlSearchClasses.forEach((klass) => {
              html.classList.add(klass)
            })
          }
          this.fetchSearchResults()
        } else {
          if (this.type === 'dashboard') {
            htmlSearchClasses.forEach((klass) => {
              html.classList.remove(klass)
            })
          }
          this.readyToShowResult = false
          this.searchResults = []
          this.setLastFocusElement()
        }
      }, 300)
```

haringsrob:
Hey @kallefrombosnia,

I feel like it is ok the way it is now, if there is any input we show the search classes, otherwise we don't.

It feels ok visually when using it.

haringsrob:
@ifox I can add this to the next 2.x minor release if you agree

ifox (approved):
Thanks, @kallefrombosnia! As a reminder, please don't push the assets to avoid conflicts.

kallefrombosnia:
Hey sorry again, assets removed. 
I need to watch out for this in future PRs. 😄

haringsrob:
> Thanks, @kallefrombosnia! As a reminder, please don't push the assets to avoid conflicts.

Maybe we can .gitignore them and force add them when a release is made?

I can make a command `php artisan twill:release VERSION` like `php artisan twill:release 2.8.1` which will build, tag and push?

PR #1618, opened 2022-04-26
Fix SQL error on twill:refresh-crops command

nedbaldessin:
When a crop dimensions are calculated, they may return float values. This can result in an SQL error when trying to save a float number as an integer in the database. So we simply round down the dimensions to prevent this.


ifox (approved):

ifox:
Hey @nedbaldessin! Nice to see you here, thank you for your contribution! 

PS: I'll drop you a text, we should catch up!

PR #1619, opened 2022-04-26
[3.x] Add selectable prop on VueSelect component

### LINKED ISSUE #1266: Disabled options in select form field
## Summary

Hi,

as far as I can see (please correct me if I'm wrong), there is no way to disable options in the[ select form field](https://twill.io/docs/#select) (like they are displayed but cannot be selected) with Twill v2.5.2. 

Twill's select component `VSelect` uses vue select. Vue Select v3.3+ has the prop `selectable` that determines if an option is selectable or not. Please update vue select to the given version or provide another solution to disable options in the select form field.

Thank you!

kallefrombosnia:
## Description
Twill uses Vue Select component which is extended in the Twill vselect component. From 3.3.0 Vue Select, there is a new component prop `selectable` which allows us to set the disabled state on specific select options.

If we don't pass any value (default case), all options would be available in the select.

Usage:
`form.blade.php`

```php
@formField('select', [
    'name' => 'office',
    'label' => 'Office',
    'placeholder' => 'Select an office',
    'options' => [
        [
            'value' => 1,
            'label' => 'New York'
        ],
        [
            'value' => 2,
            'label' => 'London'
        ],
        [
            'value' => 3,
            'label' => 'Berlin',
            'selectable' => false
        ]
    ]
])
```

Note on the third item we have new property `selectable` which allows us to set a boolean value regarding should be option selectable or not.

test6

It's hard to see the change here since Lightshot doesn't capture the icon which is shown over the non-selectable option. 

Also, it creates some CSS class conflict in which the -1 element keeps a hover state. 
I will inspect this and create documentation in the next commits.

## Related Issues
Fixes #1266


haringsrob:
👍 
Looks good and simple enough. Please however, for now do not included the compiled css/js as we are still doing many changes and want to avoid conflicts there.

kallefrombosnia:
@haringsrob Thanks, compiled stuff is removed from PR.

haringsrob:
I have updated the pr to change the form doc and also added this selectable prop to the formbuilder option object.

PR #1646, opened 2022-05-17
[2.x] Added the possibility to duplicate an object even if the model has no…

### LINKED ISSUE #673: Several problems with Duplicate action
Amazing new feature introduced in https://github.com/area17/twill/pull/556 , thanks for it!

Some problems:

1. The "Duplicate" appears in the actions list even when indexOption 'duplicate' is set to false.
2. Default implementation of duplication does not work if the model does not have revisions.
3. Action "Duplicate" appears for the items in the "Trash" list but it does not work there at all (even when enabled in indexoptions and the implementation fixed)

Thanks in advance!

ConstantMath:
Hello

## Description
Add the possibility to duplicate an object even if the model has no revisions

## Remark
There is certainly an internal function for that test ? 
→ `if(method_exists($object, 'revisions'))`

## Related Issues
Fixes #673


haringsrob:
Hey @ConstantMath,

First of all, thank you for your contribution!

I was checking this, but it is missing the replication of model translations, blocks, and repeaters.
And at the same time relations that are not model unique should be relinked.

Also, the Carbon import is missing.

Any other thoughts on this @ifox?

Cheers!

haringsrob:
No worries at all @ConstantMath,

the pr is a very good starting point. Maybe an alternative idea could be that if a model is not revisional we introduce a new table for storing the last submit (as we do in revisions).

Then when duplicating, we can take the last submit out of the "non_revisional_submits" table and keep the same logic.

This is a bit weird of an approach but could work well with the current setup.

haringsrob:
As there are some more features in https://github.com/area17/twill/pull/1731 we will use that pr to move this functionality forward.

PR #1667, opened 2022-05-31
[3.x] Code refactoring

haringsrob:
This is the first set of automatic upgrades.

The idea is to do this gradually as otherwise there may be too many conflicts + we can now check what changes were introduced before we up the level in case of issues.


haringsrob:
Hey @ifox,

I looked over this and I do not find any specific issues. I want to merge this so it gets tested automatically during further development.

Then, maybe each week we can improve on it. Next refactorings to be "typings"

What do you think?

PR #1671, opened 2022-06-01
[POC][FUTURE] Admin in Livewire

haringsrob:
This is an experiment I did a while ago. It is a simple POC that utilizes livewire instead of vue.

There is only a really small part rewritten and this is NOT usable.


haringsrob:
Closing for now as this is a far future idea.

PR #1687, opened 2022-06-08
Expose the position of each related item on the loaded related items

antonioribeiro:
I needed to use the position of the related item on an application feature and this was the only way I could find to expose it, can we do this differently?

### DIFF
--- a/src/Models/Behaviors/HasRelated.php
+++ b/src/Models/Behaviors/HasRelated.php
@@ -51,12 +51,16 @@ trait HasRelated
         return $this->relatedCache[$browser_name] = $this->relatedItems
             ->where('browser_name', $browser_name)
             ->map(function ($item) {
-                if (filled($item->related) && !isset($item->related->position))
-                {
-                    $item->related->position = $item->position ?? null;
+                if (filled($item->related)) {
+                    /** @var \A17\Twill\Models\Model $model */
+                    $model = $item->related;
+
+                    $model->setRelation('pivot', $item);
+
+                    return $model;
                 }
 
-                return $item->related;
+                return null;
             });
     }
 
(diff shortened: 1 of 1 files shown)


src/Models/Behaviors/HasRelated.php
```diff
@@ -51,6 +51,11 @@ public function loadRelated($browser_name)
         return $this->relatedCache[$browser_name] = $this->relatedItems
             ->where('browser_name', $browser_name)
             ->map(function ($item) {
+                if (filled($item->related) && !isset($item->related->position))
+                {
+                    $item->related->position = $item->position ?? null;
+                }
+
```
haringsrob, line 58: Hey @antonioribeiro,

I had a look at this and I feel it will be better if we treat "related" as a pivot table?

```php
->map(function ($item) {
  /** @var \A17\Twill\Models\Model $model */
  $model = $item->related;

  $model->setRelation('pivot', $item);

  return $model;
```

This way, you can do `Page::getRelated('test')->first()->pivot->position`

What do you think?

haringsrob: Screenshot 2022-06-10 at 09 12 40

antonioribeiro:
Yes, looks awesome, @haringsrob! Feel free to put this one down. I didn't explore this at all and I was also feeling it could do good if we had access to the other related columns, so you are solving everything. Thank you!

haringsrob:
I just realise this is already in 2.8.7 https://github.com/area17/twill/blob/2.x/src/Models/Behaviors/HasRelated.php#L53

Realised that I accidentally committed it. And @pboivin already provided a fix for it as well https://github.com/area17/twill/commit/2411499535d8d4cebc9f9bafd8c9ecbc759b817a

PR #1723, opened 2022-06-30
allow tagged table to be used with other tables than tag table

YasienDwieb:
## Description

For situations where you can create tags based on existing models other than the tag table, you will have to create pivot table for doing that. Taking into consideration that you can make this tagging relation with more than one model, therefore you will have to do more and more similar tables just to acheive that tagging functionality

In case you are making tags on 5 models, you will create 5 more tables and if this tagging will be needed for other models, hence it will produce a lot of tables

I think using the same Tagged table for doing this could be a good option.


haringsrob:
Hey @YasienDwieb,

I am not too sure about this one. This feels like something that is more for the "related" table?

So you could add a "tags" which is a browser field and unify that?

haringsrob:
Hey @YasienDwieb,

I had another brief look at this. It is missing a migration to make it available on existing sites.

Also, we needs to have a bit of doc's on this, but I feel since this is a new feature it should go straight into Twill 3.

YasienDwieb:
Hey @haringsrob ,
Thanks you so much for your great efforts. There's a migration included which adds the needed fields to the tagged table
`migrations/default/2020_02_09_000006_create_twill_default_tags_tables.php`, please let me know if this is not what is needed. 

Also, yeah, it would need some docs, I will work on that too.

Thank you

haringsrob:
Hey @YasienDwieb,

That migration works fine for new Twill installations, however, for an existing Twill installation that will not work. So the changes you made to the table should move to a new migration file instead.

haringsrob:
This will go into 3.x only, I will work on the update.

PR #1719, opened 2022-06-28
[2.x] feat(form): Move default submit options to controller method

pboivin:
## Description

This consolidates the form's default submit options into a `getSubmitOptions()` controller method. This method can more easily be extended to further customize the options based on the current `$item` and other information from the current request.

This also includes a small fix to `SingletonModuleController` to adjust the submit options, as `Save and close` and `Save and create new` are not really supported in the context of a singleton module.

## Related 

Follows https://github.com/area17/twill/pull/1467 in the spirit of adding more flexibility to the publishing workflow.

PR #1726, opened 2022-06-30
feat(users): Add hooks to extend user create/edit forms with custom fields

pboivin:
## Description

This PR introduces optional Blade includes into the users create and edit forms. This can be used to add custom fields by creating the optional view into one's project. For example:

```blade
// file: resources/views/vendor/twill/users/form-custom-fields.blade.php

@formField('checkbox', [
    'name' => 'subscribe_to_team_newsletter',
    'label' => 'Subscribe to the team newsletter',
])
```

custom-field

The `form-custom-fieldsets` view is even more generic and can support `@section('fieldsets')` and `@section('sideFieldsets')` directives.

The custom fields can be processed in various ways by extending some of Twill's external classes. The easiest way might be to extend the User model's `$fillable` property.

## Related 

[One approach for extending Twill internals](https://github.com/area17/twill/discussions/1232)

[A short discussion on Discord regarding user custom fields](https://discord.com/channels/811936425858695198/811986149064441927/956924561344036925)

### DIFF
--- a/views/users/form.blade.php
+++ b/views/users/form.blade.php
@@ -90,7 +90,11 @@
             @endcomponent
         @endunless
     @endif
+
+    @includeIf('twill::users.form-custom-fields')
 @stop
+
+@includeIf('twill::users.form-custom-fieldsets')
 
 @push('vuexStore')
     window['{{ config('twill.js_namespace') }}'].STORE.publication.submitOptions = {
--- a/views/users/create.blade.php
+++ b/views/users/create.blade.php
@@ -18,3 +18,5 @@
         ])
     @endcan
 @endunless
+
+@includeIf('twill::users.create-custom-fields')


haringsrob:
Hey @pboivin,

Can we maybe add a little bit of documentation to this?

Also wondering if we maybe can do something like a check for `resources/views/admin/users/{form/create}.blade.php` and use that instead?

Maybe that even works out of the box.

views/users/form.blade.php
```diff
@@ -90,8 +90,12 @@
...
     @endif
+
+    @includeIf('twill::users.form-custom-fields')
 @stop
 
+@includeIf('twill::users.form-custom-fieldsets')
+
 @push('vuexStore')
     window['{{ config('twill.js_namespace') }}'].STORE.publication.submitOptions = {
         draft: [
```
haringsrob, line 101: https://github.com/area17/twill/pull/1719 maybe we should also move these?

pboivin:
Hi @haringsrob, thank you, I appreciate the feedback!

> Can we maybe add a little bit of documentation to this?

I don't see a lot of value in adding documentation for this alone, as it's a bit of an escape hatch to avoid too much overriding in special cases. However, I do see the need for more generic documentation on how to work with the internal Blade views, layouts, sections, special variables, etc. In the meantime, we could add inline comments explaining the intention of the `@includeIf` directives.

> Also wondering if we maybe can do something like a check for resources/views/admin/users/{form/create}.blade.php and use that instead?

Let me know if I'm not understanding this correctly — I think adding a check for `for resources/views/admin/users/{form/create}.blade.php`, for a subsection of the internal `{form/create}.blade.php` may be confusing... It would break my expectation that `{form/create}.blade.php` corresponds to the full view. Does this make sense?

Maybe the `@includeIf` directives could be turned into `@section` directives, so the path to extensibility becomes something like this instead :

```blade
// file: resources/views/admin/users/form.blade.php

@extends('twill:users.form')

@section('form-custom-fields')
    @formField('checkbox', [
        'name' => 'subscribe_to_team_newsletter',
        'label' => 'Subscribe to the team newsletter',
    ])
@endsection
```

pboivin:
Closing this for now. It's probably best to let this simmer a bit more.

haringsrob:
Hey @pboivin,

My idea with `resources/views/admin/users/{form/create}.blade.php` would be that we do not introduce separate files, but if they want to add fields or fieldsets, they could just overwrite (copy) the original form/create, place it in their `resources/views/twill` folder and do with it as they wish.

pboivin:
Isn't this the same as manually copying the view to `resources/views/vendor/...`?

The main issue I see with overriding the `users/form.blade.php` view (and most internal views and layouts) is that it introduces the need for manual synchronization with every Twill update. I know that some of the internal views don't change all too often but still, from the perspective of an app developer introducing a few custom fields, it would be nice to have an alternative to complete overriding.

haringsrob:
Makes perfect sense, thank you for explaining. I did not think of it this way! I think this is perfectly fine then.

PR #1731, opened 2022-07-07
[2.x] Feature: Duplication without revisions

galingong:
These changes should make it possible to use the duplicate feature on a model without the need of also using revisions. All Twill features should work on duplication - medias-, files-, and related table associations are created for the new model. Blocks and child blocks (along with their files, medias and relations) are also created for the clone. Slugs are left empty on purpose.

I found it best, that **relationships** should be handled individually and thus they **are not duplicated by default** (parsing and filtering out methods from the model that are relationships would be problematic anyway). A **`$duplicatedRelations` property should be defined on the models' repository** instead, like so:

e7d45024-81b5-455e-a114-e1aab89bd45b

All formats work out of the 3 demonstrated in the image.


galingong:
#673
#1646

haringsrob:
hey @galingong.

We have decided not to implement this future in 2.x anymore. As most effort goes into 3.x at the moment and your work was (mostly) implemented there https://github.com/area17/twill/pull/1823

Thank you for proposing this change it helped us a lot.

Kind regards,
Rob

PR #1740, opened 2022-07-12
[2.x] Add wildcard to match.

### LINKED ISSUE #1733: @twillRepeaterTitleField gets echoed if below @twillRepeaterTrigger in repeater.blade.php file
## Description

`@twillRepeaterTitleField('heading', ['hidePrefix' => true])` gets echoed if the line is below `@twillRepeaterTrigger('...')` in repeater.blade.php file:

```PHP
@twillRepeaterTitle('Item')
@twillRepeaterTitleField('heading', ['hidePrefix' => false])
@twillRepeaterTrigger('Add Item')
```
This happens only if the options ( `['hidePrefix' => true]` ) are there

Screenshot:
Screenshot 2022-07-10 163639

Doesn't matter if other fields are below mentioned lines

### Versions

Twill version: 2.8.8
Laravel version: 8.83
PHP version: 7.4

haringsrob:
Fixes #1733 but I am not 100% confident about it. Will check more later.

This problem is resolved in 3.x so we might leave it for then as switching the position already fixes the issue.

### DIFF
--- a/src/Services/Blocks/Block.php
+++ b/src/Services/Blocks/Block.php
@@ -21,7 +21,7 @@ class Block
 
     public const TYPE_REPEATER = 'repeater';
 
-    public const PREG_REPLACE_INNER = '(?:\'|")(.*)(?:\'|")';
+    public const PREG_REPLACE_INNER = '(?:\'|")(.*)(?:\'|").*';
 
     /**
      * @var string


haringsrob:
@ifox I consider just closing this as it can be fixed by swapping the position in 2.x and there is no longer an issue in 3.x

ifox:
Sounds good to me!

PR #1743, opened 2022-07-18
ancestors sort by position desc to ensure the slug is following the tree

thingasd:
## Description

This pr fixes the getAncestorsSlug() return slug in wrong order if the child item is created before the parent item.


haringsrob:
Looks good @thingasd, I just added a test for this as well.

thingasd:
> 

@haringsrob Thanks a lot!

PR #1760, opened 2022-08-01
fix multilingual nestings

sineld:
Nesting multilingual sorting nodes occur 505 error, with this fix we get rid of it.


haringsrob:
Hey @sineld, could you provide some steps to reproduce?

Then I can write tests for this.

PR #1766, opened 2022-08-10
Adds Ordering To The Block->Children Relationship

freshleafmedia:
## Description

When querying `children` relationship of a block (eg when building repeaters) the returned models aren't sorted. I expected this because the `blocks` relationship defined by the `HasBlocks` trait does set ordering.

### DIFF
--- a/src/Models/Block.php
+++ b/src/Models/Block.php
@@ -45,7 +45,11 @@ class Block extends BaseModel
 
     public function children()
     {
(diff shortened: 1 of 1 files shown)


freshleafmedia:
I believe the tests are failing due to this issue https://github.com/doctrine/dbal/pull/5577

haringsrob:
Added a test in https://github.com/area17/twill/pull/1785 so all good.

Issue #1778, opened 2022-08-16
`support_subdomain_admin_routing` config is making phpunit tests fail


ptrckvzn:
## Description

I think I found a bug when running feature tests using `support_subdomain_admin_routing` set to `true`. What I think is happening is that in [`src/Http/Middleware/SupportSubdomainRouting.php`](https://github.com/area17/twill/blob/2.x/src/Http/Middleware/SupportSubdomainRouting.php#L29), the `twill-navigation` is overwritten dynamically. When running test and reusing the same app throughout, the override is run multiple times and the `key(config('twill-navigation'))` will remove a layer of navigation each time it is run, making the navigation invalid and the test failing because an error is thrown in `_overlay_navigation.blade.php`.

### Expected result

Tests run correctly and do not fail when using `support_subdomain_admin_routing`.

### Actual result

```
Next Facade\Ignition\Exceptions\ViewException: foreach() argument must be of type array|object, string given (View: /var/www/html/twill/views/partials/navigation/_overlay_navigation.blade.php)
```

### Possible solution

In this context, I think it would make more sense to check if the subdomain is available in twill-navigation (and at the same time, in the dashboard modules) and if not, leave the config as-is.

```diff
-            'twill-navigation' => config('twill-navigation')[$subdomain] ?? key(config('twill-navigation')),
+            'twill-navigation' => config('twill-navigation')[$subdomain] ?? config('twill-navigation'), 
-            'twill.dashboard.modules' => config('twill.dashboard.modules')[$subdomain] ?? key(config('twill.dashboard.modules')),
+            'twill.dashboard.modules' => config('twill.dashboard.modules')[$subdomain] ?? config('twill.dashboard.modules'),
```

I can submit a PR for this.

### Versions

Twill version:  2.7.0
Laravel version: 8.65
PHP version: 8.1
Database engine: pgsql / sqlite :memory:

PR #1779, opened 2022-08-16
[2.x] Fix tests failure when using subdomain routing

ptrckvzn:
## Description

Remove extracting `twill-navigation` and `twill.dashboard.modules` from the first key when the subdomain key doesn't exists. It will keep them as-is.

## Related Issues

#1778

PR #1786, opened 2022-08-19
[3.x] #1778 subdomain routing issues

haringsrob:
Different approach to https://github.com/area17/twill/pull/1779

My main concern is the overwriting of config. I have not dived too deep into it but I made a solution (I think) where it is irrelevant where the request comes from.

My main concern was that with overwriting config, in memory adaptions can cause conflicts (like the tests) but also Laravel octane.

One breaking change here, is that when subdomain routing is enabled, base domain routes are no longer created. I might need to revise this based on input.

Fixes https://github.com/area17/twill/issues/1778

### DIFF
--- a/examples/tests-subdomain-routing/config/twill.php
+++ b/examples/tests-subdomain-routing/config/twill.php
@@ -1,0 +1,24 @@
+<?php
+
+return [
+    'support_subdomain_admin_routing' => true,
+    'admin_app_subdomain' => 'admin',
+    'active_subdomain' => null,
+    'app_names' => [
+        'subdomain1' => 'App 1 name',
+        'subdomain2' => 'App 2 name',
+    ],
+
+    /**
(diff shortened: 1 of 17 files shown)


haringsrob:
Looking at the tests this has a lot of side effects, Investigating it :)

examples/tests-subdomain-routing/config/twill.php
```diff
@@ -0,0 +1,24 @@
...
+return [
+    'support_subdomain_admin_routing' => true,
+    'admin_app_subdomain' => 'admin',
+    'active_subdomain' => null,
+    'app_names' => [
+        'subdomain1' => 'App 1 name',
+        'subdomain2' => 'App 2 name',
+    ],
+
+    /**
```
haringsrob, line 12: Seems like it is find, except this issue which I will try to resolve now.

PR #1788, opened 2022-08-26
Add firstByField and firstByFieldOrCreate methods.

sauron:
## Description

Add `firstByField` and `firstByFieldOrCreate` methods to the ModuleRepository in order to easily find the first record by a translated field

Sometimes it is needed to return the first record by a translated field. In order to make that easy I'm creating this pull request so we can do:

```php
$country = app(CountryRepository::class)->firstByField('Argentina', 'name');

$country = app(CountryRepository::class)->firstByField('Argentine', 'name');
```

And along with it, the create method like:

```php
$country_name = 'Argentina';
$country = app(CountryRepository::class)->firstByFieldOrCreate($country_name, 'name', [
                'published' => true,
                'name' => [
                    'en' => $country_name,
                ],
            ]);
```

### DIFF
--- a/src/Repositories/ModuleRepository.php
+++ b/src/Repositories/ModuleRepository.php
@@ -363,7 +363,6 @@ abstract class ModuleRepository
      */
     public function duplicate($id, $titleColumnKey = 'title')
     {
-
         if (($object = $this->model->find($id)) === null) {
             return false;
         }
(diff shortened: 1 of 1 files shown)


ifox:
Hi @sauron,

Should we rely on laravel-trnaslatable for this? See https://docs.astrotomic.info/laravel-translatable/package/scopes#where-translation.

My first reaction to this was that I wouldn't expect the like operator to be used, at least not by default.

@haringsrob curious to get your opinion on this.

haringsrob:
Hi @sauron @ifox,

I started on the tests for this and indeed the following should be sufficient:

```php
        $result = \App\Models\Screen::orderByTranslation('title')->where('title', 'LIKE', '%title%')->get();

        $this->assertEquals('a-title-1-en', $result[0]->title);
        $this->assertEquals('b-title-1-en', $result[1]->title);

        $result = \App\Models\Screen::orderByTranslation('title', 'desc')->where('title', 'LIKE', '%title%')->get();

        $this->assertEquals('b-title-1-en', $result[0]->title);
        $this->assertEquals('a-title-1-en', $result[1]->title);

        // Now query just the first entry.
        $result = \App\Models\Screen::orderByTranslation('title')->where('title', 'LIKE', '%title%')->first();

        $this->assertEquals('a-title-1-en', $result->title);

        $result = \App\Models\Screen::orderByTranslation('title', 'desc')->where('title', 'LIKE', '%title%')->first();

        $this->assertEquals('b-title-1-en', $result->title);
```

We might consider still making a helper method but it feels a bit niche?

haringsrob:
See my note in https://github.com/area17/twill/pull/1792/files#diff-91743abe13f9cfcc9872b3ae3cee1c07961a328a42af0f3da32fc79792c56b42R72

sauron:
Thank you @ifox and @haringsrob. It works like that and yes, it is more like a niche.
I was trying to avoid the duplication among different projects and repositories, but at this point I think a custom Trait will do the trick.

PR #1789, opened 2022-08-26
fixed: loadRelated n+1 queries

ulcuber:

### DIFF
--- a/src/Models/Behaviors/HasRelated.php
+++ b/src/Models/Behaviors/HasRelated.php
@@ -43,7 +43,7 @@ trait HasRelated
      */
     public function loadRelated($browser_name)
     {
-        if (!isset($this->relatedItems)) {
+        if (!$this->relationLoaded('relatedItems')) {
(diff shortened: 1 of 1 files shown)


haringsrob:
Hello @ulcuber,

Today I was writing a test for this to see how I can reproduce. Could you provide some more information on how to get the issue so I can correctly verify the behaviour?

Thank you

ulcuber:
@haringsrob when checking property it ends with `__isset`->`offsetExists`->`getAttribute`->`getRelationValue` causing relation loading. Then `->load()` would be called causing relation loading again.
But since `relatedItems` is morphMany it will always return collection even when no related items found so `isset` will always return `true`.
Then it may be considered as reinsurance and is not reproducible on this relation.
On the other side when using many instances of models with `relatedItems` no difference between loading through property or `->load` method since `getRelationValue` will return loaded relation if it exists.

ulcuber:
@haringsrob yes, it is correct

PR #1799, opened 2022-09-05
Bug fix for changing admin assets directory

MauricioKruijer:
## Description
When running `php artisan twill:build` there seems to be an issue with the config/env variables. 
The values are not being passed through from php -> node. Therefore it is impossible changing the admin assets location, however this seems to be configureable in the twill config, this PR enables that feature.

### DIFF
--- a/src/Commands/Build.php
+++ b/src/Commands/Build.php
@@ -159,7 +159,9 @@ class Build extends Command
      */
     private function runProcessInTwill(array $command, $disableTimeout = false)
     {
-        $process = new Process($command, base_path(config('twill.vendor_path')));
+        $process = new Process($command, base_path(config('twill.vendor_path')), [
+            'TWILL_ASSETS_DIR' => config('twill.public_directory'),
+        ]);
         $process->setTty(Process::isTtySupported());
 
         if ($disableTimeout) {


haringsrob:
Hey @MauricioKruijer,

I am not entirely sure about this. 

`    'public_directory' => env('TWILL_ASSETS_DIR', 'assets/twill'),` the twill assets dir is supposed to be the directory where in the end the assets end up.

The `    'vendor_path' => 'vendor/area17/twill', ` is where Twill lives.

In the vue config, `const assetsDir = process.env.TWILL_ASSETS_DIR || 'assets/twill'` the twill asset dir is also used in that same way (reading it from the .env).

Are you changing config directly and is that why you need this change?

MauricioKruijer:
> In the vue config, const assetsDir = process.env.TWILL_ASSETS_DIR || 'assets/twill' the twill asset dir is also used in that same way (reading it from the .env).

That is exactly what I am trying to fix. This "reading from the .env" isn't working because the console script isn't passing the variable's properly.

Try setting `TWILL_ASSETS_DIR=assets/cms/admin` in your .env file and run `php artisan twill:build` it will still use the hardcoded fallback `assets/twill`, my PR will expose this value to the vue config.

haringsrob:
Hey @MauricioKruijer, I did check what you mention. 

Set `TWILL_ASSETS_DIR=assets/cms/admin` in my `.env`

Added a console log in the vue.config.js to confirm:
```
const assetsDir = process.env.TWILL_ASSETS_DIR || 'assets/admin'
console.error(assetsDir);
```

And then twill build outputs:

```
🔥 Building Twill assets in production mode.
assets/cms/admin
```

Also in the filetree they end up in the correct spot.

MauricioKruijer:
@haringsrob thank you so much for checking this. This makes me think the problem might be somewhere else (OS related?) which also can explain why some builds are failing in the pipeline...

I am running this on macOS 12.6 and it does not seem to move the assets.

Did you create your own `vue.config.js` in the root of the project? I do not have this file (which means it is using the one inside vendor/ )

ifox:
Hi @MauricioKruijer, no, you don't need a `vue.config.js` file at the root of your project. The test @haringsrob mentioned was done in the vendor one.

PR #1823, opened 2022-09-19
[3.x] duplicate

haringsrob:
Tests & 3.x implementation for https://github.com/area17/twill/pull/1731


tests/integration/Repositories/DuplicateTest.php
```diff
@@ -0,0 +1,175 @@
...
+        $duplicate = $module->getRepository()->duplicate($model->id);
+
+        $this->assertCount(1, $model->repeaterdata);
+        $this->assertCount(1, $duplicate->blocks);
+        $this->assertCount(2, Block::get());
+
+        $this->assertNotEquals($duplicate->blocks->first()->id, $model->blocks->first()->id);
+    }
+    
+    public function testDuplicateWithRevisionsAndRepeaters(): void
```
haringsrob, line 120: Hey @ifox this one currently fails, but I am not sure it should?

When you duplicate a module that has a repeater, should the new model refer the same repeater, clone the repeater or do nothing with the repeater?

Speaking about current behaviour

ifox: Current behavior is creating a new record from a cloned submission, so the expectation is that it would clone the repeater rows as well.

PR #1827, opened 2022-09-21
[2.x] avoid disabled user login

thingasd:
## Description
avoid disabled user login

PR #1828, opened 2022-09-21
[3.x] User login disable

haringsrob:
3.x tests + implementation for https://github.com/area17/twill/pull/1827


thingasd:
Thank you @haringsrob Maybe we can have a toast to notify user like which appear if credentials do not match our records. Also, may I know these changes will backport to 2.x?
image

Issue #1811, opened 2022-09-13
[3.x] Use form builder as default but add --bladeForm as option for new modules


haringsrob:
When we currently make a new module, it will automatically create a form.blade.php

We can switch this to use the formBuilder in the controller instead.

For when more customised forms are required we add the flag --bladeForm (example).

Also remove
```
    protected $indexOptions = [
    ];
```

Issue #1795, opened 2022-08-31
[3.x] Column 'name' does automatically create link to edit


S-ami:
If there is an index column `name`, it automatically becomes a link to module edit route.
If I understood correctly, this happens from JS frontend code and should be fully controllable from backend and `getIndexTableColumns()` function

[Link to Discord message](https://discord.com/channels/811936425858695198/811986149064441927/1014546649181335653)

Twill 3.0.0-alpha1

haringsrob:
Disable the automatic linking and make it form builder only.

haringsrob:
May be related/solve with https://github.com/area17/twill/issues/1761

Issue #1781, opened 2022-08-17
[3.x] Image column takes too much space.


haringsrob:
<img width="890" alt="Screenshot 2022-08-17 at 09 50 20" src="https://user-images.githubusercontent.com/866743/185064446-c83cdb3b-0aab-4c97-bf7b-c15684e0b816.png">

Issue #1761, opened 2022-08-03
[3.x] Permission Roles/Groups listing page showing incorrect name


thingasd:
![image](https://user-images.githubusercontent.com/29519267/182526383-17d5e1f1-7e9e-43fe-9140-20363cbdfc34.png)

PR #1835, opened 2022-09-27
[3.x] Various updates for next alpha

haringsrob:
fixes https://github.com/area17/twill/issues/1761
fixes https://github.com/area17/twill/issues/1795
fixes https://github.com/area17/twill/issues/1811
fixes https://github.com/area17/twill/issues/1781

Fixes issue that would not render the default text(wysiwyg) block.


haringsrob:
@ifox as I did change a bit of vue, can you do a quick review? This pr contains all planned for the next alpha and should resolve all bugs we discussed.

frontend/js/components/table/tableCell/generic/TableCellName.vue
```diff
@@ -1,30 +0,0 @@
-<template>
-  <a :href="editUrl"
-     v-if="!row.hasOwnProperty('deleted') && (row.edit)"
```
ifox, line 3: Is this condition respected by the new table builder? When we link to edit, we should check if the record is not soft deleted.

frontend/js/components/table/tableCell/generic/TableCellName.vue
```diff
@@ -1,30 +0,0 @@
...
-     class="tablecell__name"
-     @click="preventEditInPlace($event)">
```
ifox, line 5: Is the `editInModal` index option still functional after this change?

haringsrob: It was not but the last commit takes care of this.

frontend/js/components/table/tableCell/generic/TableCellName.vue
haringsrob: Was not taken care of but it is now!

PR #1837, opened 2022-09-29
Fix broken link in "Settings Section" of docs

JoshuaCrewe:
There is a link to Twill's configuration documentation which ends in a
404. This is the best guess as to where that should go.


haringsrob:
Perfect, thank you @JoshuaCrewe

PR #1843, opened 2022-10-04
Filter html for block/repeater titles.

haringsrob:
This pr filters html in block/repeater titles.

Screenshot 2022-10-04 at 08 54 42


haringsrob:
Still needs to be ported to 3.x

PR #1845, opened 2022-10-07
fix disabeld field in translated media field and slideshow

thingasd:
## Description

Bring the disabled field to MeidaFieldTranslated and SlideShow


haringsrob:
Hey @thingasd checked and looks good to me. 

I will port this change to 3.x and then merge it.

PR #1859, opened 2022-10-17
[2.x] Add command to allow developers to flush only the twill-manifest.json from cache

antonioribeiro:
I'm trying to solve two problems here:

- While deploying to Vapor, we are running `php artisan twill:update` while building the image on CI, at this point we don't have access to the AWS Redis cache, so we needed a way to clear the twill-manifest.json during the vapor deploy stage, where we cannot run  `php artisan twill:update`, as no files are supposed to change anymore at this point of the deployment.
- `php artisan twill:update` is also clearing the whole application cache and this can be seen as a bad practice, as it's potentially clearing a lot of other application data from cache that developers may want to keep between deployments.

As far as I can tell, Twill is using the Laravel cache only to store the manifest, so that's why I created a separate command to clear this particular cache, and this is also something we can safely run during Vapor deploy stage.

### DIFF
--- a/src/Commands/TwillFlushManifest.php
+++ b/src/Commands/TwillFlushManifest.php
@@ -1,0 +1,37 @@
+<?php
+
+namespace A17\Twill\Commands;
+
+use Illuminate\Console\Command;
+use Illuminate\Support\Facades\Cache;
+
+class TwillFlushManifest extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'twill:flush-manifest';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Flush twill-manifest.json file from cache';
+
+    /**
+     * Execute the console command.
+     *
+     * @return int
+     */
+    public function handle()
+    {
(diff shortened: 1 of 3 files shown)


antonioribeiro:
Hey @ifox , thanks for merging it.

The deployment we are doing on Gitlab is simulating a local directory tree as if we were deploying a Vapor application locally, this is a bit of how it works:

- On CI
- We build (vite/npm/yarn) JS/CSS assets, `composer install` and `php artisan twill:update`
- If `build` passes, Gitlab will use some up-to-date artifacts (`/public`, `/vendor`, `/storybook`) for the next phase, which is `deploy`
- CI runs `vapor deploy`, which will generate a docker image using the artifacts build previously
- Vapor deploy has two deploy sub-phases: `build` and `deploy`
- We could run `php artisan twill:update` on Vapor `build`, but we decided it's just better keep them together with the JS/CSS CI `build` phase
- But even if we did, Vapor `build` phase doesn't have access to the AWS environment yet (we flushing the cache, we would probably get an error as we cannot connect to Redis at this stage too), only Vapor `deploy` phase has, and at this stage we also cannot run any commands that change the filesystem

Issue #1883, opened 2022-11-01
Support Chinese tag


thingasd:
## Summary

Now tag with chinese is not support, probably due to Str::slug($name) return "" when the name is chinese.

## Describe the solution you'd like

Check the $name is chinese and setSlugGenerator to return the original chinese character, not using Str::slug

https://cartalyst.com/manual/tags/12.x#slug-generator

PR #1885, opened 2022-11-01
[2.x] support chinese tag

thingasd:
## Related Issues
#1883

### DIFF
--- a/src/Repositories/Behaviors/HandleTags.php
+++ b/src/Repositories/Behaviors/HandleTags.php
@@ -11,6 +11,12 @@ trait HandleTags
      */
     public function afterSaveHandleTags($object, $fields)
     {
+        if(preg_match("/\p{Han}+/u", $fields['tags'] ?? '')){
+            $object->setSlugGenerator(function($name) {
+                return str_replace(' ', '_', strtolower($name));
+            });
+        }
+
         if (!isset($fields['bulk_tags']) && !isset($fields['previous_common_tags'])) {
             if (!$this->shouldIgnoreFieldBeforeSave('tags')) {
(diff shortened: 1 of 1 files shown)


haringsrob:
Hey @thingasd do you perhaps have an example string for this?

thingasd:
> Hey @thingasd do you perhaps have an example string for this?

do u mean an example tag with Chinese? ”標籤” ”示範”

haringsrob:
Just to confirm, as I am adding tests, if you have 標籤 as a tag, 標籤 should be the slug?

haringsrob:
Because for the case below, the string should build the slug, but it is, with this implementation, just returning the same as the name:

```php            'chinesecharacters3' => [
                'string' => 'test:示@範 foo!',
                'slug' => 'test示at範-foo'
            ],
```

haringsrob:
If we use the suggestion from https://github.com/cviebrock/eloquent-sluggable/issues/392

We get:

```php
                'string' => 'test:示@範 foo!',
                'slug' => 'test-示-範-foo-'
```

Which I believe is a better outcome. What do you think?

thingasd:
> If we use the suggestion from [cviebrock/eloquent-sluggable#392](https://github.com/cviebrock/eloquent-sluggable/issues/392)
> 
> We get:
> 
> ```
>                 'string' => 'test:示@範 foo!',
>                 'slug' => 'test-示-範-foo-'
> ```
> 
> Which I believe is a better outcome. What do you think?
> 
> [8fcf87f](https://github.com/area17/twill/commit/8fcf87ffc582b7ec7c8a0bddca3b7c13349548a9)



> 

YES

PR #1886, opened 2022-11-01
[3.x] fix: Hide nav item if user is not authorized (legacy config)

pboivin:
Small fix to ensure that the `can` key in still supported in legacy `config/twill-navigation.php`:

```php
    'pages' => [
        'title' => 'Pages',
        'route' => 'admin.pages.index',
        'can' => 'access-pages',
    ],
```

### DIFF
--- a/src/TwillNavigation.php
+++ b/src/TwillNavigation.php
@@ -115,11 +115,11 @@ class TwillNavigation
             $link = NavigationLink::make()->forModule($key);
         }
 
-        if ($link && ($legacy['title'] ?? false)) {
+        if ($legacy['title'] ?? false) {
             $link->title($legacy['title']);
         }
 
-        if ($link && ($legacy['can'] ?? false)) {
+        if ($legacy['can'] ?? false) {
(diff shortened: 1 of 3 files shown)


pboivin:
@haringsrob  Looks like a mismatch between chrome/chromedriver versions is failing the Browser tests. Do you know where this can be configured?

haringsrob:
Hey @pboivin,

I fixed dusk in https://github.com/area17/twill/pull/1874

Should be fixed with adding --detect in the main.yml

```      
     - name: Upgrade Chrome Driver
        run: ./vendor/bin/testbench dusk:chrome-driver --detect
```

src/TwillNavigation.php
```diff
@@ -119,6 +119,10 @@ private function transformLegacyToNew(string $key, array $legacy): NavigationLin
             $link->title($legacy['title']);
         }
 
+        if ($link && ($legacy['can'] ?? false)) {
```
haringsrob, line 122: $link should always be set here, not sure why I added this check in line 118

Issue #1887, opened 2022-11-04
Dashboard log reordered activities


thingasd:
## Summary

It is convenient to use drag and drop to control the parent child relation in nested modules, but sometime may drag accidentally, it is helpful to have log of dragging activities

## Describe the solution you'd like
![image](https://user-images.githubusercontent.com/29519267/199890126-11b4d32a-e6c5-4d9c-8a66-ef89aa2c5f3e.png)

Log the reorder activities

## Additional context
Just log the Initiator instead of all saved node?

haringsrob:
As mentioned in the pr. We are not adding this feature at the moment.

PR #1888, opened 2022-11-04
add reordered activity to dashboard

thingasd:
## Related Issues

https://github.com/area17/twill/issues/1887

### DIFF
--- a/src/Jobs/ReorderNestedModuleItems.php
+++ b/src/Jobs/ReorderNestedModuleItems.php
@@ -2,13 +2,13 @@
 
 namespace A17\Twill\Jobs;
 
-
+use A17\Twill\Models\Model;
+use A17\Twill\Models\User;
 use Illuminate\Bus\Queueable;
 use Illuminate\Contracts\Queue\ShouldQueue;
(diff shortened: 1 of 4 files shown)


src/Jobs/ReorderNestedModuleItems.php
```diff
@@ -2,31 +2,37 @@
...
     }
 
     public function handle()
     {
         DB::transaction(function () {
-            app($this->modelClass)->saveTreeFromIds($this->ids);
+            $savedModels = app($this->modelClass)->saveTreeFromIds($this->ids);
+            // Log the reordered models
+            foreach ($savedModels as $model) {
+                activity()->performedOn($model)->causedBy($this->user)->log('reordered');
```
haringsrob, line 34: Wouldnt this cause a huge spam if you would move an item from pos 100 to 1? Since all other items are being moved with this as well?

thingasd: Yes, that is the problem, it should just log the initiator, but I cannot find a simple way do this

PR #1892, opened 2022-11-05
pay attention to use updateRepeaterMorphMany for dealing with polymorphic relationships

YasienDwieb:
## Description

Guide users that they need to use updateRepeaterMorphMany to deal with polymorphic relationships while using repeaters


haringsrob:
Thanks @YasienDwieb

PR #1897, opened 2022-11-08
[3.x] Slugs

haringsrob:
Wip


tests/integration/SlugTest.php
```diff
@@ -0,0 +1,140 @@
...
+        $this->module->getRepository()->delete($model->id);
+
+        // Create a new model after the delete.
+        $newModel = $this->module->getRepository()->create([
+            'title' => 'My title',
+            'slug' => ['en' => 'slug-update'],
+        ]);
+
+        // Total slugs should be 4.
+        $this->assertCount(4, $this->module->getSlugModelClassName()::withTrashed()->get());
```
ifox, line 105: Not sure if I'm missing something, but this should be 3, no?

- `$model` should have 2: `my-title` (inactive and soft-deleted) and `slug-update` (active and soft-deleted)
- `$newModel` should have 1: `slug-update` (active).

haringsrob: You are correct, that is more optimal, this was due to the current flow, which I now have changed.

haringsrob:
@ifox 

https://github.com/area17/twill/actions/runs/3572391896/jobs/6005228337 

I believe we can just update these failing tests as we are generating less slugs now?

PR #1904, opened 2022-11-10
[3.x] Add an autologin feature to improve DX

antonioribeiro:
## Description

This feature will allows developers to skip the login form.

Having way too many Twill websites and moving from one to the other during the day can take a while to type the correct username and passwords on each one. Storing them on password managers is not a very good option as sometimes we have 3 different domains for a single test website and this can quickly bloat our passwords list, mixing development sites with real sites.


docs/content/1_documentation/13_user-management/index.md
```diff
@@ -133,3 +133,14 @@ You can use your new permission and existing ones in many places like the `twill
...
+## Auto login
+
+Developers can configure Twill to do auto login using a pre-defined username and password and skip the login form:
+
+``` dotenv
+TWILL_AUTO_LOGIN_EMAIL=email@email.com
+TWILL_AUTO_LOGIN_PASSWORD=passv0rt
+```
+
+This feature is only available for the `local` environment by default. Please check the `config/twill.php` file for more configuration options.
```
haringsrob, line 146: Since we do not copy the full config stub on install, I feel we should just explain here what needs to be modified with a small example.

src/Http/Controllers/Admin/LoginController.php
```diff
@@ -289,4 +293,43 @@ protected function credentials(Request $request): array
     {
         return array_merge($request->only($this->username(), 'password'), ['published' => 1]);
     }
+
+    /**
+     * Try to autologin a user
```
haringsrob, line 298: I feel the doc-blocks are a bit redundant here, the function names are descriptive enough here.

docs/content/1_documentation/13_user-management/index.md
antonioribeiro: Done

src/Http/Controllers/Admin/LoginController.php
antonioribeiro: Done

src/Http/Controllers/Admin/LoginController.php
```diff
@@ -95,6 +95,10 @@ protected function guard()
      */
     public function showLoginForm()
     {
+        if ($this->autologin()) {
```
haringsrob, line 98: Sorry that I did not comment this before, but can we add a simple check here that only does the autologin attempt if the environment is not production? I think we can safely make the assumption a prod environment should NEVER even attempt to do this.

antonioribeiro: I don't believe we should patronize developers this way, they are big enough to decide what's good or wrong. The example only has `local` (they have to explicitly and manually add a `production` entry in order to have it on production) and it has a comment saying it should be used on local environments only, I agree we can use **red and bold** to make this very clear, but this should be enough, because people must be able to exercise their free will and take ownership of their own fuckups. Also, we cannot predict all use cases, not all production environments will be named `production`, people may name it `prod`. And Twill itself has [autologin on production](https://demo.twill.io/), yes it's a bit different, but it's still autologin on a slightly different use case.

kallefrombosnia: Twill uses the naming `production` for the live deployment of app all over the source code. So in my opinion, simple naming convection should be mentioned in the starting [docs](https://github.com/area17/twill/blob/844d38a0bb698974b280b0b3d1f5ba26472e0c1a/docs/src/getting-started/index.md), and simply check in this `development` only feature if this is not a production setup. I've had the same issue with this before, this can be really helpful.

antonioribeiro: Why prevent people from doing what they explicitly want?

Let's say I just want to create a demo CMS and put in production, why should I be prevented from doing this just because the Twill team decided this feature is not to be used in production?

We could maybe add another `enabled` configuration key and tie it to the `APP_DEBUG` var, to make sure people would have to put the app in debug mode and manually add `production` in the `environments` array in order to have it enabled in production, but forbidding people from doing it, is patronizing, is not something I believe we should do.

ifox: Twill currently does not enforce anything depending on the environment or the debug feature, outside of logging or exception throwing, so I would tend to agree with @antonioribeiro, though I want to make sure enabling this is done explicitly by the developer.

antonioribeiro: Moved the configuration to an auto-merged config and added the `enabled`, and a docs warning about the danger of running this in production. 

Screenshot 2022-11-17 at 15 08 15

So now in order to have it in production the developer will have to:

- Find and copy the config array from `config/autologin.php` to `twill.php` adding the `autologin` entry
- Change `APP_DEBUG=true` or change the `enabled` entry to `true`
- Add the `production` to the environments entry
- Create an user in production
- Add username and password of this user to the .env file

If a developer does all that, they can't blame us for anything, they wanted to have it in production and I believe they have the right to have.

ifox: If you're curious about the autologin and "read-only" feature on demo.twill.io, here's the source code of the middleware:

```php
<?php

namespace App\Http\Middleware;

use A17\Twill\Helpers\FlashLevel;
use Auth;
use Closure;

class Demo
{

    public function handle($request, Closure $next)
    {
        if (!Auth::guard('twill_users')->user()) {
            if (session()->has('hasReadDisclaimer')) {
                Auth::guard('twill_users')->loginUsingId(1);
            } else {
                session([
                    'hasReadDisclaimer' => true,
                    'demoRedirectUrl' => session('url.intended'),
                ]);

                return redirect(route('admin.demo'));
            }
        }

        $route = $request->route()->getName();

        // accept requests to blocks previews
        if ($route === 'admin.blocks.preview') {
            return $next($request);
        }

        // accept requests to modules previews
        if (ends_with($route, '.preview')) {
            return $next($request);
        }

        $demoMessage = 'Whoa! This is a demo, my friend. You can preview your edits, but not save.';

        // only GET requests are accepted
        if ($request->method() !== 'GET') {
            if ($request->expectsJson()) {
                return response()->json([
                    'message' => $demoMessage,
                    'variant' => FlashLevel::SUCCESS,
                ]);
            }

            session()->flash('status', $demoMessage);

            return redirect()->back();

        }

        return $next($request);
    }
}
```

src/Http/Controllers/Admin/LoginController.php
```diff
@@ -289,4 +293,42 @@ protected function credentials(Request $request): array
...
+        ], false);
+    }
+
+    protected function autologinEnabled(): bool
+    {
+        if (!$this->config->get('twill.autologin.enabled', false)) {
+            return false;
+        }
+
+        dd('enabled');
```
ifox, line 315:
```suggestion

```

src/Http/Controllers/Admin/LoginController.php
antonioribeiro: @ifox instead of APP_DEBUG, which can be left to true by mistake on production, we can go with a 

```
'enabled' => env('TWILL_AUTO_LOGIN_ENABLED', false),
```

Or both, whatever you prefer.

src/Http/Controllers/Admin/LoginController.php
antonioribeiro: Yeah, fixed, sorry :(

src/Http/Controllers/Admin/LoginController.php
antonioribeiro: Cool! Done.

PR #1923, opened 2022-11-17
[2.x] Fix edit() not using the $id being passed as parameter

antonioribeiro:
## Description

I just found that `ModuleController::edit()` is not using the `$id` that we can pass as parameter to it, in favor of a route parameter that may not even exist.

Having a route like so:

```
https://twill.io/global/moduleLive
```

Baked by this controller method:

``` php
public function simpleEdit()
{
    return $this->edit($this->pageKey);
}
```

Will result on a 404 because `last($params)` is `false`.

This PR fixes this by also trying to find the record by the `$id` (which in my opinion should be primary, but it can lead to BC breaks) if the route id is not found.


antonioribeiro:
Hey @haringsrob , me and @ifox looked into this together. I found that other methods have the same problem and just pushed to some fixes.

Trying to explain it a little better: we are losing the original `$id` passed to the method because of these lines:

``` php
$params = $this->request->route()->parameters();

...

$id = last($params);
```

These methods are assuming there will always be an id passed as URL parameter, but if we pass the `$id` from this signature here: 

``` php
public function update($id, $submoduleId = null);
```

By calling the controller method directly

``` php
public function simpleEdit()
{
    return $this->edit($this->pageKey);
}
```

The `$id` will be null, because my (kind of a singleton) route doesn't have the id.

It keeps trying to use `$id` as a submodule id, but it doesn't mean anything for root modules calling these methods directly:

``` php
$submoduleParentId = $this->getParentModuleIdFromRequest($this->request) ?? $id;
```

haringsrob:
Hey @antonioribeiro I have taken the same approach as in the other pr, could you give it a try?

antonioribeiro:
Just did, @haringsrob , it passed all my tests! Thank you!

haringsrob:
Awesome!

PR #1954, opened 2022-11-25
Submodule improvements.

haringsrob:
Reworks nested handeling

This needs to be ported to https://github.com/area17/twill/pull/1923


src/Http/Controllers/Admin/ModuleController.php
```diff
@@ -2372,19 +2360,30 @@ protected function getModulePermalinkBase(): string
...
+        $prev = [];
         foreach ($moduleParts as $index => $name) {
             if (array_key_last($moduleParts) !== $index) {
                 $singularName = Str::singular($name);
                 $modelClass = config('twill.namespace') . '\\Models\\' . Str::studly($singularName);
 
                 if (!class_exists($modelClass)) {
-                    $modelClass = TwillCapsules::getCapsuleForModel($name)->getModel();
+                    // First try to construct it based on the last.
+                    $modelClass = config('twill.namespace') .
```
haringsrob, line 2371: This should be attempted first as it is stricter.

src/Http/Controllers/Admin/ModuleController.php
```diff
@@ -1098,39 +1098,28 @@ public function show($id, $submoduleId = null)
...
         if ($id instanceof TwillModelContract) {
             $item = $id;
             $id = $item->id;
         } else {
-            $id = last($params);
-
-            $item = $this->repository->getById($submoduleId ?? $id);
+            $parameter = Str::singular(Str::afterLast($this->moduleName, '.'));
+            $id = $this->request->route()->parameter($parameter, $id);
+            $item = $this->repository->getById($id, $this->formWith, $this->formWithCount);
```
haringsrob, line 1109: @antonioribeiro this is a slightly simplified version to handle the id's.

This avoids the use of last and is more strict. If there is no argument for the current module in the url, it defaults to the id being passed in. This (should) work for you case.

PR #1956, opened 2022-11-27
1955: Honour position/order in nested blocks

### LINKED ISSUE #1955: [3.x] BlockRenderer does not render nested blocks in order of position
## Description

When fetching root blocks, the default behaviour is to order by position. However, the same cannot be said for nested blocks, as the query simply fetches by parent_id without any ordering.

### Steps to reproduce

1. Create a block that uses a block editor that references any block type.
2. Add an instance of that block to a module, such as a page.
3. Create a template for both blocks, and use `renderChildren(name)` in the root block template.
4. Add some child blocks and re-order them. The order will not be as intended.

### Expected result

Nested blocks at all levels should rendered be in the order defined.

### Actual result

Nested blocks are ordered as per the database’s default ordering.

### Steps to correct

In BlockRenderer, exchange the `$childBlocks` assignment in `getNestedBlocksForBlock` to:

```php
$childBlocks = A17Block::whereParentId($block->id)->orderBy(
    config('twill.blocks_table', 'twill_blocks') . '.position',
    'asc'
)->get();
```

(May I open a PR against 3.x?)

### Versions

Twill version: 3.0.0-beta2
Laravel version: 9.x
PHP version: 8.1
Database engine: Postgres 14

mikerockett:
Fixes #1955

### DIFF
--- a/src/Helpers/BlockRenderer.php
+++ b/src/Helpers/BlockRenderer.php
@@ -230,10 +230,7 @@ class BlockRenderer
         $class = Block::findFirstWithType($block->type);
 
         /** @var \A17\Twill\Models\Block[] $childBlocks */
-        $childBlocks = A17Block::whereParentId($block->id)->orderBy(
-            config('twill.blocks_table', 'twill_blocks') . '.position',
-            'asc'
-        )->get();
+        $childBlocks = A17Block::whereParentId($block->id)->orderBy('position')->get();
 
         $children = [];
 
--- a/src/Helpers/BlockRenderer.php
+++ b/src/Helpers/BlockRenderer.php
@@ -230,7 +230,10 @@ class BlockRenderer
         $class = Block::findFirstWithType($block->type);
 
         /** @var \A17\Twill\Models\Block[] $childBlocks */
-        $childBlocks = A17Block::whereParentId($block->id)->get();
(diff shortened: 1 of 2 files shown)


src/Helpers/BlockRenderer.php
```diff
@@ -230,7 +230,10 @@ private static function getNestedBlocksForBlock(
         $class = Block::findFirstWithType($block->type);
 
         /** @var \A17\Twill\Models\Block[] $childBlocks */
-        $childBlocks = A17Block::whereParentId($block->id)->get();
+        $childBlocks = A17Block::whereParentId($block->id)->orderBy(
+            config('twill.blocks_table', 'twill_blocks') . '.position',
```
haringsrob, line 234: Makes sense, however, I wonder, do we really need to recall the block table name? I feel like that's already assumed because of the A17Block model?

haringsrob: Hey, yes please

haringsrob: Hey @mikerockett the tests are indeed unrelated, will work on that today

haringsrob:
Hey @mikerockett tried to update your branch but I think there's no permission. Can you merge latest 3.x? That should resolve the tests.

PR #1959, opened 2022-11-30
[2.x][Hotfix] Fix laravel 9 compatibility

haringsrob:
Laravel added singleton as a route in the last release.

As we also use that as a macro we have to introduce this breaking change.

To add to release notes:

To fix replace all:
- `Route::singleton` -> `Route::twillSingleton`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1551,7 +1551,7 @@ It also gives up a slight performance boost in HMR mode when developing.
 - Replace image in media form field
 - Context based image alternative text and caption
 - Restore soft deleted models
-- Suggested frontend controller for show and preview routes with its associated router macro (Route::twillModuleShowWithPreview)
+- Suggested frontend controller for show and preview routes with its associated router macro (Route::moduleShowWithPreview)
 - Allow custom buckets routes prefixes, default to "featured" prefix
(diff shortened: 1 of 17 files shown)


haringsrob:
@ifox anything I may be missing here?

ifox:
Nope, looking good! 

The only thing is that we're not supposed to do breaking changes on 2.x but it wouldn't make sense for it to be 3 either.

Considering the change is very simple to apply for consumers, I think this is fine to do as a minor. We'll just blame Laravel for stealing our route macro name if someone isn't happy about it 🙃

haringsrob:
It is indeed a breaking change but required to continue to support Laravel 9. Can we message users on composer update somehow?

haringsrob:
> Maybe overkill, but what do you think about providing a Rector rule for it? I guess we'll want it for Twill 3 anyway, so it could be nice to ship it here too.

This type of change cannot be changed with rector that easily I am afraid. It's a simple find-and-replace, we could provide an example on how to do that? But this as well seem's too much, it is/should be just  a change in routes/admin.php

ifox:
https://gist.github.com/delenamalan/3c0d998a316031af8f0f9c0165eacc28

PR #1963, opened 2022-11-30
[3.x] Form utils support in Twill 3's controller form definitions

ifox:
This PR adds support for form utilities with Twill 3 new approach to defining the form from the controller:
- columns, using the new `A17\Twill\Services\Forms\Columns` class, which can be used like this within fields in `getForm`:
```php
Columns::make()
    ->left([
        Input::make()
            ->name('description')
            ->translatable(),
    ])
    ->right([
        Input::make()
            ->name('subtitle')
            ->translatable(),
    ]),
```
- fieldsets, using the new `A17\Twill\Services\Forms\Fieldsets` and `A17\Twill\Services\Forms\Fieldset` classes, which can be used like this and returned in `getForm`:
```php
Fieldsets::make()->push(
    Fieldset::make()
        ->title('Meta')
        ->fields([
            Input::make()->name('description')->translatable(),
            Medias::make()->name('image')
        ]),
    Fieldset::make()->title('Content')->closed()->fields([
        BlockEditor::make()->withoutSeparator(),
    ])
);
```
- side fieldsets, using the new `getSideFieldsets` method of a module controller, from which you can return a `Form` or a `Fieldsets` object, depending on the need for multiple side fieldsets or not
- custom blade partials, using the new `A17\Twill\Services\Forms\BladePartial` class, which effectively also adds support for conditional fields by mixing the controller and view approaches, eg.:

In the controller:
```php
Form::make()->push(
    Checkbox::make()->name('condition_field'),
    BladePartial::make()->view('twill.fields.field-under-condition')
)
```
In `resources/views/twill/fields/field-under-condition.blade.php`
```php
<x-twill::formConnectedFields
    field-name="condition_field"
    :field-values="true"
    :render-for-blocks="false"
>
    <x-twill::input
        name="field_under_condition"
        label="A field under a condition from a blade partial in Twill 3 forms"
    />
</x-twill::formConnectedFields>

```

Screenshot 2022-11-30 at 5 27 14 PM

TODOs:
- [x] Documentation


src/Http/Controllers/Admin/ModuleController.php
```diff
@@ -2671,7 +2690,12 @@ protected function getTransLabel($key, $replace = [])
         return twillTrans(Arr::has($this->labels, $key) ? Arr::get($this->labels, $key) : $key, $replace);
     }
 
-    public function getForm(TwillModelContract $model): Form
+    public function getForm(TwillModelContract $model): Form | Fieldsets
+    {
+        return new Form();
```
haringsrob, line 2695: Just an idea, but what if we `Form::fieldsets([])` this would avoid a separate class/union type?

Then we can use $formController->fieldsets() as our condition instead of the instanceof?

views/layouts/form.blade.php
```diff
@@ -149,37 +190,37 @@
 
 @section('initialStore')
     window['{{ config('twill.js_namespace') }}'].STORE.form = {
```
haringsrob, line 192: I have wanted to do this many times, but now I will. We are going to move these store initializers to a helper class as arrays and just print them out here, what do you think?

src/Http/Controllers/Admin/ModuleController.php
```diff
@@ -1154,10 +1155,28 @@ public function edit(TwillModelContract|int $id, $submoduleId = null)
             }
         }
 
-        return View::make($view, $this->form($id))->with(
-            'renderFields',
-            $controllerForm
-        );
+        $viewWithData = [];
```
haringsrob, line 1158: Can we move this out of the edit method maybe? Just to avoid it from growing.

Perhaps a `buildFormBuilderData(TwillModelContract $item): array` method

views/layouts/form.blade.php
haringsrob: (can be a separate pr)

haringsrob (changes requested):
So nice! Left a few suggestions.

src/Http/Controllers/Admin/ModuleController.php
ifox: This is the suggestion I was looking for. Wasn't happy about the union type but wasn't sure how to avoid it.

views/layouts/form.blade.php
ifox: I agree, but let's make sure they are easily changed. I'm sure a lot of users are overriding the layout to adjust them.

src/Http/Controllers/Admin/ModuleController.php
ifox: Yes, definitely

haringsrob: Hey, I have merged https://github.com/area17/twill/pull/1971 so we also have to apply this there, so the abstraction is for sure needed :).

haringsrob (approved):

PR #1980, opened 2022-12-07
[2.x] Apply maxlength property for textarea input type

zeezo887:
## Description

This PR applies the max length property set in an input form field of type textarea


ifox:
@zeezo887 this change makes the textarea prevent admins from entering more characters than specified, correct? In order to introduce this behavior, we need to preserve the current one and add a new option to enable it, because it is intended that we're not preventing entering more characters at the moment, with the visual indicator showing to the admin that they are above the limit.

zeezo887:
@ifox I see, So we could do something like this
`@formField('input', [
    'maxlength' => 250,
    'strictMaxlength' => true
    'type' => 'textarea',
    'rows' => 3
])`

`strictMaxlength` will default to false

ifox:
Exactly, that'd be great!

PR #1997, opened 2022-12-13
[3.x] Capsules/packages fixes: permissions compatibility and route macros

joyceverheije:


src/Helpers/modules_helpers.php
```diff
@@ -36,7 +36,7 @@ function getModelByModuleName($moduleName)
             try {
                 $model = \A17\Twill\Facades\TwillCapsules::getCapsuleForModule($moduleName)->getModel();
             } catch (NoCapsuleFoundException) {
-                throw new Exception($model . ' not existed');
```
haringsrob, line 39: 🤦

src/RouteServiceProvider.php
```diff
@@ -195,6 +197,32 @@ private function registerRouteMiddlewares()
         Route::aliasMiddleware('permission', Permission::class);
     }
 
+    /**
+     * Registers Route macros.
+     *
+     * @return void
+     */
+    protected function registerMacros()
```
haringsrob, line 205: Ah yes I was still supposed to do this! Please add void as a return type, then we can remove it from the docblock

joyceverheije: Thank you! I made that change.

PR #2007, opened 2022-12-16
[3.x] Block components.

haringsrob:
Quickly made a thing this morning! Standalone component blocks. 

These blocks extend TwillBlockComponent and have no view of their own. The formBuilder is used for rendering.

There will also be a method "renderUsing" which takes care of the rendering. So everything in a single file!

~~These components are loaded using a namespace like this: `TwillBlocks::registerComponentBlocks('\App\View\Components\Twill');`~~

this means packages can ship their own fully self-contained blocks.
This also means, packages can become templates as the full markup can be shipped easily.

- [x] docs
- [x] Generate command


src/TwillBlocks.php
```diff
@@ -87,21 +105,50 @@ public function getBlockCollection(): BlockCollection
...
                 $this->blockCollection->add($block);
             }
 
             unset(self::$blockDirectories[$blockDir]);
         }
 
+
+        if (self::$componentBlockNamespaces !== []) {
+            foreach (self::$componentBlockNamespaces as $namespace) {
+                $path = Str::replace('\\', '/', $namespace);
```
haringsrob, line 132: This "should" break on case sensitive filesystems.

haringsrob:
Hey @ifox,

I think this is ready to be the final feature we add in 3.x :).

Let me know what you think.

haringsrob:
Bugs:

return new self($this->file, $this->type, $this->source, $this->name, $this->renderNamespace); -> Missing parameters

PR #2012, opened 2022-12-19
fix(2fa): qrcode generation

tuanbinhtran:
## Description

Fixes the qr code generation
Reverts to the change in version 2.x

`$this->request->get('user')` always returns NULL on my instance


haringsrob:
Hey @tuanbinhtran thank you. This is indeed a regression.

PR #2014, opened 2022-12-20
[2.x] Fixed BlockTitle max-width not working in fieldset by using CSS

domihagen:
## Description

The max-width was always 0px for Blocks inside of a fieldset until a window.resize triggered.

This PR removes the max-width calculation with JS and uses CSS-Only now.

## Related Issues

https://github.com/area17/twill/pull/1843


haringsrob:
Cherry-picked into https://github.com/area17/twill/pull/2013 as well for 3.x

PR #2016, opened 2022-12-21
Make A17Block to be configurable class in some non-configurable places

deckchan:
## Description
Twill should accept A17\Twill\Models\Block to be configurable as mentioned in https://github.com/area17/twill/blob/3.x/config/models.php

### DIFF
--- a/src/Helpers/BlockRenderer.php
+++ b/src/Helpers/BlockRenderer.php
@@ -97,7 +97,7 @@ class BlockRenderer
         }
 
         // Load the original block if it exists or make a new one then fill it with the data from the request.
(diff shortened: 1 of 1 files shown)


ifox:
Hi @deckchan the change to make the block model configurable makes sense and I would like to approve it, however I'm wondering why the order by was removed?

deckchan:
> Hi @deckchan the change to make the block model configurable makes sense and I would like to approve it, however I'm wondering why the order by was removed?

Oh no!! Sorry it was a mistake, I will commit it back.

By the way, should I make other A17Block to be configurable like this way?

Thanks so much!

PR #2018, opened 2022-12-22
[3.x] Make A17Block to be configurable class

deckchan:
## Description

Twill should accept A17\Twill\Models\Block to be configurable as mentioned in https://github.com/area17/twill/blob/3.x/config/models.php

## Related Issues

Remake for https://github.com/area17/twill/pull/2016


haringsrob:
hey @deckchan thanks for your contribution. This makes a lot of sense!

PR #2024, opened 2022-12-25
[2.x] Fix for VSelect component to use float values

eudaco:
## Description
Fixed handling for float values in Select component.
Previously, all Number values were handled always as int.
If you used float, the Select field just didn't determine the stored value.


haringsrob:
hey @eudaco can you share with me how you use your field?

eudaco:
> 

Hi @haringsrob ,
For example, when I use a field for rating, I have a set of values like this:

` @formField('select', [
            'name' => 'rating.value',
            'label' => 'Rating',
            'unpack' => false,
            'options' => [
                [
                    'value' => 5.0,
                    'label' => '5.0 / 5.0'
                ],
                [
                    'value' => 4.9,
                    'label' => '4.9 / 5.0'
                ],
                [
                    'value' => 4.8,
                    'label' => '4.8 / 5.0'
                ],
                [
                    'value' => 4.7,
                    'label' => '4.7 / 5.0'
                ],
                [
                    'value' => 4.6,
                    'label' => '4.6 / 5.0'
                ],
                [
                    'value' => 4.5,
                    'label' => '4.5 / 5.0'
                ],
                ...
            ]
        ])`

In this case, the field after saving the value doesn't work correctly, and the saved value isn't displayed (or displayed incorrectly)

PR #2028, opened 2023-01-02
Allow multi-select to act as tags

antonioribeiro:
## Description

As I could not find a way to use any of the Twill selects as a "tag" select inside a repeater, this PR provides some features:

1. Allows us to use the multi-select form field as a tags, just by exposing some of the props used by the tags form field:

```blade
@formField('multi_select', [
    'name' => 'custom_tags',
    'label' => 'Custom tags',
    'unpack' => false,
    'taggable' => true,
    'searchable' => true,
    'endpoint' => '/global/homepages/custom-tags',
])
```

Screenshot 2023-01-02 at 17 47 47

2. Allows searching on the multi-select.

3. Fixes the need to always pass `$options` to the multi-select, even when we were passing `$endpoint`


haringsrob:
This will only be added in 3.x as that is the main focus for new features.

https://github.com/area17/twill/pull/2068 commits have been transferred.

Thanks for contributing!

PR #2045, opened 2023-01-06
[3.x] Add ability to accepts parameters when using BladePartial directive

sauron:
Add ability to accepts parameters when using the FormBuilder to include blade partials.

## Description

Right now we have the ability to use BladePartial in the FormBuilder for including view, but it is lacking the ability to send parameters/options to those views.
This pull requests aims to solve that situation.

## Example

With this PR we'll be able to do:
```php
use A17\Twill\Services\Forms\BladePartial;

...

    BladePartial::make()
        ->view('twill.shared.link', [
           'field_prefix' => 'recirculation',
           'label_note' => 'Optional. Page title will be used if empty.',
        ])
```

### DIFF
--- a/src/Services/Forms/BladePartial.php
+++ b/src/Services/Forms/BladePartial.php
@@ -9,6 +9,7 @@ class BladePartial
 {
     protected function __construct(
         protected ?string $view = null,
+        protected ?array $params = [],
     ) {
     }
 
@@ -17,9 +18,10 @@ class BladePartial
         return new self();
     }
 
-    public function view(string $view): static
+    public function view(string $view, array $params = []): static
     {
         $this->view = $view;
+        $this->params = $params;
 
         return $this;
     }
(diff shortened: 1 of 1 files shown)


src/Services/Forms/BladePartial.php
```diff
@@ -30,13 +32,19 @@ public function render(): ?View
...
-            'item' => $form['item'] ?? null,
-            'form_fields' => $form['form_fields'] ?? [],
-            'formModuleName' => $form['moduleName'] ?? null,
-            'routePrefix' => $form['routePrefix'] ?? null,
-        ]);
+        $form = ViewFacade::shared('form');
+
+        return ViewFacade::make(
+            $this->view,
+            array_merge(
```
ifox, line 39: Thanks @sauron this is a great and needed addition. I wonder if it would be better to reverse the parameters in this merge, so that, if one needs that, they can override the parameters that Twill is passing by default.

ifox (changes requested):

src/Services/Forms/BladePartial.php
sauron: Makes total sense!
Done!

haringsrob:
Hey!

I feel we could maybe split this into a second parameter, as we currently provide one method for each action.
```php
BladePartial::make()->view('bar')->withAdditionalParameters(['bar' => 'foo'])
```

We could on top of that allow to add a closure option that we can provide the model in?
```
BladePartial::make()->view('bar')->withAdditionalParameters(fn(TwillModelInterface $model) => $model->attributes)
```

sauron:
> Hey!
> 
> I feel we could maybe split this into a second parameter, as we currently provide one method for each action.
> 
> ```
> BladePartial::make()->view('bar')->withAdditionalParameters(['bar' => 'foo'])
> ```
> 
> We could on top of that allow to add a closure option that we can provide the model in?
> 
> ```
> BladePartial::make()->view('bar')->withAdditionalParameters(fn(TwillModelInterface $model) => $model->attributes)
> ```
> 
> What do you think?

Makes total sense to me @haringsrob !
I took care of the new `withAdditionalParameters` function for now. I think we can grow from there.

haringsrob:
hey @sauron 

Perfect, indeed we'll see what's requested in the future regarding other features.

PR #2069, opened 2023-01-16
[3.x] Fix modules, webpack 5, live/hot reloading.

haringsrob:
This updates vue cli to the latest version
Fixes live reload as well.


config/twill.php
```diff
@@ -190,7 +190,7 @@
     'dev_mode' => env('TWILL_DEV_MODE', false),
     'dev_mode_url' => env('TWILL_DEV_MODE_URL', 'http://localhost:8080'),
     'public_directory' => env('TWILL_ASSETS_DIR', 'assets/twill'),
-    'manifest_file' => 'twill-manifest.json',
+    'manifest_file' => 'assets/twill/twill-manifest.json',
```
haringsrob, line 193: Not sure if this has any side effect yet.

@ifox

vue.config.js
```diff
@@ -107,17 +107,17 @@ if (!isProd) {
 }
 
 // Define npm module resolve order: 1. local (Twill), 2. root (App)
-const appModuleFolder = path.resolve(__dirname, '../../../node_modules') // vendor/area17/twill/
-const resolveModules = ['node_modules']
-if (fs.existsSync(appModuleFolder)) {
-  resolveModules.push(appModuleFolder)
-}
+// const appModuleFolder = path.resolve(__dirname, '../../../node_modules') // vendor/area17/twill/
```
haringsrob, line 110: @ifox No side effect on the changes I did in this file afaik. But maybe they are for something special I am not aware of.

vue.config.js
```diff
@@ -133,39 +110,25 @@ const config = {
...
+    },
+    devtool: false,
     resolve: {
       alias: {
-        'prosemirror-tables': path.join(__dirname, 'node_modules/prosemirror-tables/src/index.js'),
-        'prosemirror-state' : path.join(__dirname, 'node_modules/prosemirror-state/src/index.js'),
-        'prosemirror-view' : path.join(__dirname, 'node_modules/prosemirror-view/src/index.js'),
-        'prosemirror-transform' : path.join(__dirname, 'node_modules/prosemirror-transform/src/index.js')
-      },
-      modules: resolveModules
```
ifox, line 150: We need to keep this. This allows building Twill with components importing dependencies installed in the host project's node modules.

package.json
```diff
@@ -23,46 +23,44 @@
     "fine-uploader-wrappers": "^1.1.0",
     "flatpickr": "4.6.13",
     "lodash": "^4.17.21",
-    "quill": "1.3.7",
+    "quill": "^1.3.7",
```
ifox, line 26: I'm a bit worried by this as there was a reason we locked Quill's version before. I'll take a closer look.

ifox (changes requested):

package.json
ifox: Oh there was no release, ok. Quill feels clearly abandoned at this point.

haringsrob: Maybe Twill 3 is a good moment to ditch one of the wysiwyg implementations?

ifox: We will need to keep Quill available as most of our users are relying on it and migrating content would be painful, since, for example, Quill's output includes custom classes for some of its features. If the link feature is completed on Tiptap, I would be willing to set it as the new default.

haringsrob: I guess Quill 2 could also be an option once it's done. Anyway, will leave it for now.

config/twill.php
haringsrob: This was changed back to the default.

PR #2004, opened 2022-12-15
Added foreign key to email in password_resets table

baccadiginepro:
## Description

This PR adds a foreign key to email field of twill_password_resets table and allows the password reset token to be deleted when the corresponding user is hard deleted.
This change allows a user deleted before he had set his password, to receive a new token and a new invitation email.

PR #2070, opened 2023-01-17
[3.x] Password reset new existing users

haringsrob:
Port of https://github.com/area17/twill/pull/2004


src/Models/Behaviors/HasNesting.php
```diff
@@ -32,23 +32,17 @@ public function getNestedSlugAttribute()
...
      * Returns the combined slug for all ancestors of this item.
-     *
-     * @param string|null $locale
-     * @return string
      */
-    public function getAncestorsSlug($locale = null)
+    public function getAncestorsSlug(?string $locale = null): string
     {
-        return collect($this->ancestors->sortByDesc('position') ?? [])
+        return $this->getAncestors()->reverse()
```
haringsrob, line 38: This should be the correct order according to the docs.

PR #2071, opened 2023-01-17
Fix site-link key

vladimirmartsul:
Fix mistype in key `config('twill.enabled.site-link', false)` at https://github.com/area17/twill/blob/0118fe0e641cbd407664b46a15e9ff35e2418137/src/TwillNavigation.php#L180
There is should be `twill.enabled.site-link` instead of `twill.enable.site-link`


haringsrob:
Thanks for spotting this @vladimirmartsul

PR #2072, opened 2023-01-17
Fix searchColumns overriding

vladimirmartsul:
Assigning `$this->searchColumns = [$this->titleColumnKey];` (https://github.com/area17/twill/blob/3.x/src/Http/Controllers/Admin/ModuleController.php#L401) immediately overwrites `$this->searchColumns` from call `$this->setUpController();` several lines before.
This simple patch fixes this and merges `$this->titleColumnKey` with the previously seted up `$this->searchColumns`.

### DIFF
--- a/src/Http/Controllers/Admin/ModuleController.php
+++ b/src/Http/Controllers/Admin/ModuleController.php
@@ -398,7 +398,7 @@ abstract class ModuleController extends Controller
             $this->getSideFieldsets($this->repository->getBaseModel())->registerDynamicRepeaters();
         }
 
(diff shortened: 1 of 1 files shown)


haringsrob:
Hey @vladimirmartsul 

Thank you for your contribution. I also ran into this issue and I resolved it here:

https://github.com/area17/twill/pull/2070/files#diff-8a3084d617335690e3d80eadc43ada1afa56002be98b5b69ca778ead9a4849ec

We should not use the merge as you proposed but instead move the default up a bit allowing it to be overwritten entirely.

If it is still unset at the point of the current setter, we can still add the value.

PR #1889, opened 2022-11-04
add dashboard pagination

thingasd:
## Description
Pagination for dashboard, this pr is not perfectly done, some reviews and modification are needed, especially for paginating my activity cause the filtering seems applied in frontend.
image

## Related Issues

https://github.com/area17/twill/issues/1800

PR #2078, opened 2023-01-20
[3.X] Dashboard pagination

haringsrob:
3.x port of https://github.com/area17/twill/pull/1889

Screenshot 2023-01-20 at 10 34 29


frontend/js/components/dashboard/activityFeed.vue
```diff
@@ -71,14 +73,15 @@
       })
     },
     methods: {
+      getData(pageNumber) {
+        this.$http.get('admin?' + this.navFilters[this.navActive].slug + '=' + pageNumber).then(({data}) => {
```
haringsrob, line 77: @ifox do we have a helper to get the "admin" url?

config/dashboard.php
```diff
@@ -24,5 +24,5 @@
         'login' => false,
         'logout' => false,
     ],
-    'auth_activity_causer' => 'users',
+    'auth_activity_causer' => 'users'
```
ifox, line 27: Is our formatter config removing trailing commas? If you're not against them @haringsrob I would prefer to keep them. Diffs are more readable and it is less typing when adding a new line.

frontend/js/components/dashboard/activityFeed.vue
ifox: Could we use the document location here?

ifox (changes requested):

config/dashboard.php
haringsrob: Will add a standard.

frontend/js/components/dashboard/activityFeed.vue
haringsrob: Decided to pass the base url from the backend to ensure it is correct. https://github.com/area17/twill/pull/2078/commits/92de5fc12f580c115eb3191ac85d6595a85cdbc3

config/dashboard.php
haringsrob: Not super relyable but I added https://github.com/squizlabs/PHP_CodeSniffer/issues/582

PR #2082, opened 2023-01-21
Allow creating blocks and repeaters inside Capsules

antonioribeiro:
## Description

This PR adds the ability for Capsules to register blocks and repeaters.

Screenshot 2023-01-21 at 17 54 19

Screenshot 2023-01-21 at 17 59 30

Screenshot 2023-01-21 at 17 46 14

### DIFF
--- a/src/Helpers/Capsule.php
+++ b/src/Helpers/Capsule.php
@@ -112,6 +112,8 @@ class Capsule
     public function registerViews(): void
     {
         View::addLocation(Str::replaceLast(DIRECTORY_SEPARATOR . $this->name, '', $this->path));
+
+        $this->registerBlocksAndRepeatersViewPaths();
     }
 
     public function loadMigrations(): void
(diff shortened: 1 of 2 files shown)


src/Helpers/Capsule.php
```diff
@@ -365,4 +367,22 @@ public function getType(): string
     {
         return '';
     }
+
+    public function registerBlocksAndRepeatersViewPaths(): void
+    {
+        $resourcePath = $this->getConfig()['views_path'] ?? 'resources/views/admin';
```
haringsrob, line 373: Hey @antonioribeiro 

Was having a look at this. But I feel we do not need the admin part here.

As capsules are already specific to twill, we could just have `resources/views/[blocks|repeaters]`

What do you think?

antonioribeiro: I believe we do, @haringsrob , we are following this structure on some of our Capsules: https://github.com/area17/twill-google-recaptcha/tree/1.x/src/resources/views.

But how can we make this flexible in order to support any structure? Some other people might structure their Capsules other ways...

PR #2088, opened 2023-01-23
[3.x] Make `skipCreateModal` compatible with Twill 3 form builder

joyceverheije:


haringsrob:
Thanks! That's something completely overlooked :). I added a commit with some cleanup as there was some code duplication still there.

PR #2093, opened 2023-01-24
[3.x] Fix password resets constraint down migration

joyceverheije:
In a test suite that uses `use DatabaseMigrations;`, the down method of this migration was failing with:

```
PDOException: SQLSTATE[42704]: Undefined object: 7 ERROR:  constraint "email" of relation "twill_password_resets" does not exist
```


haringsrob:
checked and works great. I am not sure why the original does not as it should work with both.

PR #2113, opened 2023-01-27
Allow disabling automatic navigation on a Capsule package

antonioribeiro:
Currently the Capsule package injects itself automatically into the main navigation, this is not desirable on larger apps with lots of menu options.


antonioribeiro:
@haringsrob , these compatibility checks are great, but it's currently abending on Twill versions 😅

PR #2115, opened 2023-01-29
Allow configuring crops from block components.

haringsrob:
This pr allows block components to define their crops inline instead of using the config file:

```php
<?php

namespace App\View\Components\Twill\Blocks;

use A17\Twill\Services\Forms\Fields\Input;
use A17\Twill\Services\Forms\Fields\Medias;
use A17\Twill\Services\Forms\Fields\Wysiwyg;
use A17\Twill\Services\Forms\Form;
use A17\Twill\View\Components\Blocks\TwillBlockComponent;
use Illuminate\Contracts\View\View;

class Quote extends TwillBlockComponent
{
    public function render(): View
    {
        return view('components.twill.blocks.quote');
    }

    public function getForm(): Form
    {
        return Form::make([
            Input::make()->name('title'),
            Wysiwyg::make()->name('quote'),
            Medias::make()->name('content_image')->max(1)
        ]);
    }

    public static function getCrops(): array
    {
        return [
            'content_image' => [
                'default' => [
                    [
                        'name' => 'default',
                        'ratio' => 16 / 9,
                        'minValues' => [
                            'width' => 100,
                            'height' => 100,
                        ],
                    ],
                ]
            ]
        ];
    }
}
```

### DIFF
--- a/docs/content/1_documentation/5_block-editor/02_creating-a-block-editor.md
+++ b/docs/content/1_documentation/5_block-editor/02_creating-a-block-editor.md
@@ -100,6 +100,31 @@ public static function getBlockIcon(): string
 }
 ```
 
+#### Crops
+
+Usually we would define image crop's in the block_editor config, but with block components you can define
+them inline in your component like this:
+
+```php
+public static function getCrops(): array
+{
+    return [
+        'content_image' => [
+            'default' => [
+                [
+                    'name' => 'default',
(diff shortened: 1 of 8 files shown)


ifox:
Very cool @haringsrob! Could this be an opportunity to namespace crops keys in the background to avoid conflicts between model and block crops?

PR #2117, opened 2023-01-30
[3.x] Fix form builder check to allow not using the default content fieldset

joyceverheije:
This fixes an issue that was breaking the form when no fields were added to the form, only fieldsets.

### DIFF
--- a/src/Http/Controllers/Admin/ModuleController.php
+++ b/src/Http/Controllers/Admin/ModuleController.php
@@ -1144,7 +1144,7 @@ abstract class ModuleController extends Controller
 
         $controllerForm = $this->getForm($item);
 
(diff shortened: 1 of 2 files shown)


haringsrob:
I believe that if there's fieldsets it should still use that view for rendering.

Maybe we have to override the isNotEmpty method with our own conditions.

joyceverheije:
@haringsrob it should still use that view if there are fieldsets, this is what the [`hasForm` check](https://github.com/joyceverheije/twill/blob/1c7e63626aad3723f53a1054a7b4245c43e23046/src/Services/Forms/Form.php#L91-L94) is doing.

haringsrob:
Ah so that's our custom isNotEmpty check haha :D perfect in that case

PR #2135, opened 2023-02-07
[3.x] Fix conflicts between native Twill blocks and app/vendor blocks

ifox:
`twill.block_editor.directories` contains both Twill and app directories by default. Without this fix, Twill's default text block is taking over in the form instead of the app's text block.


haringsrob:
Weird behavior, when creating it showed the correct then on refresh the Twill's block.

Anyway tested and works perfectly!

PR #2140, opened 2023-02-09
[3.x] Fix return type

aksiome:
### A quick fix to silence IDE errors

I've noticed that setter methods used the self return type when returning $this. I believe it's best to return static now. This way users can extend classes without getting errors when chaining method calls.

I also removed the slugs mixin from the model interface since it is supposed to be an optional trait. Noticed it was added with the caching route fix but I'm not entirely sure why.


ifox:
Awesome @aksiome, we were actually discussing addressing this with @haringsrob a couple days ago. And @joyceverheije was looking into it to to fix PHPstan issues.

Thank you for working on this!

haringsrob:
Hey @aksiome this was indeed on my list too long! Thanks for fixing, the capsule phpcs issue is fixed in another pr so this is good to go!

PR #2157, opened 2023-02-17
Routing refactoring

### LINKED ISSUE #2017: [3.x]  Admin dashboard fails to load due to routes being cached
## Description

Twill admin dashboard fails to load due to routes being cached

### Steps to reproduce

1) create new laravel project
2) install twill 3.0.0-rc1 
3) install portfolio preset
4) execute `php artisan route:cache` to cache routes
5) try to login into the dashboard

### Expected result

The dashboard opens without any errors when Laravel routes are cached.
More likely we need to generate a cache file for TwillRoutes:$registry and load it when laravel routes are cached.

### Actual result

When trying to log in I get:
```Undefined array key "projects" (View: /var/www/html/vendor/area17/twill/views/layouts/main.blade.php) (View: /var/www/html/vendor/area17/twill/views/layouts/main.blade.php)```

This is due TwillRoutes::$registry property being empty when routes are cached.
Manually calling addToRouteRegistry for each route in RouteServiceProvider.php fixes the problem.

```
        TwillRoutes::addToRouteRegistry('projects', 'projects');
        TwillRoutes::addToRouteRegistry('partners', 'partners');
        TwillRoutes::addToRouteRegistry('comments', 'comments');
```

### Versions

Twill version: 3.0.0-rc1
Laravel version: v9.45.0
PHP version: 8.1.2 
Database engine: MySql

aksiome:
## Description
I wanted at first to fix the issue mentioned below and got a bit carried away and refactored a bit more code. My goal was to provide an experience similar to laravel resource routes. This pr allow to chain methods like `only` and `except` when defining a module or a singleton route.

## Motivation and future goals
I find the route registration method signature to be quite confusing for new users. There is a resource_options and a resource bool. My proposal is to deprecate these and only use one options array. This way twill routes are there own kind of self contained "resource like" routes. And with the way it is now written, adding new options should be quite simple.

I also wanted to be sure to not introduce any breaking change so I tried to cache multiple routes and checked that the output was the same as before. In the future I think it should be a good thing to reduce the number of unused routes especially in singletons. This refactoring also allow all these little changes to be really simple to make. Overall the code should be more future proof.

## Related Issues
- Fixes #2017 when only using resource routes and avoid undeclared class property errors in the future. If I remember correctly in php 8 we only get a deprecation notice for this.


aksiome:
I missed that one... I simplified the registerCapsuleRoutes method signature because I thought it was only used internally by the Capsule Helper class. Not sure now if I should change the test or the signature of the registerCapsuleRoutes method. In my opinion the test should be modified but would that be considered a breaking change? If it is then I'll just add back the router parameter to the registerRoutes and registerCapsuleRoutes methods.

haringsrob:
Hey @aksiome,

Thank you for the contribution. For now we are not going to progress on this pr. Maybe for 4.x we can have a fresh look at the routing but for now it is too close to 3.x to deal with such a change.

Kind regards,

aksiome:
Hey @haringsrob,

No problem at all, this pr wasn't really meant for 3.x anyway but more for future versions. I should have been more clear about it. The bug still exists with the cache but I think it's a minor issue (it appears when only resource routes are defined). What worries me is setting undefined properties on the PendingRouteRegistration object because it's deprecated and if I remember correctly this should throw an error on future php versions.

PR #2161, opened 2023-02-24
#2154: Use scope.

### LINKED ISSUE #2154: [3.x] Models\Feature::scopeForBucket() return type of Builder doesn't match returned collection
## Description

I am upgrading a project from `2.x` to `3.x`

When using feature buckets to organize content, and then calling `\A17\Twill\Models\Feature::forBucket('bucket-name')` on the frontend, it throws the following error:

> A17\Twill\Models\Feature::scopeForBucket(): Return value must be of type Illuminate\Database\Eloquent\Builder, Illuminate\Database\Eloquent\Collection returned

This appears to be because the `scopeForBucket()` method is now typed, and the code within probably does too much for a Laravel Scope method, since it does a `->get()` and then does some post processing on the query result.

### Steps to reproduce

Follow the steps from the documentation to create [Featured Content](https://twill.io/docs/featuring-content/) and then in the front end of your application, render the contents of a bucket using `$content = \A17\Twill\Models\Feature::forBucket('bucket-name')`

### Expected result

A Collection of Models that are stored in the Bucket

### Actual result

Throws 

> A17\Twill\Models\Feature::scopeForBucket(): Return value must be of type Illuminate\Database\Eloquent\Builder, Illuminate\Database\Eloquent\Collection returned

### Versions

Twill version:  `3.0.0-rc4`
Laravel version: `9.52.0`
PHP version: `8.2.2`
Database engine: `mariadb-10`

## Note

I'm not sure what would be actual end result should be. It's programmed as a `scope`, so should return a builder and allow it to be chained with more query arguments. So maybe it should just be converted to a simple method on the Eloquent Model so that you can return the processed Collection

haringsrob:
Fixes #2154


src/Models/Feature.php
```diff
@@ -23,9 +24,14 @@ public function featured(): MorphTo
...
     {
-        return $query->where('bucket_key', $bucketKey)->get()->map(function ($feature) {
+        return $query->where('bucket_key', $bucketKey)->with('featured');
+    }
+
+    public static function getForBucket(string $bucketKey): Collection
+    {
+        return self::forBucket($bucketKey)->get()->map(function (Feature $feature) {
             return $feature->featured;
-        })->filter();
```
ifox, line 28: We may need to keep this filter call to remove empty rows due to deleted records.

haringsrob: 👍 added it back

Issue #1530, opened 2022-03-19
Error retrieving credentials from the instance profile metadata service


joaocagnoni:
## Description

Error retrieving credentials from the instance profile metadata service. (cURL error 28: Connection timed out after 1001 milliseconds (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for http://169.254.169.254/latest/meta-data/iam/security-credentials/) (View: /var/www/html/vendor/area17/twill/views/layouts/dashboard.blade.php)

### Steps to reproduce

1. Install Laravel / Twill:

```
curl -s "https://laravel.build/example-app" | bash
cd example-app
./vendor/bin/sail up
sail composer require area17/twill:"^2.0"
sail artisan twill:install
```

2. Login as admin

### Expected result

Dashboard should open.

### Actual result

Error retrieving credentials from the instance profile metadata service. (cURL error 28: Connection timed out after 1001 milliseconds (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for http://169.254.169.254/latest/meta-data/iam/security-credentials/) (View: /var/www/html/vendor/area17/twill/views/layouts/dashboard.blade.php)

### Versions

**Twill:** 2.8.0-beta.2
**Laravel:** v9.5.1
**PHP:** 8.1.3
**Database:** MySQL 8.0.28-1.2.7-server

plexus77:
The issue above seems to be related to AWS and S3, even after swapping my media library to using the local storage the issue persists. 

Turns out regardless of your media library configuration you will still need to set up AWS and S3

haringsrob:
The media library does not need aws nor s3 to work locally.

@joaocagnoni make sure you have at least this in your .env if working with local storage:

```
FILE_LIBRARY_ENDPOINT_TYPE=local
MEDIA_LIBRARY_ENDPOINT_TYPE=local
MEDIA_LIBRARY_IMAGE_SERVICE=A17\Twill\Services\MediaLibrary\Glide
```

plexus77:
I tried that setup but was still receiving the error until I set Laravel’s aws configuration in the .env file.

plexus77:
Thands @haringsrob for the heads up with the FILE_LIBRARY_ENDPOINT_TYPE, that worked well... might be good to update the docs so that this is mentioned here https://twill.io/docs/getting-started/installation.html#setting-up-the-media-library

joaocagnoni:
[...]

PR #2162, opened 2023-02-24
Missing .env line for disabling S3 storage

DarthMikke:
## Description
Missing env variable FILE_LIBRARY_ENDPOINT_TYPE means Twill tries to utilize S3. This fix comes from a comment to issue #1530 .

## Related Issues
Addressed in #1530.


haringsrob:
hey @DarthMikke 

This is actually already documented on the configuration page, but it does not hurt to have it here as well.

https://github.com/area17/twill/blob/51c141a9424c8a96f03c84121ee6cc99726d5de4/docs/src/getting-started/configuration.md#local-endpoint-1

Thanks!

PR #2163, opened 2023-02-24
Add fromArray to options

aksiome:
This PR  adds a fromArray helper to the Options class and use it inside the options method.

```php
Select::make()->name('select')->options(Options::make([
	Option::make('foo', 'bar'),
	Option::make('val', 'label', true),
	Option::make('awesome', 'twill', true),
]))
```
All following syntaxes are supported.

```php
Select::make()->name('select')->options([
	'foo' => 'bar',
	[
		'value' => 'val',
		'label' => 'label',
		'selectable' => true,
	],
	Option::make('awesome', 'twill', true),
]))
```


haringsrob:
Hey @aksiome,

Great, if you can add a bit of docs this can go in.

PR #2165, opened 2023-02-25
Ignore missing or disabled modules/capsules for permissions

antonioribeiro:
When a package Capsule is disabled, Twill will raise a Model not found exception. This PR fixes this.


haringsrob:
Seems like there has been some overlap: https://github.com/area17/twill/blob/3.x/src/Helpers/modules_helpers.php can you check if this is still needed and if so have a look at the conflicts?

PR #2167, opened 2023-02-27
Set position on multi selects

antonioribeiro:
## Description

We have a model with a bunch of different tags:

Screenshot 2023-02-27 at 12 27 54

That are using multi-selects 

```php
$this->updateMultiSelect($object, $fields, 'countries');
```

One of our frontend pages needs to use the first tag of each of these fields. So I I tried to get the ordered:

```php
public function countries()
{
    return $this->belongsToMany(Country::class, 'analysis_country', 'analysis_id', 'country_id')
                ->orderBy('analysis_country.position');
}
```

Just find that the position field is not being filled up, they are all `null`, so this PR is intended to do it automatically, unless there's already a position being set by the application.


antonioribeiro:
CLA assistant is in trouble... :( 

Screenshot 2023-02-27 at 14 20 53

PR #2170, opened 2023-03-03
Laravel 10 support.

haringsrob:


tests/integration/PasswordsTest.php
```diff
@@ -77,7 +77,7 @@ public function testCanResetPassword()
             'token' => Notification::token(),
         ]);
 
-        $this->assertSee('Your password has been reset!');
+        $this->assertSee('You don&#039;t have any activity yet.');
```
haringsrob, line 80: @ifox not sure but this started to fail so I changed the assert to a different piece of text on the page.

PR #2174, opened 2023-03-08
Added dontAsk option to twill:make:module command

baccadiginepro:
## Description

This PR adds a feature to the CLI Generator command.
You can add options to twill:make:module but you will be prompted for each not specified option.
Adding --dontAsk the command takes only the specified options and skip the others (as if you had answered no).

This could be useful if you want to automate some module creation processes.

### DIFF
--- a/src/Commands/ModuleMake.php
+++ b/src/Commands/ModuleMake.php
@@ -30,6 +30,7 @@ class ModuleMake extends Command
         {--P|hasPosition}
         {--R|hasRevisions}
         {--N|hasNesting}
+        {--dontAsk}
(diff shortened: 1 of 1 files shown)


ifox:
Hi @baccadiginepro! Thank you for your contribution, it makes sense. I wonder if you have tried artisan's `--no-interaction` flag before implementing a custom option on Twill's command. Haven't tested yet but wanted to share that it might be kind of already supported natively.

baccadiginepro:
Same result can be reached with --no-interaction flag. Pull request not needed.

PR #2185, opened 2023-03-15
it translation: missing string and fixes

LucaRed:
## Description

Italian translation:

1. Added a missing string.
2. Some enhancements to the existing translations, in particular to prevent strange sentences with masculine/feminine model names.

Should I also backport these fixes to v2.0?


ifox:
Hi @LucaRed, thank you for this! Yes, please open against 2.x as well so that we can release these changes there too.

PR #2187, opened 2023-03-16
Docs home blog and guides styling updates

13twelve:
* home page updated (uses cards in grid)
* blog landing layout updated (uses cards in a grid - manually generated)
* blog h1 updated
* guides listing updated (uses cards in a grid - auto generated)
* now uses `text-`, `bg-` and `border-` colours from named usages in config, rather than colour tokens

It is possible to auto generate the blog listing but we can't do this and get close to the design with the description, tag and date as we don't have this data in the markdown file.

I wonder if this system can be updated to have [yaml front matter](https://jekyllrb.com/docs/front-matter/) which adds extra data to the items? @ifox @haringsrob


haringsrob:
Hey @13twelve,

Looks great. I have to be honest, I made this doc's system to be as simple as possible so I intentionally never looked at things like frontmatter.

If we just needs some additional metadata for blogs, we may be able to do that with some alternative syntax, or maybe a metadata json file alongside the blog list?

```json
{
  "what-is-new-in-twill-3": {
    "date": "26-01-1992",
    "tags": [
      "a",
      "b",
      "c"
    ],
    "description": "Lorem ispum dolor..."
  },
  ...
}
```

13twelve:
@haringsrob JSON metadata seems like a good plan. I've updated the `GenerateDocsCommand` to try and load a JSON file with the same filename as the markdown file and bundle and pass as `$metadata` with the `$treeData`. And then using this extra data to auto generate the blog listing with date and tag as per the design.

PR #2188, opened 2023-03-16
float min, max and step props for numeric input

iedex:
## Description

Allow using decimals for min, max and step properties for numeric input

### DIFF
--- a/src/View/Components/Fields/Input.php
+++ b/src/View/Components/Fields/Input.php
@@ -27,9 +27,9 @@ class Input extends TwillFormComponent
         public ?string $onChange = null,
         public ?string $onChangeAttribute = null,
         public ?string $prefix = null,
-        public ?int $min = null,
-        public ?int $max = null,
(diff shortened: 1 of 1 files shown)


src/View/Components/Fields/Input.php
```diff
@@ -27,9 +27,9 @@ public function __construct(
         public ?string $onChange = null,
         public ?string $onChangeAttribute = null,
         public ?string $prefix = null,
-        public ?int $min = null,
```
haringsrob, line 30: I would prefer to use `null|int|float` for this instead.

iedex: Sure, Can I change this in this PR?

haringsrob: sure!

iedex: done

ifox (approved):

PR #2195, opened 2023-03-18
[3.x] Prefix tables with twill_

aksiome:
Just noticed when migrating a test project with twill 3.0 that tables are still not prefixed with twill_.
I have 2 ideas to help update existing projects.
- I think the better would be to add a migration file that renames old tables if the config value is not overridden in the project.
- The other solution is to automatically add back the old default config to the project config.


haringsrob:
Hey @aksiome!

That is indeed something we still have to do. We do however, need a migration for this as well.

@ifox not sure what would be the correct way to go? We could also keep this for 3.1 to have more time to work on it.

A migration needed could be any of:
- Update the config to keep without prefix for existing 2.x sites.
- Update existing tables for existing 2.x sites.

There may also be cases where we currently use hard coded names instead of config. Those will need to be updated as well.

aksiome:
Hey @haringsrob,

I implemented the first solution. I thought a bit about it and updating the config and not the DB seems like the safest option. I also reworked a bit the other rector rules but it's basically the same code.
I removed the composer dump between rector runs but maybe it was needed. In my case it works fine like that!

rector-upgrade-twill-config.php
```diff
@@ -0,0 +1,28 @@
...
+use Rector\Config\RectorConfig;
+
+return static function (RectorConfig $config): void {
+    $config->paths([
+        getcwd() . '/config/twill.php',
+    ]);
+
+    $config->ruleWithConfiguration(LegacyTableConfig::class, [
+        'path' => getcwd(),
+        'tables' => [
```
haringsrob, line 15: I got a bit confused about this because it does it right, but there is no 'tables' prefix. Just to realise it is just the key used for the rule config.

haringsrob:
Hey, looks great! Can you take care of the phpstan outcome? If you have no time, let me know so I will take care of it.

PR #2198, opened 2023-03-24
[3.x] Avoid recreating related items.

### LINKED ISSUE #2179: Stop recreating related items on every save
## Summary

Blocks and browsers when connected to related items make use of this code:

```php
public function saveRelated($items, $browser_name)
{
    $this->clearRelated($browser_name);

    $position = 1;

    Collection::make($items)->map(function ($item) {
        return Arr::only($item, ['endpointType', 'id']);
    })->each(function ($values) use ($browser_name, &$position) {
        RelatedItem::create([
            'subject_id' => $this->getKey(),
            'subject_type' => $this->getMorphClass(),
            'related_id' => $values['id'],
            'related_type' => $values['endpointType'],
            'browser_name' => $browser_name,
            'position' => $position,
        ]);
        $position++;
    });
}
```

Not only this increases the number of dead rows on the related table, by deleting then recreating them every time, but, when nothing changed on a module, it also gives the false impression something has changed, as Laravel will fire `eloquent.created` events for these changes.

haringsrob:
Fixes #2179


composer.json
```diff
@@ -31,7 +31,7 @@
     "doctrine/dbal": "^3.0",
     "guzzlehttp/guzzle": "^7.0",
     "imgix/imgix-php": "^3.0",
-    "laravel/framework": "^9.0|^10.0",
+    "laravel/framework": "^9.0|^10.0 !=10.4.1 !=10.4.0",
```
ifox, line 34:
```suggestion
    "laravel/framework": "^9.0|^10.0",
```

ifox (changes requested):

PR #2204, opened 2023-03-30
[3.x] Set position on multi selects 

haringsrob:
Port of #2167


frontend/js/components/VSelect.vue
```diff
@@ -168,7 +168,16 @@
             if (this.taggable) {
               this.value = value
             } else {
-              this.value = this.options.filter(o => value.includes(o.value))
```
haringsrob, line 171: @ifox @antonioribeiro 

I had to add this chunk of js to ensure that the options are set to the order of how the id's come in. I believe this also needs to be in 2.x as otherwise it will take the order of all the available options (which is not sorted)

ifox: @antonioribeiro can you share more about your use case for storing the position on multiselect? How are you using the field: which options do you use and how are the values loaded?

ifox: ping @antonioribeiro

antonioribeiro: Hey @ifox, use case is basically when we need to read the selected items in the order we are selecting them.

On the screenshot below I selected "Education" first, because I want Education to be shown as the main sector in my blog post page.

Screenshot 2024-01-26 at 15 16 29

When I save and reload the model, Education is no more the first item and the frontend will show "Housing" as the main sector:

Screenshot 2024-01-26 at 15 14 26

It doesn't really matter how I'm getting these records later because their positions are all `null` on the relationship table:

Screenshot 2024-01-26 at 15 26 25

There will be now way for me to read them in the right order.

antonioribeiro: This is how I'm creating the form field to test on Twill 3, where I get the same behaviour:

```php
$form->add(
    MultiSelect::make()
        ->name("sectors")
        ->inline()
        ->options(
            Options::make(
                Sector::all()
                    ->map(
                        fn($sector) => Option::make(
                            $sector->id,
                            $sector->title
                        )
                    )
                    ->toArray()
            )
        )
);

```

PR #2215, opened 2023-04-12
[3.x] Fix duplicate action redirect route for nested parent-child mod…

agnonym:
…ules

For nested parent-child modules, the moduleRoute() call doesn't have the required parameters as the moduleName is "plural of parent module name + . + plural of child module name" (https://twill.io/docs/3.x/crud-modules/nested-modules.html#parent-child-modules).

So, if the moduleName contains '.':
- retrieve the real module name
- compute the parent module name and retrieve the value from the request query parameters

### DIFF
--- a/src/Http/Controllers/Admin/ModuleController.php
+++ b/src/Http/Controllers/Admin/ModuleController.php
@@ -1443,6 +1443,21 @@ abstract class ModuleController extends Controller
             $this->fireEvent();
             activity()->performedOn($item)->log('duplicated');
 
+            // Handle nested module.
+            if (Str::contains($this->moduleName, '.')) {
+                $moduleName = Str::afterLast($this->moduleName, '.');
+                $singularParentModuleName = Str::singular(Str::beforeLast($this->moduleName, '.'));
+
+                $parameters = [
+                    Str::singular($moduleName) => $newItem->id,
+                    $singularParentModuleName => request()->query($singularParentModuleName),
+                ];
+            } else {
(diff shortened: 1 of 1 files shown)


agnonym:
Sorry @ifox for the extra spaces...

ifox:
No worries, thanks for the fix!

Would you mind checking how much your solution overlaps with this [PR](https://github.com/area17/twill/pull/1846) on Twill 2?

agnonym:
From what I see, it doesn't seem to resolve the moduleName "naming problem" which is dot notation of parent and child module names (from the doc of Twill 2, it seems to be the same requirement to make this feature work: https://twill.io/docs/crud-modules/nested-modules.html#parent-child-modules).

The idea of allowing data injection in the newly created object is interisting (but maybe not necessary as you can override the duplicate() method in your repository, calling the parent that returns the new object and then do what you want with it). From what I understand from this PR, the approach is to manually share the parent id value, but in my case on Twill 3, I saw the value was already in the request query parameters with the singular parent module name as attribute name (ex: `.../clients/landings/duplicate/31?client=3`). So I retrieve it automatically without needing extra code to inject it.

If I may give my opinion, I'm not sure this PR fixes the problem of missing required parameters in the route... I don't have Twill 2 projects with nested parent-child modules, but I can test this PR in the next days if you want and/or see if I can port this PR in the Twill 2 branch.

Let me know, and "bon courage" for the 3.0.0 release

Issue #2214, opened 2023-04-12
Sidefieldsets doesn't get rendered if there is only a fieldset


EpicKau:
## Description

Sidefieldsets doesn't get rendered if there is only a fieldset

### Steps to reproduce

`    public function getSideFieldsets(TwillModelContract $model): Form
    {
        $form = parent::getSideFieldsets($model);

        //  test dummy
        //  $form->add(Input::make()->name('dummy')->label('dummy'));

        $form->addFieldset(
            Fieldset::make()->title('Opengraph')->id('opengraph')->fields([
                Input::make()->name('og_title')->label('OG Title'),
                Input::make()->name('og_description')->label('OG Description'),
                Files::make()->name('og_image')->label('OG Image')
            ])
        );

        return $form;
    }`

### Expected result

the fieldset should be rendert without a extra field

### Actual result

nothing get rendered 

### Fix for this
Services\Forms\Forms.php Line 97
` public function hasSideForm(): bool
    {
        return $this->isNotEmpty() || ($this->sideForm && $this->sideForm->isNotEmpty());
        // return $this->sideForm && $this->sideForm->isNotEmpty();
    }`

### Versions

Twill: 3.00 - RC4 (git)

ifox:
Thanks for reporting @EpicKau.

For reference, in the Blade forms, we have 2 sections (`sideFieldset` and `sideFieldsets`) to allow adding fields to a single fieldset or have the ability to add multiple fieldsets manually: https://github.com/area17/twill/blob/3.x/docs/content/1_documentation/4_form-fields/index.md#sidebar

EpicKau:
fixed - https://github.com/area17/twill/pull/2234

PR #2230, opened 2023-04-15
hasSideForm fix (issue 2214)

EpicKau:
## Description
hasSideForm is false if the form only has a fieldset and no direct fields. 

## Related Issues
Related to #2214


EpicKau:
the right fix is https://github.com/area17/twill/pull/2234

PR #2234, opened 2023-04-15
~ | Services - Form::hasSideForm: side form is not shown if it only contains fieldsets

agnonym:


ifox:
Thanks for all your contribution @agnonym!

Please check if this is PR is addressing the same issue, if not, it'd be great to address both in the same PR https://github.com/area17/twill/pull/2230.

agnonym:
Hmm... not sure #2230 fixes the problem as it checks if the main form is empty... meaning hasSideForm is true if you have fields in the main form. (I tried its code, if I have no field in the main form, my side form fieldset is not shown).

My guess is both PR address the same issue, but... maybe should you make up your own mind on how it should be addressed 🤔

agnonym:
> Thanks for all your contribution @agnonym!

It's just small participations! Thank you for all the work done on this new version

EpicKau:
@ifox  @agnonym  My bad, my code was using `$this->isNotEmpty()` but that referenced to the main form and not the side form.
It should be 
`$this->sideForm->fieldsets && $this->sideForm->fieldsets->isNotEmpty()`  
So [agnonym](https://github.com/agnonym) is right.

ifox (approved):

PR #2238, opened 2023-04-17
~ | Block::forComponent(): missing call to $componentClass::getBlockIcon()

agnonym:
Using TwillBlockComponent for a Block, the static method getBlockIcon() is never called, so its icon is always the default 'text'

### DIFF
--- a/src/Services/Blocks/Block.php
+++ b/src/Services/Blocks/Block.php
@@ -154,6 +154,7 @@ class Block
         );
 
         $class->title = $componentClass::getBlockTitle();
+        $class->icon = $componentClass::getBlockIcon();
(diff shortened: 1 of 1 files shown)


src/Services/Blocks/Block.php
```diff
@@ -154,8 +154,10 @@ public static function forComponent(string $componentClass): self
         );
 
         $class->title = $componentClass::getBlockTitle();
+        $class->icon = $componentClass::getBlockIcon();
         $class->titleField = $componentClass::getBlockTitleField();
         $class->hideTitlePrefix = $componentClass::shouldHidePrefix();
+
```
ifox, line 160:
```suggestion
```

ifox (approved):

PR #2252, opened 2023-04-29
Add title prefix support to component blocks.

haringsrob:
This pr allows component blocks to set the title prefix and title field:

```php
<?php

namespace App\View\Components\Twill\Blocks;

use A17\Twill\Services\Forms\Fields\BlockEditor;
use A17\Twill\Services\Forms\Fields\Wysiwyg;
use A17\Twill\Services\Forms\Form;
use A17\Twill\Services\Forms\Fields\Input;
use A17\Twill\View\Components\Blocks\TwillBlockComponent;
use Illuminate\Contracts\View\View;

class Demo extends TwillBlockComponent
{
    public static function getBlockTitleField(): ?string
    {
        return 'title';
    }

    public static function shouldHidePrefix(): bool {
        return true;
    }

    public function getForm(): Form
    {
        return Form::make([
            Input::make()->name('title'),
            Wysiwyg::make()->name('text'),
            BlockEditor::make()->name('left')
        ]);
    }

    public function render(): View
    {
        return view('components.twill.blocks.demo');
    }

}
```

### DIFF
--- a/src/Services/Blocks/Block.php
+++ b/src/Services/Blocks/Block.php
@@ -154,6 +154,8 @@ class Block
         );
 
         $class->title = $componentClass::getBlockTitle();
+        $class->titleField = $componentClass::getBlockTitleField();
+        $class->hideTitlePrefix = $componentClass::shouldHidePrefix();
         $class->rulesForTranslatedFields = (new $componentClass())->getTranslatableValidationRules();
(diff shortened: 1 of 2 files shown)


haringsrob:
@ifox I removed the block argument from the getTitle because that was the intended purpose but that could never work.

ifox (approved):

PR #2255, opened 2023-05-03
Fix 'Larvel' typo in docs

amuhenbergs:


peterhauke:
@ifox I see that the typo in the docs here on github has been corrected, but it appears that this has not propagated online: I can still see the typo: [in the Twill online documentation](https://twillcms.com/guides/page-builder-with-blade/creating-the-page-module.html)

PR #2256, opened 2023-05-03
Update 1_index.md

radumro:
Fix error in morph map


ifox:
Hi @radumro that is not how morphmaps should be defined, the value should be the FQCN (fully qualified class name), see the Laravel documentation: https://laravel.com/docs/10.x/eloquent-relationships#custom-polymorphic-types

PR #2270, opened 2023-05-21
Corrected option definition time24h

kerkness:
## Description

The option `time24Hr` is incorrectly documented.  The option is actually defined as `time24h` in the DatePicker class.

While the DatePicker class does use the property/variable $time24Hr, I am assuming the option method was changed to `time24h` in v3 for a reason and the documentation is incorrect.

Alternatively this PR could be rejected and the method could be updated to match the documentation.


ifox:
Hi @kerkness, the confusion comes from the fact that the OOP form builder uses a different name than the blade component, which exposes an issue with our documentation approach since we're not documenting the options for the form builder and blade components separately.

PR #2277, opened 2023-05-30
Handling groups of fields fix

yamaha252:
If I use a dot as `fieldsGroupsFormFieldNameSeparator` some values in a group is not saves because laravel's method `Arr::forget` has a deep removing functionality which is not necessary here.


ifox:
Thanks @yamaha252 that makes sense. Can you please sign the CLA?

ifox (approved):

PR #2285, opened 2023-06-08
* | Dashboard: add Fathom Analytics support

agnonym:
## Description

Twill dashboard allows only Google Analytics as analytics solution.

This PR add Fathom Analytics support (https://usefathom.com)

To enable Fathom Analytics service, the following configuration is needed:

```
    'dashboard' => [
        'analytics' => [
            'enabled' => true,
            'service' => 'fathom',
            'fathom' => [
                'api_token' => '11222233333',
                'site_id' => 'FATHOMSITEID',
            ],
        ],
    ],
```

Procedure to get an API token: https://usefathom.com/api

### DIFF
--- a/frontend/js/components/dashboard/statFeed.vue
+++ b/frontend/js/components/dashboard/statFeed.vue
@@ -38,9 +38,6 @@
       </a>
       </template>
     </div>
-    <footer class="box__footer statFeed__footer">
-      <a href="https://analytics.google.com/analytics/web" class="f--external" target="_blank">Google Analytics</a>
-    </footer>
   </div>
 </template>
 
--- a/src/Http/Controllers/Admin/DashboardController.php
+++ b/src/Http/Controllers/Admin/DashboardController.php
@@ -13,6 +13,7 @@ use Illuminate\Http\JsonResponse;
 use Illuminate\Http\Request;
 use Illuminate\Pagination\LengthAwarePaginator;
 use Illuminate\Support\Collection;
+use Illuminate\Support\Facades\Http;
 use Illuminate\Support\Str;
 use Illuminate\View\Factory as ViewFactory;
 use Psr\Log\LoggerInterface as Logger;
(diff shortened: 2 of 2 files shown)


agnonym:
Hi @antonioribeiro, I totally agree with you! If I may make a suggestion, maybe Twill should allow to plug widgets on the dashboard, widgets provided by extra packages (Google Analytics, Fathom Analytics, ...) or custom code in your project.

agnonym:
Hi @antonioribeiro,
I haven't had the chance to work with Twill Capsules yet. However, upon reviewing the code, using Capsules could be a viable option for implementing a "Dashboard Widget" feature. One potential approach could involve introducing an additional attribute in the twill.capsules.list configuration, such as `'dashboard' => true`. And maybe creating a dedicated class akin to Twill Block Component classes, implementing a method in the Capsule ServiceProvider, ...

Enhancing the Dashboard Controller seems relatively straightforward, invoking the declared Capsules and retrieve the data, potentially even rendering the view directly for simplicity.
As for the Dashboard View, maintaining it as Blade code, with each section employing A17 Vue components, could be a practical solution. The decision to use pure Vue within the Blade could be left to the discretion of each Capsule. In an initial iteration, arranging the widgets in the right column below or above the drafts, ordered based on the declaration sequence of the Capsules, could be a simple yet effective layout. Of course, there's room to explore additional features, such as switching between the right and left columns, configuring the width percentage for each widget, and so on.

I'm unsure whether you or the Twill team are interested in delving deeper into this topic. If you value my opinion, I'm usually available on Discord for further discussion.

antonioribeiro:
@agnonym, speak for myself here, but I believe the idea from the beginning was to have a plugin system on Twill, when we created Capsules it became the natural way to try to make this happen, as Capsules were also decoupled from Twill. You can already do a lot with them, but if what we need to have even more "low level" plugins on Twill depends only on some minor changes to Twill, like adding new properties to the capsules list, I believe the Core team will probably not have a problem with introducing them, and I can surely help too

PR #2292, opened 2023-06-15
Add DI support for twill Block getData method

Afting:
Hello, I have been trying to use Twill Block, and extend it like it's documented in https://twillcms.com/docs/block-editor/rendering-blocks.html#content-modifying-block-data, though no method supports DI. This change would allow to use DI in getData() method.
Regarding the changes Laravel itself likes to do things like that, ex. https://github.com/laravel/framework/blob/10.x/src/Illuminate/Mail/Mailable.php#L1612-L1614

Also had a discussion on Discord to create this PR

### DIFF
--- a/docs/content/1_docs/3_modules/12_nested-modules.md
+++ b/docs/content/1_docs/3_modules/12_nested-modules.md
@@ -151,7 +151,7 @@ use A17\Twill\Services\Breadcrumbs\NestedBreadcrumbs;
 
 class IssueArticleController extends BaseModuleController
 {
-    protected $modelName = 'BrandProduct';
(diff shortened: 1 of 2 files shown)


ifox:
Hi @Afting I was about to merge this but I'm concerned about the example provided in the docs needing a correction if we remove the `getData` method implementation in the Block parent class:

```php
public function getData(array $data, \A17\Twill\Models\Block $block): array
{
    $data = parent::getData($data, $block);
    // ...
}
```

`parent::getData()` wouldn't work anymore, and it is almost certain that developers using this feature are currently using it by pasting the example from the docs. Can we keep it while benefiting from DI in the method override?

PR #2301, opened 2023-06-25
Use pushOnce builtin, remove custom pushonce directive

### LINKED ISSUE #2300: Blocks in block editor have vanished, but they do exist
## Description

I've seen this a few times now, and not 100% sure how to reproduce, but from time to time, the block editor in a module fails to render for all items, only showing an "Add content" button.

**Edit:** Once it has disappeared from one editor, it affects all modules.

The last time this happened, all I did was expand all blocks in the editor, and then collapse them again – I did not save anything.

I checked the JS code being hydrated, and the `form.blocks` assignment is not present at all – ie, this one:

https://github.com/area17/twill/blob/6457e56318d87819b5230511dd35efd9ad59188f/views/partials/form/utils/_block_editor_store.blade.php#L1

However, when I view the item on the site, the blocks render perfectly fine, which means they are accessible. But, if I were to go back to the editor and save the item, the blocks will disappear – so now, I'm unable to save without destroying the entire item.

I'll do some more digging to see what I can find – my hope is that something on my end is breaking the flow of things on the backend (even though I doubt it – it's not a super custom setup).

### Versions

Twill version: 3.0.2
Laravel version: 10.13.5
PHP version: 8.1
Database engine: Postgres 15

mikerockett:
## Description

This PR removes the custom `pushonce` Blade directive, and uses the `pushOnce` directive baked into Laravel 9 and 10 instead.

In order to reduce the chances of a sub-stack `$id` collision, I have prepended the two usages with `form:`. Perhaps it might be a good idea for me to open a PR against Laravel, which does not check if an `$id` is used in a specific stack, but rather across all stacks, which I don't think makes a lot of sense.

## Related Issues

Resolves #2300

### DIFF
--- a/src/TwillServiceProvider.php
+++ b/src/TwillServiceProvider.php
@@ -507,17 +507,6 @@ class TwillServiceProvider extends ServiceProvider
             ?>";
         });
 
-        $blade->directive('pushonce', function ($expression) {
-            [$pushName, $pushSub] = explode(':', trim(substr($expression, 1, -1)));
-            $key = '__pushonce_' . $pushName . '_' . str_replace('-', '_', $pushSub);
-
-            return "<?php if(! isset(\$__env->{$key})): \$__env->{$key} = 1; \$__env->startPush('{$pushName}'); ?>";
-        });
-
-        $blade->directive('endpushonce', function () {
-            return '<?php $__env->stopPush(); endif; ?>';
-        });
-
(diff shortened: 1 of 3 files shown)


mikerockett:
I’ve also [opened a PR](https://github.com/laravel/framework/pull/47565) for Laravel, though it can’t have any bearing on this PR unless it is also merged into 9.x.

PR #2311, opened 2023-07-30
fix:bug preserve default vselect value when undefined

bwat-dev:
## Description
Filters select :
When clearing filters  the select loses its default value, and the field displays an empty select :

Capture d’écran 2023-07-30 à 12 20 07


ifox (changes requested):
Thanks @bwat-dev! Submitted a few cleanup changes but otherwise should be good to go.

PR #2314, opened 2023-08-02
Duplicate nested columns keys fix

yamaha252:
If there are several `NestedData` table columns, the value is duplicated because the keys are always `children` and it's impossible to overwrite them

### DIFF
--- a/docs/content/1_docs/3_modules/6_table-builder.md
+++ b/docs/content/1_docs/3_modules/6_table-builder.md
@@ -172,9 +172,9 @@ unpublished.
 
 ##### NestedData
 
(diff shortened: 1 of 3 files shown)


ifox:
Thanks @yamaha252 ! @joyceverheije ran into this issue recently and had the intention to PR as well, but let's go with yours since you opened it first. If you can bring in her [docs updates](https://github.com/joyceverheije/twill/commit/1edbfbb16feefe866d8f1a6d652c4f97edd23360) though, that'd be great.

yamaha252:
@ifox done, thanks!

ifox (approved):

PR #2336, opened 2023-09-15
Add predefined class selection to TipTap link component

zipavlin:
## Description

This minor feature adds option for editors to add one or more predefined classes to anchor elements using TipTap editor's link modal. This allows editors to create CTAs or other button-like hyperlinks inside wysiwyg editor, removing the need for developers to add separate fields.

image

Editor can add classes using checkboxes, rather than input field. There are two reasons for this decision: firstly, user/editor can already see source and add specific classes there. Secondly, we want to limit what classes are actually added.

Therefore this feature expects class sets to be defined when creating a Wysiwyg field using new method 'classList' which take associative array of `['class' => 'Label']` for example:

```php
 Wysiwyg::make()
   ->classList([
     'btn' => 'Show as button',
     'btn-big' => 'Show as big button'
   ])
```


ifox:
Thanks @zipavlin, this is a cool feature. I improved the styling a bit:

Untitled

PR #2344, opened 2023-10-03
[3.x] Fix - create icons target folder if it doesn't exist

adam-riha:
## Description

Quick fix for the command `twill:build` if the app contains custom icons.


ifox:
Thanks for your contribution! This has also been fixed and merged in #2392.

PR #2362, opened 2023-10-31
pass model name for getModelController helper function

jwkuo:
## Description

Tweak for getting the Model Controller when belonging to a Capsule. TwillCapsules::getCapsuleForModel() expects the model name as a string.

### DIFF
--- a/src/Helpers/modules_helpers.php
+++ b/src/Helpers/modules_helpers.php
@@ -92,7 +92,7 @@ if (!function_exists('getModelController')) {
 
         if (!class_exists($controller)) {
             try {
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "area17/twill",
+  "name": "jwkuo/twill",
```
antonioribeiro, line 2: Why are you changing the name of the package?

jwkuo: My bad, I was just trying to use this package to test and committed it to the 3.x branch. I've removed that.

Issue #2153, opened 2023-02-14
[3.x] Add Dusk to require-dev


antonioribeiro:
Because Twill 3 is refering to `Laravel\Dusk\Browser` I believe Dusk should be added to composer.json. Otherwise any packages doing static analysis on a Twill based code will have to add Dusk to their own composer.json.

```bash
Error thrown in /Users/antonioribeiro/code/area17/twill-google-recaptcha/vendor/area17/twill/src/DuskServiceProvider.php on line 17 while loading bootstrap file /Users/antonioribeiro/code/area17/twill-google-recaptcha/vendor/nunomaduro/larastan/bootstrap.php: Class "Laravel\Dusk\Browser" not found
```

haringsrob:
Hey @antonioribeiro,

Might be wrong but require-dev is never installed as part of the install process of packages.

Are you running composer install in twill vendor's folder?

antonioribeiro:
Hey @haringsrob, if you are developing a package and you need to do a `composer install` inside the package folder (vendor/area17/twill-firewall for instance) to have all dependencies plus PHPUnit and PHPStan available for running package tests, Twill will be installed, because it's a dependency of the package, so when I run 

```bash
vendor/bin/phpstan analyse -v
```

I get 

```
Error thrown in 
/Users/antonioribeiro/code/client/project/vendor/area17/twill-firewall/vendor/area17/twill/src/DuskServiceProvider.php 
on line 17 while loading bootstrap file 
/Users/antonioribeiro/code/client/project/vendor/area17/twill-firewall/vendor/nunomaduro/larastan/bootstrap.php: 
Class "Laravel\Dusk\Browser" not found
```

Another solution would be to add Dusk to the package, but Dusk is not really a dependency of the package, and it's only a dependency for Twill 3.x, on Twill 2.x it works fine.

pboivin:
Hi @antonioribeiro, I ran into the same issue recently. I added Dusk as a dev dependency of my package as you said. I agree that the package doesn't really care if Twill is using Dusk in its own test suite. Could this be handled on Twill's side, by only registering `DuskServiceProvider` if class `Laravel\Dusk\Browser` exists?

PR #2366, opened 2023-11-02
Don't load DuskServiceProvider if dusk doesn't exist

Tofandel:
## Description

When running tests without dusk installed this triggers an error, it also causes errors with static analysis

## Related Issues

Fixes https://github.com/area17/twill/issues/2153


src/DuskServiceProvider.php
```diff
@@ -14,286 +14,288 @@ class DuskServiceProvider extends ServiceProvider
...
-        });
-
-        Browser::macro('assertVselectHasOptions', function (string $wrapperClass, array $optionLabels) {
-            $this->with($wrapperClass, function (Browser $element) use ($optionLabels) {
-                $element->click('.vs__search');
-                $element->waitFor('.vs__dropdown-menu');
-
-                foreach ($optionLabels as $optionLabel) {
-                    $element->assertSeeIn('.vs__dropdown-menu', $optionLabel);
+        if (class_exists(Laravel\Dusk\Browser::class)) {
```
ifox, line 17: Please return early by checking the inexistence instead, to avoid wrapping all that code under  a conditional. This will also make this PR diff a lot easier to review. Thanks for your contribution!

ifox (changes requested):

src/DuskServiceProvider.php
Tofandel: As someone once showed me, you can use the `hide whitespace` feature to see a clean diff
image

ifox: For sure! I use this too and it's great. In this case I was more concerned by the unnecessary indentation of all that code. Thanks for addressing it.

ifox (dismissed):

src/DuskServiceProvider.php
```diff
@@ -14,6 +14,9 @@ class DuskServiceProvider extends ServiceProvider
 {
     public function boot(): void
     {
+        if (!class_exists(Laravel\Dusk\Browser::class)) {
```
Tofandel, line 17:
```suggestion
        if (!class_exists(\Laravel\Dusk\Browser::class)) {
```

src/DuskServiceProvider.php
```diff
@@ -14,6 +14,9 @@ class DuskServiceProvider extends ServiceProvider
 {
     public function boot(): void
     {
+        if (!class_exists(\Laravel\Dusk\Browser::class)) {
```
Tofandel, line 17:
```suggestion
        if (!class_exists(Browser::class)) {
```

Issue #2387, opened 2024-01-03
$item->ancestorsSlug returns items in reverse order? Issue with HasNesting trait


josh-ivyjack:
## Description

I have pages module with self-nesting. Nesting value set greater than 2. When calling: $item->ancestorsSlug. It returns the slug in the incorrect order from the child parent relationship.

Looking at the HasNesting trait: vendor/area17/twill/src/Models/Behaviors/HasNesting.php

It is reserving the array before mapping?

Also, documentation seems incorrect at: https://twillcms.com/docs/modules/nested-modules.html

You will get:

Declaration of App\Http\Controllers\Twill\PageController::form($id, $item = null) must be compatible with A17\Twill\Http\Controllers\Admin\ModuleController::form(?int $id, ?A17\Twill\Models\Contracts\TwillModelContract $item = null): array

When copying from docs, had to update to: 
protected function form(?int $id, TwillModelContract $item = null): array {}

### Steps to reproduce

Added this to my page controller:

protected function form(?int $id, TwillModelContract $item = null): array
    {

        $item = $this->repository->getById($id, $this->formWith, $this->formWithCount);
        //incorrect order..
        $this->permalinkBase = $item->ancestorsSlug;
        
        //quick fix...
        $this->permalinkBase = implode('/', array_reverse(explode('/', $item->ancestorsSlug)));

        return parent::form($id, $item);

    }

### Expected result

The slug should be returned in the correct order as nested.

### Actual result

The slug order is reversed.

### Versions

3.0.2

10.10

8.2

MiaraDB

Issue #2302, opened 2023-07-03
AncestorsSlug 3rd level and more does not show correctly


mefutu01:
## Description

Incorrect behavior of nested pages from the 3rd and further levels. Nesting is reversed in the file src/Models/Behaviors/HasNesting.php:38 and the final uri is formed incorrectly

```php 
    /**
     * Returns the combined slug for all ancestors of this item.
     */
    public function getAncestorsSlug(?string $locale = null): string
    {
        return $this->getAncestors()->reverse()
            ->map(function ($i) use ($locale) {
                return $i->getSlug($locale);
            })
            ->implode('/');
    }
```

```php 
  ->reverse()
```

### Steps to reproduce

Update PageCotroller 

```php 
    protected function form(?int $id, ?TwillModelContract $item = null): array
    { 
        $item = $this->repository->getById($id, $this->formWith, $this->formWithCount);

        $this->permalinkBase = $item->ancestorsSlug;

        return parent::form($id, $item);
    }
```

1. Creating nested pages;
  2. Create page /1st/;
  3. Create page /1st/2nd/;
  4. Create page  /1st/2nd/3rd;
  5. Create page  /1st/2nd/3rd/4th;

### Expected result
I expect that after creating the 3rd level, the ancestorsSlug will look like: 
_/1st/2nd/3rd_
_/1st/2nd/3rd/4th_

### Actual result

_/2nd/1st/3rd_
_/3rd/2nd/1st/4th_

### Versions

Twill version: 3.0.2  
Laravel version: v9.52.9 
PHP version: PHP v8.1.12-1ubuntu4
Database engine: PostgreSQL: 15.2

ifox:
Thanks for reporting @mefutu01. We appear to be missing a call to `defaultOrder`: https://github.com/lazychaser/laravel-nestedset#default-order to fix this.

mefutu01:
Thanks for the quick response @ifox )  You make a great CMS :)

Issue #2280, opened 2023-06-02
Self-nested modules don't have the correct ancestor slug when nestedItemsDepth is 2+


Tom-revere:
## Description

When setting up self nested modules, if the `$nestedItemsDepth` is set to 2+, it will give the wrong ancestor slug. 

### Steps to reproduce

1. Setup self-nested modules following [this ](https://twillcms.com/docs/modules/nested-modules.html#content-creating-self-nested-modules) guide.
2. Set `protected $nestedItemsDepth = 2;` inside PageController.php
3. Create pages in the following structure
![image](https://github.com/area17/twill/assets/70330106/85c516a2-444e-4855-90b2-1c94c4633e48)

### Expected result

Given the following structure, you should expect the slugs to be:
1. /parent-page
2. /parent-page/child-page
3. /parent-page/child-page/child-of-child-page

### Actual result

You actually get:

1. /parent-page
2. /parent-page/child-page
3. /child-page/parent-page/child-of-child-page

I believe this is due to this line.

https://github.com/area17/twill/blob/540fc45871f64df7f80823c20f951acf1b796443/src/Models/Behaviors/HasNesting.php#L38

Removing `->reverse()` fixes the issue.

### Versions

Twill: 3.0.2

Laravel: 9.52.4

PHP: 8.0

PR #2388, opened 2024-01-03
Fixes incorrect order of nested items when item is more than 2 levels deep

pauldwight:
## Description
The getAncestors() from the nestedset package doesn't set any default order which means that the ancestors returned are in order of creation (id) rather than hierarchical order.

Rather than making the assumption as to what order the ancestor items were created it is better to sort the collection by the _lft column (as the defaultOrder() method does in the package).

See https://github.com/lazychaser/laravel-nestedset#ancestors-and-descendants for more info.

## Related Issues
Fixes https://github.com/area17/twill/issues/2387 https://github.com/area17/twill/issues/2302 https://github.com/area17/twill/issues/2280

### DIFF
--- a/src/Models/Behaviors/HasNesting.php
+++ b/src/Models/Behaviors/HasNesting.php
@@ -35,7 +35,7 @@ trait HasNesting
      */
     public function getAncestorsSlug(?string $locale = null): string
     {
-        return $this->getAncestors()->sortBy('_lft')
+        return $this->getAncestors()->sortBy($this->getLftName())
             ->map(function ($i) use ($locale) {
                 return $i->getSlug($locale);
             })
--- a/src/Models/Behaviors/HasNesting.php
+++ b/src/Models/Behaviors/HasNesting.php
@@ -35,7 +35,7 @@ trait HasNesting
      */
     public function getAncestorsSlug(?string $locale = null): string
     {
-        return $this->getAncestors()->reverse()
+        return $this->getAncestors()->sortBy('_lft')
             ->map(function ($i) use ($locale) {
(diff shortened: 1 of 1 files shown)


src/Models/Behaviors/HasNesting.php
```diff
@@ -35,7 +35,7 @@ public function getNestedSlugAttribute()
      */
     public function getAncestorsSlug(?string $locale = null): string
     {
-        return $this->getAncestors()->reverse()
+        return $this->getAncestors()->sortBy('_lft')
```
ifox, line 38:
```suggestion
        return $this->getAncestors()->sortBy($this->getLftName())
```

Considering that our generated migration relies on [the package helper](https://github.com/area17/twill/blob/3a01156078c1fed7c10523b8c8c0cd765b18cd9a/src/Commands/stubs/migration.stub#L28C21-L28C30), I think we should also rely on [the package method to retrieve the name of that column](https://github.com/lazychaser/laravel-nestedset/blob/f7a36eb3779ee4119a8ea72404aaf90c44fc9a15/src/NodeTrait.php#L845-L848).

ifox (changes requested):
Thanks @pauldwight, really appreciate you jumping on this! 

This looks like the appropriate solution, just made a minor suggestion.

I looked at the history of this method and it looks like we got it wrong a couple of times and that the confusion came from the fact that it used to depend on the order of record creation.

src/Models/Behaviors/HasNesting.php
pauldwight: Updated

Issue #2342, opened 2023-09-24
Multiple JSON repeaters (of the same type) in one form


S-ami:
## Description

When there are multiple JSON repeaters (same type, different names) in one form, the data are saved correctly but they won't get loaded to the repeaters.

```BLADE
<x-twill::repeater
        type="row"
        name="table1"
    />

<x-twill::repeater
        type="row"
        name="table2"
    />

<x-twill::repeater
        type="row"
        name="table3"
    />
```

I made a workaround in module repository class by overwriting `getJsonRepeater` function for this case

### Versions

Twill version: 3.0.2

PR #2400, opened 2024-01-10
Fix getCapsuleForModel() for when passing a Model object

antonioribeiro:
## Description

When trying to to `$model->fullUrl()` I found that I was always getting a `#` instead of the article URL, even having a HasSlug trait added to the model. I ended up finding that because this was a Twill Capsule it was not resolving the Capsule controller correctly, all because an Eloquent Model was being passed to `getCapsuleForModel()` instead of the model name and as it was also hidden behind this exception handler:

```php
try {
    $controller = getModelController($this);
} catch (Exception) {
    // Fallback to never crash on production.
    return '#';
}
```

This PR fixes this and correctly gives us an URL. It may also fix other Capsules resolution errors that could fall in the same "pass a model instead of a model name" category.

### DIFF
--- a/src/TwillCapsules.php
+++ b/src/TwillCapsules.php
@@ -6,6 +6,7 @@ use A17\Twill\Exceptions\NoCapsuleFoundException;
 use A17\Twill\Helpers\Capsule;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Str;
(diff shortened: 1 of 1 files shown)


ifox:
Thanks @antonioribeiro! I believe this supersedes #2342 in a more impactful way (it solves more cases of resolution issues).

ifox (approved):

PR #2408, opened 2024-01-23
Use model configuration in the relationship morph map

redelschaap:
There is a bug in the package in case you change the model configuration and you already have relational data in the database.

In my case, I had an app setting that saves a relation to a page. After changing the block class through the model configuration, I got some strange errors. This adjustment in the package resolved the errors.


ifox (approved):

redelschaap:
@ifox Any idea when the next 3.x version will be released?

PR #2413, opened 2024-01-29
bug fix where 'php artisan migrate:reset' would error.

NuktukDev:
## Description

When executing `php artisan migrate:reset` an error would happen. The error would say:
> SQLSTATE[42S02]: Base table or view not found: 1146 Table 'twill-db.related' doesn't exist (SQL: alter table `related` drop `id`)

This is fixed in this PR.

### DIFF
--- a/migrations/default/2023_03_24_125122_add_id_to_related.php
+++ b/migrations/default/2023_03_24_125122_add_id_to_related.php
@@ -27,8 +27,10 @@ return new class extends Migration {
 
     public function down(): void
     {
-        Schema::table(config('twill.related_table', 'twill_related'), function (Blueprint $table) {
-            $table->dropColumn('id');
-        });
+      if (Schema::hasColumn(config('twill.related_table', 'twill_related'), 'id')) {
+            Schema::table(config('twill.related_table', 'twill_related'), function (Blueprint $table) {
+                $table->dropColumn('id');
+            });
+        }
     }
 };
--- a/migrations/default/2023_03_24_125122_add_id_to_related.php
+++ b/migrations/default/2023_03_24_125122_add_id_to_related.php
@@ -27,7 +27,7 @@ return new class extends Migration {
 
     public function down(): void
     {
-      if (Schema::hasColumn(config('twill.related_table', 'twill_related'), 'id')) {
+        if (Schema::hasColumn(config('twill.related_table', 'twill_related'), 'id')) {
             Schema::table(config('twill.related_table', 'twill_related'), function (Blueprint $table) {
                 $table->dropColumn('id');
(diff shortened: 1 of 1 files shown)


NuktukDev:
Signed

migrations/default/2023_03_24_125122_add_id_to_related.php
```diff
@@ -27,8 +27,10 @@ public function up(): void
 
     public function down(): void
     {
-        Schema::table(config('twill.related_table', 'twill_related'), function (Blueprint $table) {
-            $table->dropColumn('id');
-        });
+        if (! Schema::hasColumn(config('twill.related_table', 'twill_related'), 'id')) {
```
Tofandel, line 30: It's the other way around

```suggestion
        if (Schema::hasColumn(config('twill.related_table', 'twill_related'), 'id')) {
```

Tofandel:
I also ran into this issue and was about to submit the exact same PR

migrations/default/2023_03_24_125122_add_id_to_related.php
Tofandel, line 30: It's the other way around

```suggestion
        if (Schema::hasColumn(config('twill.related_table', 'twill_related'), 'id')) {
```

Tofandel (changes requested):

NuktukDev:
You're correct, thanks.

NuktukDev (commented):
Correct.

migrations/default/2023_03_24_125122_add_id_to_related.php
Tofandel: I didn't see you commit the change

ifox: @NuktukDev as @Tofandel suggested the current change your propose is incorrect. Let us know if you can fix that, otherwise we'll push the change in a separate PR. Thanks for your contribution!

migrations/default/2023_03_24_125122_add_id_to_related.php
```diff
@@ -27,8 +27,10 @@ public function up(): void
 
     public function down(): void
     {
-        Schema::table(config('twill.related_table', 'twill_related'), function (Blueprint $table) {
-            $table->dropColumn('id');
-        });
+      if (Schema::hasColumn(config('twill.related_table', 'twill_related'), 'id')) {
```
ifox, line 30:
```suggestion
        if (Schema::hasColumn(config('twill.related_table', 'twill_related'), 'id')) {
```

ifox (approved):

Issue #2424, opened 2024-02-07
FormBuilder Browser field with single module


iedex:
Can this be true?
https://github.com/area17/twill/blob/063b4bbbe2b133f48008de250be8f767ce2350c1/src/Services/Forms/Fields/Browser.php#L148

What will be the best change?

ifox:
Yes it can, in the case where $modules is an array that contains exactly one element, but that element is not at the index 0. This can happen if the array uses non-sequential or non-numeric keys.

Non-sequential keys: If the array's keys are not sequential, and the starting key is not 0, then count($modules) could return 1 (indicating that there is one element in the array), but $modules[0] would not be set because the key 0 does not exist in the array.

Associative array with a single element: If $modules is an associative array (where keys are strings instead of integers), and it contains just one element with a key other than 0, count($modules) would be 1, but isset($modules[0]) would evaluate to false since there's no element with the key 0.

This condition is to account for someone passing something like this: `['posts' => Post::class]`.

iedex:
Ah, got it. I was looking for a method to use only one module, because moduleName doesn't work in FormBuilder, and when using modules(['module_name']) the browser has that selector for the single module.
Tried now and it works as expected with ->modules(['name' => 'module_name']).
Thanks a lot!

PR #2474, opened 2024-02-13
Fix single module browser

Tofandel:
[The documentation](https://twillcms.com/docs/form-fields/browser.html) refers to a `moduleName` option for a single module browser, which doesn't seem to exist on the form builder, the code suggests that it should be possible using `modules([Module::class])` and so does the doc

```php
Browser::make()
    ->modules([Publications::class])
    ->name('publications')
    ->max(4);
```

And so the previous form field should be a browser for a single module

However that's not the case as can be seen from the dropdown

image

This is due to ` count($modules) === 1 && ! isset($modules[0])` which doesn't make much logical sense

With the PR this is what becomes:
image

### DIFF
--- a/docs/content/1_docs/4_form-fields/10_browser.md
+++ b/docs/content/1_docs/4_form-fields/10_browser.md
@@ -47,8 +47,7 @@ Browser::make()
 |:----------------------|:--------------------------------------------------------------------------------|:--------|:--------------|
 | name                  | Name of the field                                                               | string  |               |
 | label                 | Label of the field                                                              | string  |               |
-| moduleName            | Name of the module (single related module)                                      | string  |               |
-| modules               | Array of modules (multiple related modules), must include _name_                | array   |               |
+| modules               | Array of modules (multiple related modules), must include _name_ or be a module class | array   |               |
 | endpoints             | Array of endpoints (multiple related modules), must include _value_ and _label_ | array   |               |
 | params                | Array of query params (key => value) to be passed to the browser endpoint       | array   |               |
 | max                   | Max number of attached items                                                    | integer | 1             |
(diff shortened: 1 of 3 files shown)


docs/content/1_docs/4_form-fields/10_browser.md
```diff
@@ -47,8 +47,7 @@ Browser::make()
 |:----------------------|:--------------------------------------------------------------------------------|:--------|:--------------|
 | name                  | Name of the field                                                               | string  |               |
 | label                 | Label of the field                                                              | string  |               |
-| moduleName            | Name of the module (single related module)                                      | string  |               |
-| modules               | Array of modules (multiple related modules), must include _name_                | array   |               |
+| modules               | Array of modules (multiple related modules), must include _name_ or be a module class | array   |               |
```
Tofandel, line 51:
```suggestion
| moduleName | Name of the module (single related module) (use the `modules` method with a single class for the FormBuilder) | string | |
| modules               | Array of modules (multiple related modules), must include _name_ or be a module class | array   |               |
```

ifox:
@Tofandel related discussion here: https://github.com/area17/twill/issues/2424#issuecomment-1931642021

Tofandel:
I figured it could be true and I in fact worked around it using the key 1, but it's not clean and just a workaround and not what the doc says to use, so I fixed the condition so that now any array with a single string in it, no matter its key will work correctly

It handles `[Post::class]` or `['post' => Post::class]`as a single module and `[['name' => 'Post']]` as multiple modules which is the normally expected behaviour

ifox:
I'm totally with you on this, just wanted to relate the 2 discussions and give more context to avoid introducing any breaking changes.

ifox (approved):

PR #2499, opened 2024-02-26
Replace phpcs with laravel/pint

Tofandel:
Now that pint is the standard for laravel project, it would be nice to switch to it, it uses phpcs under the hood but with an extensive set of rules

It really homogenizes the code and removes all the unused imports/variables

I added cghooks so that any code committed (with the project installed with composer) will automatically get pinted while committing

I'm aware that's a big diff and I can resubmit it without the diff if you prefer to apply it separatly (but tests will fail)

I also suggest to merge a maximum of PR first and I can rebase this PR to avoid conflicts with existing PRs


ifox:
I'm open to do this, thanks for suggesting it. I will ping you when it is ready to rebase.

Issue #2377, opened 2023-11-29
Repeater number squishes up sometimes when using @twillRepeaterTitleField


Bunksly:
## Description

![image](https://github.com/area17/twill/assets/94976871/b1fbb3fd-a4a6-4613-baec-43005cf8b982)

### Steps to reproduce

![image](https://github.com/area17/twill/assets/94976871/bba10bc5-027f-4518-a2b5-e5ee840ee000)

Using @repeaterTitleField to add long strings into the repeater "accordion panel title"

### Expected result

![image](https://github.com/area17/twill/assets/94976871/7b5fc7ef-70d2-4af3-902b-c9eff0ac0751)

### Actual result

![image](https://github.com/area17/twill/assets/94976871/de34d2d9-3623-425f-a397-73b2255f84fd)

### Versions

Twill version:  Version 3.0.2 
Laravel Framework 10.30.1
PHP 8.1.18
Database engine: mysql

PR #2501, opened 2024-02-27
fix: prevent block counter to shrink

Tofandel:
image

vs

image


ifox:
Thanks! This fixes #2377.

PR #2503, opened 2024-02-27
Fix availableBlocks override eachother

### LINKED ISSUE #2502: Editors with the same name override eachother's availableBlocks
## Description

When making a `BlockEditor::make()` with `->blocks()` or `->excludedBlocks()`  if any other editor (eg one in a block on in the form, or two from different blocks) on the page has the same name, then the available blocks of both blocks are the same (one overrode the available blocks of the other)

### Steps to reproduce

Create a few different blocks with nesting using

```php
BlockEditor::make()
            ->name('description')->label(twillTrans('Description'))
            ->withoutSeparator()
            ->excludeBlocks(['text']);
```
```php
BlockEditor::make()
            ->name('description')->label(twillTrans('Description'))
            ->withoutSeparator()
            ->excludeBlocks(['image']);
```

When you add one of this blocks into the other block, the nested block will show the wrong available blocks

This is due to this line which only uses the name of the block editor as a key
https://github.com/area17/twill/blob/df08d1affdf489c953376e6b7612b8381b15c985/views/partials/form/_block_editor.blade.php#L11

Tofandel:
Fixes #2502 

Simple fix is to add a unique id when rendering the component to find the correct availability key

Also included a bit of future proofing and backward compatibility even if the partial is overidden


frontend/js/components/blocks/Blocks.vue
```diff
@@ -265,7 +267,7 @@
         }
       },
       checkIfBlockTypeIsAvailable (editorName, type) {
-        const availableBlocks = JSON.parse(JSON.stringify(this.availableBlocks(editorName)))
+        const availableBlocks = this.availableBlocks(editorName)
```
Tofandel, line 270: No need to deepclone this in this function, it's only being searched

frontend/js/components/blocks/Blocks.vue
```diff
@@ -179,9 +182,8 @@
       }),
       ...mapGetters([
         'blocks',
-        'fieldsByBlockId',
```
Tofandel, line 182: This is not used in the component

PR #2506, opened 2024-02-28
fix: Table builder Browser column overrides parent module field

### LINKED ISSUE #2284: Table builder Browser column overrides parent module field
## Description

When using Table builder to add Browser column, its value overrides parent module field.

For example, parent module has a browser to other modules, and when I try to add Browser column to the index table to show related modules, their representation overrides parent modules "title" column.

```
$table->add(
    Browser::make()->browser('operation_types')->field('title')->title('Operation Types')
);
```

### Steps to reproduce

1. Create two modules.
2. Add browser to one of them which allows to add relations to other module.
3. Add Browser column to the module index table which takes related module "title" field.

### Expected result

Table should show Browser column as a separate column.

### Actual result

Table shows Browser as a separate column, but at the same time it overrides parent module title field.

### Versions

Twill version: 3.0.2
Laravel version: 10.13.0
PHP version: 8.1.0
Database engine: MySQL 8.0.29

zeezo887:
## Description

This PR overrides the base `TableColumn` class `getKey()` to prepend the browser's name to the browser key.

```
public function getKey(): string
    {
        if ($this->key === null) {
            throw new ColumnMissingPropertyException();
        }

        if (null === $this->browser) {
            throw new ColumnMissingPropertyException('Browser column missing browser value: ' . $this->field);
        }

        return "$this->browser.$this->key";
    }
```

## Related Issues

Fixes #2284

### DIFF
--- a/src/Services/Listings/Columns/Browser.php
+++ b/src/Services/Listings/Columns/Browser.php
@@ -27,4 +27,17 @@ class Browser extends TableColumn
             ->pluck($this->field)
             ->join(', ');
     }
+
+    public function getKey(): string
+    {
+        if ($this->key === null) {
+            throw new ColumnMissingPropertyException();
+        }
+
+        if (null === $this->browser) {
+            throw new ColumnMissingPropertyException('Browser column missing browser value: ' . $this->field);
+        }
+
+        return "$this->browser.$this->key";
+    }
 }


Tofandel:
Isn't it possible for two browsers to also have the same name?

zeezo887:
Browsers can use the same module, but cannot have the same name within the same form.

ifox (approved):

PR #2514, opened 2024-03-01
Fix alt text stripping accents and single quotes

Tofandel:
The altTextFrom method doesn't work well for non english characters and accents, it also uppercases every word which is weird


src/Helpers/media_library_helpers.php
```diff
@@ -63,9 +63,6 @@ function bytesToHuman($bytes)
      */
     function replaceAccents($str)
     {
-        if (function_exists('mb_convert_encoding')) {
-            return mb_convert_encoding($str, 'ASCII', 'UTF-8');
```
Tofandel, line 67: I had to remove this after adding tests `é` would be converted into `?` instead of `e` which could result in empty filenames

ifox (approved):

PR #2522, opened 2024-03-05
fix: allow buttonAsLink option on inline repeaters

### LINKED ISSUE #2521: buttonAsLink option doesn't work on Inline Repeaters
## Description

The `buttonAsLink` option has no effect in an inline repeater using the form builder.

### Steps to reproduce

Add an inline repeater field in a block component and add the buttonAsLink option.

```php
InlineRepeater::make()->name('buttons')
            ->fields([
                Input::make()->name('title'),
                Input::make()->name('url'),
            ])->buttonAsLink(true);
```
### Expected result

The button should display as a centred text link
### Actual result

`Undefined method 'buttonAsLink'`

### Versions

Twill version:  3.2.0

danieldevine:
## Description

Allows the `buttonAsLink` option on inline repeaters.

## Related Issues

Fixes #2521


src/Services/Forms/InlineRepeater.php
```diff
@@ -149,6 +150,13 @@ public function max(int $max): static
         return $this;
     }
 
+    public function buttonAsLink(bool $buttonAsLink): static
```
ifox:
```suggestion
    public function buttonAsLink(bool $buttonAsLink = true): static
```

For boolean properties like this we can set a default to true so that one can call the method without having to pass a boolean.

ifox (changes requested):

ifox (dismissed):

ifox (approved):

PR #2525, opened 2024-03-06
fix: Non-translated media fields display as empty

### LINKED ISSUE #483: Non-translated media fields display as empty
**Description of the issue:**

On a site with translated media fields, using a non-translated media field will display the field as empty in the edit form even though it is saved correctly in the database and holds data.

**Steps to reproduce:**

1. Enable translated media form fields by setting `twill.media_library.translated_form_fields` to `true`
2. Add a non-translated media form field to a module form by setting its `translated` parameter to `false`
3. Open the edit form for a single entry of that module
4. Attach an image from the media library to the non-translated media field
5. Save changes
6. Reload browser window

**Expected behavior:**

The media relation is saved in the database and displayed in the edit form on reload.

**Actual behavior:**

The media relation is saved correctly but the media form field is empty on reload.

**Notes:**

I think the `translated_form_fields` shouldn't be global. Some fields require it, some don't, so there should be a way of overriding the global setting for single fields. Twill's data structure is set up well for handling this. It's probably a question of how to detect the data format.

zeezo887:
## Description

This PR adds a new form field property `disableTranslate()` to force the field to disable translation when `twill.media_library.translated_form_fields
` is set to `true`.

Usage:

FormBuilder
```
Medias::make()
    ->name('cover')
    ->label('Cover image')
    ->disableTranslate()
    ->max(5)
```
Directive
```
@formField('medias', [
    'name' => 'cover',
    'label' => 'Cover image',
    'note' => 'Also used in listings',
    'fieldNote' => 'Minimum image width: 1500px',
    'disableTranslate' => true
])
```
FormView
```
<x-twill::medias
    name="cover"
    label="Cover image"
    note="Also used in listings"
    field-note="Minimum image width: 1500px"
    disable-translate
/>
```

## Related Issues

Fixes #483

### DIFF
--- a/frontend/js/mixins/block.js
+++ b/frontend/js/mixins/block.js
@@ -18,8 +18,9 @@ export default {
     open: function () {
       this.opened = true
     },
-    fieldName: function (id) {
-      return this.name + '[' + id + ']' // output : nameOfBlock[UniqID][name]
+    fieldName: function (id, extra = null) {
+      const fieldName = this.name + '[' + id + ']' // output : nameOfBlock[UniqID][name]
+      return extra ? fieldName + extra : fieldName
     },
     repeaterName: function (id) {
       return this.name.replace('[', '-').replace(']', '') + '|' + id // nameOfBlock-UniqID|name
--- a/src/Services/Forms/Fields/Medias.php
+++ b/src/Services/Forms/Fields/Medias.php
@@ -32,6 +32,8 @@ class Medias extends BaseFormField
 
     protected bool $activeCrop = true;
 
+    protected bool $disableTranslate = false;
+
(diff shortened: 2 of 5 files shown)


ifox (changes requested):
This is not working as expected, let's discuss and revisit.

ordigital:
This is not a adequate solution. Translation of media fields should work in the same way as in other types of fields, so just by using `->translatable()`. Translations should not be globally enabled by default, and there should not be a requirement to disable it in every field of an existing project, when a developer decides to add one translatable media field somewhere. No one would want this behavior in text fields, so why would anyone want this in media fields?

PR #2527, opened 2024-03-07
Prettify instructions and add the capsules>list array keys to improve DX

antonioribeiro:
# Description

Just to prettify instructions a bit and add the capsules>list array as people would need to remember/check that array structure the first time they create a Capsule.

## Before
Screenshot 2024-03-07 at 15 39 38

## After
Screenshot 2024-03-07 at 16 11 36

### DIFF
--- a/src/Commands/ModuleMake.php
+++ b/src/Commands/ModuleMake.php
@@ -386,8 +386,8 @@ class ModuleMake extends Command
         $newEntry = <<<PHP
 'capsules' => [
     'list' => [
+        ...
         [
-            ...
             
             'name' => '$name',
             'enabled' => true,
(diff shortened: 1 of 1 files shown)


src/Commands/ModuleMake.php
```diff
@@ -383,13 +383,21 @@ public function handle(): void
 
     private function addCapsuleToConfigFile(string $name): void
     {
-        $newEntry = [
-            'name' => $name,
+        $newEntry = <<<PHP
+'capsules' => [
+    'list' => [
+        [
+            ...
```
ifox:
```suggestion
        ...
        [
```

ifox (changes requested):

ifox (approved):

PR #2528, opened 2024-03-07
Allow case-insensitive search for translated models for postgres

### LINKED ISSUE #2112: CMS search case sensitivity with translated models and Postgres
## Description

When using translated models, the index search field in the CMS is case-sensitive.

### Expected result

When using translated models, the index search field in the CMS is case-insensitive.

zeezo887:
## Description

This PR overrides the `scopeOrWhereTranslationLike` method of `astrotomic/laravel-translatable` to select the like operator based on the database driver used.
```
public function scopeOrWhereTranslationLike(Builder $query, string $translationField, $value, ?string $locale = null): Builder
{
    return $this->scopeWhereTranslation($query, $translationField, $value, $locale, 'orWhereHas', getLikeOperator());
}
```

## Related Issues

Fixes #2112

### DIFF
--- a/src/Models/Behaviors/HasTranslation.php
+++ b/src/Models/Behaviors/HasTranslation.php
@@ -97,13 +97,6 @@ trait HasTranslation
             ->with('translations');
     }
 
-    /**
-     * @param Builder $query
-     * @param string $translationField
-     * @param $value
-     * @param string|null $locale
-     * @return Builder
-     */
     public function scopeOrWhereTranslationLike(Builder $query, string $translationField, $value, ?string $locale = null): Builder
     {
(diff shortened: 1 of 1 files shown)


src/Models/Behaviors/HasTranslation.php
```diff
@@ -97,6 +97,18 @@ public function scopeOrderByRawByTranslation($query, $orderRawString, $groupByFi
             ->with('translations');
     }
 
+    /**
+     * @param Builder $query
+     * @param string $translationField
+     * @param $value
+     * @param string|null $locale
+     * @return Builder
+     */
```
ifox:
```suggestion
```
We shouldn't need the docblock with the typed parameters and return.

ifox (changes requested):

ifox (approved):

PR #2529, opened 2024-03-08
fix: Allow file fields to be non translatable

### LINKED ISSUE #1712: Allow file fields to be non translatable
### Discussed in https://github.com/area17/twill/discussions/1690

<div type='discussions-op-text'>

<sup>Originally posted by **vinceinfluo** June 10, 2022</sup>
It seems like files are always translated, when setting the translated property to false in the form or setting 'translated_form_fields'  to false in the twill file_library doesn't work. 
It seems like this should be an option. Is there a way to solve this?
Thanks!</div>

zeezo887:
## Description

This addresses the issue where setting the `translatable(false)` method on a file field does not affect the field (i.e. Always translated) by adding a new `disableTranslate()` method on the file field.

Usage:
```
Files::make()
    ->name('single_file')
    ->label(twillTrans('Single file'))
    ->disableTranslate();
```

## Related Issues

Fixes #1712

### DIFF
--- a/frontend/js/mixins/block.js
+++ b/frontend/js/mixins/block.js
@@ -18,8 +18,9 @@ export default {
     open: function () {
       this.opened = true
     },
-    fieldName: function (id) {
-      return this.name + '[' + id + ']' // output : nameOfBlock[UniqID][name]
+    fieldName: function (id, extra = null) {
+      const fieldName = this.name + '[' + id + ']' // output : nameOfBlock[UniqID][name]
+      return extra ? fieldName + extra : fieldName
     },
     repeaterName: function (id) {
       return this.name.replace('[', '-').replace(']', '') + '|' + id // nameOfBlock-UniqID|name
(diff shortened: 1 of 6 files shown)


Tofandel:
Why create a new `disableTranslate` when there is already `translatable(false)`?

Lets not make the API more confusing shall we

ifox:
Let me clarify this. We can agree this is not pretty, but it is temporary to avoid a breaking change. 

In https://github.com/area17/twill/pull/2525 I asked @zeezo887 to add a new option to work around the current situation with the `translate` prop on `medias` field that isn't working because of the global configuration option. Similarly, the `files` field currently forces translations without any ability to control it globally or at the field level. 

To avoid breaking changes in Twill 3 while enabling the missing functionality we introduce these temporary methods. In a future breaking release we will fix the behavior of the `translate` prop and `translatable` method, remove the global configuration option for medias fields, as well as these unnecessary methods.

PR #2536, opened 2024-03-12
fix(Buckets): save featured type using morph map instead of module name

Tofandel:
## Description

Currently when using the bucket feature, the morphMap has to match *exactly* the name of the twill module, this is very restrictive especially in projects where the morphMap was already defined, because the featured_type column is saved using the module name instead of the MorphMap of the class

This effectively results in a fatal error when loading the Feature because the Morph fails

This PR aims to fix this issue by correctly saving the Feature using the morph map

It also fixes the label displayed here to be taken from the bucketable config rather than the current behavior of ucfirst the module name on reload
image
And it improves the performance of the bucket controller which currently loads all of the repositories in any case even though only one is ever displayed at a time, it now only loads the first one or the requested one

Also added a documented 'repository' option for bucketables to allow passing the repository class directly (or it cannot work with Package capsules)

PS: it would also be a nice addition to buckets to add an option to edit the featured model either clicking on the title or with an edit button somewhere here
image


LucaRed:
Thanks, your PR saved me! I couldn't make the Buckets feature work, even though I followed the [official docs](https://twillcms.com/docs/buckets/index.html).

The trick was to specify both the module and the repository (which is not even mentioned in the docs) as fully qualified names.

```
'module'     => \App\Models\Article::class,
'repository' => \App\Repositories\ArticleRepository::class,
```

I hope this PR will be merged soon.

PR #2550, opened 2024-03-18
Fix user creation is missing a role input if role is Administrator or lower

### LINKED ISSUE #2548: Twill users creation and roles permissions
When trying to create a user with the roles and permissions feature, it seems a role input is missing

image

Tofandel:
Fixes #2548 

Also fixes the "All" option being in the role list

### DIFF
--- a/views/users/create.blade.php
+++ b/views/users/create.blade.php
@@ -11,16 +11,14 @@
         type="email"
     />
 
-    @can('edit-user-roles')
-        @php $userModel = twillModel('user') @endphp
+    @php $userModel = twillModel('user') @endphp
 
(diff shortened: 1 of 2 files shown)


views/users/create.blade.php
```diff
@@ -11,16 +11,14 @@
         type="email"
     />
 
-    @can('edit-user-roles')
```
Tofandel, line 14: Should we check if `TwillPermissions::enabled()` instead?

Tofandel: Nope, when permissions are disabled, roles still exist but taken from an enum

PR #2561, opened 2024-03-27
Vue.js 3 upgrade

zeezo887:
## Description

This PR intends to migrate Twill's frontend from Vue 2 to Vue 3.


frontend/js/components/Browser.vue
```diff
@@ -13,11 +13,13 @@
         </div>
         <div class="browser__search">
           <a17-filter @submit="submitFilter">
-            <div slot="additional-actions" v-if="browserNote">
-              <span class="browser__note f--small">
-                {{ browserNote }}
-              </span>
-            </div>
+            <template v-slot:additional-actions>
```
Tofandel, line 16: You can also use `#additional-actions` instead of `v-slot` which I personnally find easier to read and recognize it as a slot

PR #2569, opened 2024-04-01
Default login redirect to admin url

### LINKED ISSUE #2568: Redirected to '/' on reset password
## Description

While configurable with undocumented config `twill.auth_login_redirect_path` I think it should default to the url of the admin panel by default

Tofandel:
Fixes #2568

### DIFF
--- a/config/twill.php
+++ b/config/twill.php
@@ -172,7 +172,7 @@ return [
     |--------------------------------------------------------------------------
     |
      */
-    'auth_login_redirect_path' => '/',
+    'auth_login_redirect_path' => null,
 
     'templates_on_frontend_domain' => false,
 
--- a/src/Http/Controllers/Admin/LoginController.php
+++ b/src/Http/Controllers/Admin/LoginController.php
@@ -2,8 +2,8 @@
 
 namespace A17\Twill\Http\Controllers\Admin;
 
+use A17\Twill\Facades\TwillRoutes;
 use A17\Twill\Http\Requests\Admin\OauthRequest;
-use A17\Twill\Models\User;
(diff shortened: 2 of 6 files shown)


ifox:
Thanks @Tofandel, I agree however this PR doesn't actually change the default config value, right? We can document that it can be set to null for now to get the new default behavior and make the config breaking change in Twill 4.

Tofandel:
No it doesn't, I forgot about it, the config is set by default there https://github.com/area17/twill/blob/3.x/config/twill.php#L175, I and so this PR doesn't apply unless explicitely set to null, don't think changing it to null would be considered a breaking change would it?

ifox:
If we set it to null I'm afraid the case where one is using both an admin domain and a path at the same time would break. It's very much an edge case but I think the mistake was to release Twill 3 with the path approach as the default without updating the configuration, so we could technically consider this a bug I guess.

Tofandel:
> admin domain and a path at the same time 

Not sure I understand the edge case? If we set just `twill.auth_login_redirect_path` to null, all the other instances of usage are already replaced by the PR, so then it just uses the new default behavior

ifox:
You're right.

Do we need a new helpers class for this though? I'm curious why you didn't put this under `TwillRoutes`.

Tofandel:
`TwillRoutes` is a good place for it, I didn't see it, I just wanted to create a Route helper to prepare for the moving of the https://github.com/area17/twill/blob/3.x/src/Helpers/routes_helpers.php to it, but I guess since there is only one method in it, it can also move to that class, which already fits well for that purpose

ifox (approved):

PR #2581, opened 2024-05-06
feat: added clearable method to select form field

### LINKED ISSUE #2562: Can't unselect option from Select (form field)
## Description

After selecting an option in the Select, you can't unselect. There is also no option/config available to deal with empty options.

zeezo887:
## Description

Allows the user to clear the selected option:

```
Select::make()
  ->name('field')
  ->options(
      Options::make([
          Option::make('value', 'label'),
          Option::make('value2', 'label2'),
      ])
  )
  ->clearable()
```

## Related Issues

Fixes #2562


Tofandel:
I think `clearable` makes more sense, then we could show a clear button on the select input

Tofandel:
There seems to be already what is needed on the VSelect component to make it work https://github.com/area17/twill/blob/3.x/frontend/js/components/VSelect/ExtendedVSelect.vue#L42-L44C7

But it seems it's not the component that is used https://github.com/area17/twill/blob/3.x/frontend/js/components/SingleSelect.vue

ifox:
I implemented clearable instead as discussed. By default the Select field does use vue-select. It doesn't only if `unpack` or `native` are true.

PR #2599, opened 2024-06-05
Improve related save

Tofandel:
A quick PR to improve `saveRelated` to work with models, I wanted to use it and according to the phpdoc it accepts an array of Twill model, but this is not the case (it only accepts an array with `id` and `endpointType`) 

This PR allows the method to accept models as well and improves the performance by using only one `where` instead of 3 which would iterate 3 times over the collection

I also added `$item->unsetRelation('related')` in loadRelated, otherwise there is a circular reference in the model relation (model -> pivot -> related -> model ...) which causes an infinite loop when using `model->toArray()`


src/Models/Behaviors/HasRelated.php
```diff
@@ -63,18 +63,24 @@ public function loadRelated(string $browserName): Collection
...
+                $type = $item->getMorphClass();
+            } else {
+                $id = $item['id'];
+                $type = $item['endpointType'];
+            }
+
             $firstMatchKey = $itemsToProcess
-                ->where('related_id', $item['id'])
-                ->where('related_type', $item['endpointType'])
-                ->where('browser_name', $browserName)
```
Tofandel, line 77: This is already taken care of in line 71

ifox (approved):

PR #2606, opened 2024-06-11
Fix published scope

Tofandel:
If the model does not have the publishing feature and column, then the published scope throws an error

It's called in `TwillUtil::parseInternalLinks` for example, indiscriminately of the model's features

### DIFF
--- a/src/Models/Model.php
+++ b/src/Models/Model.php
@@ -33,12 +33,15 @@ abstract class Model extends BaseModel implements TaggableInterface, TwillModelC
         return Str::endsWith(get_class($this), 'Translation');
     }
 
-    public function scopePublished($query): Builder
+    public function scopePublished(Builder $query): Builder
     {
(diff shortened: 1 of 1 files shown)


src/Models/Model.php
```diff
@@ -33,12 +33,15 @@ protected function isTranslationModel(): bool
         return Str::endsWith(get_class($this), 'Translation');
     }
 
-    public function scopePublished($query): Builder
+    public function scopePublished(Builder $query): Builder
```
Tofandel, line 36: All of the scopes have `Builder $query` in the twill interfaces, I'm actually really surprised php isn't complaining that the signature isn't compatible

https://github.com/area17/twill/blob/6d022399eb8ab1d76aaea3e49d76fece54d62cc1/src/Models/Contracts/TwillModelContract.php#L15-L26

ifox (approved):

PR #2612, opened 2024-06-13
Rethink the way the error handler works

### LINKED ISSUE #2607: The exception handler is overriden if calling $model->getFullUrl() on the frontend
The method `getFullUrl` has an unexpected side effect, because it constructs the controller of the model which binds a different exception handler even if the request doesn't originate from twill

This is problematic as then exceptions are not rendering as expected or reported to sentry

https://github.com/area17/twill/blob/6d022399eb8ab1d76aaea3e49d76fece54d62cc1/src/Models/Model.php#L185-L197
https://github.com/area17/twill/blob/6d022399eb8ab1d76aaea3e49d76fece54d62cc1/src/Http/Controllers/Admin/Controller.php#L20-L25

(And yes I spent 3 hours trying to figure out why my 404 pages were rendering like this in production 
image instead of with my  custom view, long story short I have a menu that uses a Page model and gets the url of those pages using getFullUrl)

Tofandel:
Fixes #2607 

We can use the user provided handler and simply add render cases via methods, so we don't need to provide our own exception handler, which will alleviate the need for more configuration/setup when installing twill

I also optimized the provider to have all the config in one place and not have it run if the config is already cached


ifox:
This is great @Tofandel thank you. The custom exception handler has definitely been a source of issues and this approach is so much better.

src/TwillRoutes.php
```diff
@@ -402,4 +402,18 @@ public function getAuthRedirectPath(): string
         return config('twill.auth_login_redirect_path')
             ?? rtrim(config('twill.admin_app_url') ?: '', '/') . '/' . ltrim(config('twill.admin_app_path') ?? 'admin', '/');
     }
+
+    public function isTwillRequest(): bool
```
ifox, line 408: I think we have something like this elsewhere, will look for it.

Tofandel: I did a search on 'twill.admin_app_url' and 'twill.admin_app_path' and found nothing resembling this (except non exposed in the ExceptionHandler which I just removed)

ifox: You're right, I was remembering this from a packaged capsule, not in the core.

ifox (approved):

PR #2624, opened 2024-06-27
 Preview - Update default width value for the mobile preview (to a more realistic value)

mrdoinel:
## Description

 Preview - Update default width value for the mobile preview (to a more realistic value 390px)

## Related Issues

320px is not widely used anymore.
https://gs.statcounter.com/screen-resolution-stats/mobile/united-states-of-america

### DIFF
--- a/frontend/js/components/Previewer.vue
+++ b/frontend/js/components/Previewer.vue
@@ -92,7 +92,7 @@
             name: 'preview-tablet-v'
           },
           {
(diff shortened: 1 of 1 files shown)


ifox:
I agree with updating the default. You can use this until this is released: https://github.com/area17/twill/pull/2299

ifox (approved):

PR #2625, opened 2024-06-27
modified after save handlers for medias and files

zeezo887:


ifox:
@Tofandel we noticed a regression on Twill 3.3 following your last [PR](https://github.com/area17/twill/pull/2567) touching this. Would you mind reviewing this?

Tofandel:
What is the regression? I'm having a hard time understanding what is this PR achieving that `sync` doesn't

ifox:
There are multiple reports since 3.3 where subsequent saves of medias fields are resulting in incorrect pivot rows in the `twill_mediables` table. 

https://discord.com/channels/811936425858695198/1254442932400816148

https://discord.com/channels/811936425858695198/1254726742556414062

PR #2628, opened 2024-07-02
Fix sync of medias and files with multiple fields

Tofandel:
If you have multiple media fields, then the sync would drop some records quite randomly because the unicity was made on the media id and not media_id + role

This PR superseedes #2625 and fixes the issue in a more generic manner using the primary key of the record (laravel is normally built so that pivots don't need/use primary keys hence the issue)

### DIFF
--- a/src/Models/Behaviors/HasFiles.php
+++ b/src/Models/Behaviors/HasFiles.php
@@ -18,7 +18,7 @@ trait HasFiles
             File::class,
             'fileable',
             config('twill.fileables_table', 'twill_fileables')
-        )->withPivot(['role', 'locale'])
+        )->withPivot(['id', 'role', 'locale'])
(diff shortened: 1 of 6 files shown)


ifox (approved):

galingong:
Thanks for fixing this @Tofandel. My client just complained that after the update, medias were acting weird. Since when was this issue present? I don't remember encountering it in earlier versions. I'd say it's a fairly high impact bug, everyone who updated recently have had broken medias fields for at least 2 months now - have you considered a minor fix release for this @ifox ?

ifox:
We didn't realize the impact at first, but yes, I am planning a minor release with this this week, actually. FYI, it only affects forms and blocks that have more than 1 medias or files fields.

ifox:
This is a 3.3 regression, the bug isn't present before that.

galingong:
Ah thanks, OK then it's easy, I'll just simply downgrade to 3.2.x until then.

PR #2633, opened 2024-07-11
Fix server error in the dashboard when a nested module has a deleted parent

Tofandel:
If you have a nested module that you want to show in the dashboard, if you trash a parent without soft deleting the subject, then the dashboard throws an error 500 when it tries to display any activity that happened on one of the children of the deleted parent

```Missing required parameter for [Route: twill.articleCategories.articles.edit] [URI: admin/articleCategories/{articleCategory}/articles/{article}/edit] [Missing parameter: article]```

The error is because the empty parent is ignored and twill tries to use the `moduleRoute` as if it was a non nested module route, instead of ignoring the record

This PR fixes this issue

### CODE REVIEW

File: src/Http/Controllers/Admin/DashboardController.php [hunk 1]
```diff
@@ -554,10 +572,21 @@ private function getDrafts(Collection $modules): Collection
             if ($repository->hasBehavior('revisions')) {
                 $query->mine();
             }
-            $parentRelationship = $module['parentRelationship'] ?? null;
+            
+            if (isset($dashboardModule['parentRelationship'])) {
+                /** @var BelongsTo $parent */
+                $parent = call_user_func([$activity->subject, $dashboardModule['parentRelationship']]);
```
Tofandel, line 578:
```suggestion
                $query->whereHas($dashboardModule['parentRelationship']);
```

src/Http/Controllers/Admin/DashboardController.php
```diff
@@ -554,10 +572,21 @@ private function getDrafts(Collection $modules): Collection
             if ($repository->hasBehavior('revisions')) {
                 $query->mine();
             }
-            $parentRelationship = $module['parentRelationship'] ?? null;
+            
+            if (isset($dashboardModule['parentRelationship'])) {
```
Tofandel, line 576:
```suggestion
            if (isset($module['parentRelationship'])) {
```

src/Http/Controllers/Admin/DashboardController.php
```diff
@@ -554,10 +572,21 @@ private function getDrafts(Collection $modules): Collection
...
-                $parent = $draft->$parentRelationship;
+            return $query->get()->map(function ($draft) use ($module, $parent) {
+                if (isset($parent)) {
+                    $parent_id = $parent->getParentKey();
+        
+                    if (empty($parent_id)) {
+                        // Prevent module route error
+                        return null;
+                    }
+                }
```
Tofandel, line 589:
```suggestion
            return $query->get()->map(function ($draft) use ($module) {
                if (isset($module['parentRelationship'])) {
                    $parent_id = call_user_func([$draft, $module['parentRelationship']])->getParentKey();
                }
```

src/Http/Controllers/Admin/DashboardController.php
```diff
@@ -136,16 +137,25 @@ public function search(Request $request): Collection
                     $date = $item->created_at->toIso8601String();
                 }
 
-                $parentRelationship = $module['parentRelationship'] ?? null;
-                $parent = $item->$parentRelationship;
+                if (isset($dashboardModule['parentRelationship'])) {
+                    /** @var BelongsTo $parent */
+                    $parent = call_user_func([$item, $dashboardModule['parentRelationship']]);
```
Tofandel, line 142:
```suggestion
                    $parent = call_user_func([$item, $module['parentRelationship']]);
```

src/Http/Controllers/Admin/DashboardController.php
```diff
@@ -554,10 +572,16 @@ private function getDrafts(Collection $modules): Collection
             if ($repository->hasBehavior('revisions')) {
                 $query->mine();
             }
-            $parentRelationship = $module['parentRelationship'] ?? null;
+            
+            if (isset($module['parentRelationship'])) {
+                /** @var BelongsTo $parent */
```
Tofandel, line 577:
```suggestion
```

ifox (approved):

PR #2639, opened 2024-07-31
Fix menu order 9_adding-navigation.md

plokko:


Amiejah:
@plokko Not sure if this PR is going to be picked up, but I believe that if any ordering method etc is missing we should add it to the HasPosition trait. 

The MenuLink uses the HasPosition trait and this contains a scopeOrdered() method. So you could do something like: 

`MenuLink::published()->ordered()->get()->toTree()`

PR #2668, opened 2024-10-14
Fix the close datepicker with no update situation when manually typing time

### LINKED ISSUE #2667: Datepicker not updating the state on close
## Description

Manually entering the time does not update the state.

### Steps to reproduce

1. Open the **published from** datepicker
2. Type the time using the hour / minute input fields (not the increase / decrease arrows)
3. Press enter or click outside the datepicker.

### Expected result

**Published on** date and time should change both in the datepicker field and in the accordion trigger. Pressing update should update with the selected date & time.

### Actual result

**Published on** date and time is changed in the datepicker field but not in the accordion trigger. Pressing update updates with the previously set date and time.

### Versions

Twill version:  3.3.1
Laravel version: 10.43

alextepes:
## Description

Fixes a datepicker problem occurring when manually typing the time.

## Related Issues

Fixes #2667


frontend/js/components/DatePicker.vue
```diff
@@ -178,6 +178,7 @@
           },
           onClose: function (selectedDates, dateStr, instance) {
             self.$nextTick(function () { // wait for the datepicker to properly update the UI
+              self.date = dateStr;
```
Tofandel, line 181: I think it would make more sense to maybe call this.saveIntoStore onBlur if this.disabled is false

Tofandel: After testing myself, the onBlur method is never called and this does seem to be the only reliable fix

baeroe:
can we please merge this one?

Issue #2682, opened 2024-10-29
Relation columns not displaying correctly


alextepes:
## Description

When adding a A17\Twill\Services\Listings\Columns\Relation column to a listing, a string of commas and spaces is displayed instead of the display field. 

### Steps to reproduce
```php
use A17\Twill\Services\Listings\Columns\Relation;
```

```php
    public function getIndexTableColumns(): TableColumns
    {
        $tableColumns = TableColumns::make();

        $tableColumns->add(
            Text::make()
                ->title('Title')
                ->field('title')
                ->linkToEdit()
                ->sortable(),
        );
        $tableColumns->add(
            Relation::make()
                ->relation('category')
                ->title('Category')
                ->field('title')
                ->sortable()
        );
        return $tableColumns;
    }
```

In the model:
```php
    public function category()
    {
        return $this->belongsTo(Category::class);
    }
```

### Expected result

The title of the column should be displayed on the "Category" column.

### Actual result

, , , , , , , , , , , , , is displayed on the "Category" column.
<img width="1455" alt="Screenshot 2024-10-29 at 12 00 22" src="https://github.com/user-attachments/assets/1b6477a1-d0db-48b6-9a4f-82150eb6d4af">

### Versions

Twill: 3.4.1
Laravel: 11.29
PHP 8.2.24
MariaDB

zachgarwood:
I have also submitted a fix for this: #2720

PR #2683, opened 2024-10-29
Fix for one to one relation column in index table

iedex:
## Description

Not sure if this is the best solution, but this way works for both relation types that I use (belongsTo and belongsToMany).
Before this change, `collect($model->getRelation($this->relation));` for belongsToMany, and worked:
authors

but this for a belongsTo relation:
category
I had something like ", , , , , , , , , , ," on that column.

### DIFF
--- a/src/Services/Listings/Columns/Relation.php
+++ b/src/Services/Listings/Columns/Relation.php
@@ -46,7 +46,8 @@ class Relation extends TableColumn
 
         /** @var \Illuminate\Database\Eloquent\Collection $relation */
         $model->loadMissing($this->relation);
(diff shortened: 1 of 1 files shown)


alextepes:
This is related to #2682.
I have also submitted a similar merge request #2684 before noticing yours @iedex.

zachgarwood:
I have also submitted a fix for this: https://github.com/area17/twill/pull/2720

ifox:
Thanks everyone, I went with @zachgarwood's solution, `Collection::wrap()` is exactly what this needed.

ifox:
Fixed in https://github.com/area17/twill/pull/2720

PR #2688, opened 2024-11-01
retrieve slug from slugs table if the $slugAttribute field was not changed

### LINKED ISSUE #2681: All module slugs get reset on reorder.
## Description

When using self-nested module and reordering, the slugs for any nested pages get reset when one of the nested items gets reordered.

### Steps to reproduce

1. Setup a self-nested module as per the Twill docs. 
2. Enable re-ordering 
3. Have a parent page and 2+ nested pages, such as below
image
4. As per the above image, manually change one of the nested slugs to something different to default
5. Change the order of the nested pages

### Expected result

The order of these pages should update, but the slugs should stay the same

### Actual result

The order is updated, but the slugs reset back to their default, as per here:
image

### Versions

Twill: 3.4.1

Laravel: 11.29.0

PHP: 8.3.4

MySQL: 8.0.30

zeezo887:
## Description
The changes made here was to use the previous slug value if the `$slugAttribute` field was not changed. By doing this, we won't have duplicate slug entries when a model is not updated via the repository.

```
/**
 * Returns changed value of slugAttribute field or previous slug value
 */
private function getSlugValue($slugAttribute, $locale): mixed
{
    return $this->wasChanged($slugAttribute)
        ? $this->$slugAttribute
        : $this->slugs()->where('locale', $locale)
            ->where('active', true)->value('slug');
}
```

## Related Issues
Fixes #2681

### DIFF
--- a/src/Models/Behaviors/HasSlug.php
+++ b/src/Models/Behaviors/HasSlug.php
@@ -124,9 +124,9 @@ trait HasSlug
      */
     public function handleSlugsOnSave(): void
     {
-        $this->disableLocaleSlugs();
-
         $slugParams = $this->twillSlugData !== [] ? $this->twillSlugData : $this->getSlugParams();
+
+        $this->disableLocaleSlugs();
 
         foreach ($slugParams as $params) {
             if (in_array($params['locale'], config('twill.slug_utf8_languages', []))) {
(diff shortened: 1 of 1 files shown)


Tofandel:
Is the disableLocaleSlug even needed before? This seems to be done already in updateOrNewSlug and I'm sure it's interfering with getOldSlug->active

src/Models/Behaviors/HasSlug.php
```diff
@@ -358,9 +358,13 @@ public function getSlugParams(?string $locale = null): ?array
                     throw new \Exception("You must define the field {$slugAttribute} in your model");
                 }
 
+                $slug = $this->getSlugValue($slugAttribute, $translation->locale)
+                    ?? $translation->$slugAttribute
```
Tofandel, line 362: This will always return the default translation instead of the current locale

src/Models/Behaviors/HasSlug.php
```diff
@@ -415,6 +419,21 @@ public function getSingleSlugParams(?string $locale = null): ?array
...
+    /**
+     * Returns changed value of slugAttribute field or previous slug value
+     *
+     * @param $slugAttribute
+     * @param $locale
+     * @return mixed|null
+     */
+    private function getSlugValue($slugAttribute, $locale): mixed
+    {
+        return $this->wasChanged($slugAttribute)
```
Tofandel, line 431: This fix doesn't work for slugs from translations, because then translations are never detected as changed and it will return the slug before change

Issue #2689, opened 2024-11-05
Caption of Image does not persist


emil-es:
## Description

when using the block editor with a medias form input, the metadata s caption doesn't persist to db. 

### Steps to reproduce

install a fresh laravel app, install twill, install twill basic-page-builder

### Expected result

Create a new page. Save page with an image block, edit info - fill caption field. Reload page, caption field is persisted.

### Actual result

Create a new page. Save page with an image block, edit info - fill caption field. Reload page, caption field is empty.

### Versions

Twill version:  3.4.1
Laravel version: 11.30.0
PHP version: 8.3.13
Database engine: sqlite

emil-es:
reverting back to twill 3.3.1, this works as expected

LucaRed:
Not only for image blocks, the caption is not saved for any Medias field.

PR #2694, opened 2024-11-14
Allow positioning media files

### LINKED ISSUE #2690: Cannot reorder images on A17\Twill\Services\Forms\Fields\Medias
## Description

We cannot reorder images on version 3.4.1.

### Steps to reproduce

It's very simple, change the order, update, refresh the cms page and the images are back on the old order.

### Versions

Twill version:  3.4.1

zeezo887:
## Description

I added a migration file to include a position column in the `mediable` and `fileable` tables, enabling the retrieval of items in a specific order based on their position.

Additionally, there is a bug in the `$this->attachNew($records, $current, false)` method when updating existing pivot records. The issue arises because:
```
$this->newPivotStatementForId($this->parseId($id))->update(
    $this->castAttributes($attributes)
);

public function newPivotStatementForId($id)
{
    return $this->newPivotQuery()->whereIn($this->relatedPivotKey, $this->parseIds($id));
}
```
This code uses the `relatedPivotKey` (such as `media_id` or `file_id`) to locate the record for updating. However, in our case, the IDs passed to this method are the IDs of the pivot table itself, not `media_id` or `file_id`. To resolve this, I overrode the `attachNew` method and filtered by the pivot table ID:

```
$updated = $this->newPivotQuery()->whereIn('id', $this->parseIds($id))->update(
    $this->castAttributes($attributes)
);
```

## Related Issues

Fixes #2690 , #2689


ifox (approved):

LucaRed:
If I revert back to Twill 3.3.1, the positioning of media files used to work, without the need for this new migration. I'm just afraid there's something broken hiding somewhere.

Issue #2693, opened 2024-11-13
Saving slugs results in memory exhaustion


Tofandel:
## Description

Just saving a non unique slug in a database that has a lot of entries (Not like an insane amount, just 500) seems to result in a lot of queries and then memory exhaustion

I already increased the memory to 512Mo and it was ok for a few weeks, but now it exceeds the 512Mo limit again

Something is terribly wrong in there, here is the query trace from sentry as we can see there is some huge queries being run

![image](https://github.com/user-attachments/assets/47b08c0f-1624-4da2-bbc3-080fc1ee1611)

![image](https://github.com/user-attachments/assets/6a749ea9-eb92-4ffb-bc62-52e03c4bba6f)

![image](https://github.com/user-attachments/assets/bbc322b7-a904-4a9b-94ee-b61c43ea1bd5)

I will try to investigate the cause

Tofandel:
![image](https://github.com/user-attachments/assets/c4e92479-6ae1-4d93-afa0-90bd267354cb)

So apparently the slug table has grown to an incredible size (70Mb) with each entry having hundreds of duplicated non active slugs

ifox:
Ugh, this sounds really bad, we'll be looking into it asap. 

Does this [PR](https://github.com/area17/twill/pull/2688) help by any chance? I have not reviewed it yet.

Tofandel:
Very likely yes, not sure if the PR solves it, but the issue should definitely because of the same thing, I have been looking at the query log and it seems 3 queries are duplicated (because a function is called multiple times when the result could be cached) in the code path and the one retrieving the list of all the existing ids is not even needed (it can be inlined with a not condition on the foreign key), I will send in a PR with those improvements

Issue #755, opened 2020-10-28
[Missing docs] Slugs and slugAttributes


henze-housepedia:
It seems straightforward, but when you need a concatenation of a string it is not. When you want to make a concatenation of an id and a title, you might think: "Hey, this $slugAttributes is an array of column names. I just do the following":

`public $slugAttributes = [
        'id',
        'title',
    ];`

Right? Wrong! the first is being used for the slug. The rest is used for dependencies. I don't know what this means, but it breaks if the columns aren't in the slugs table.
If you need a concatenation of two strings, just do the following:

`public $slugAttributes = [
        'slugifyThis',
    ];

    public function getSlugifyThisAttribute(): string
    {
        return $this->id.' '.$this->name;
    }`

 This does only work when not using translations. If you need translations to work, I suggest a rewrite of the getSlugParams. (create your own trait, use HasSlug {getSlugParams as public getTwillSlugParams;} )

This does nothing with the permalink, as this is being generated client side, and there are NO options to alter it's behavior, so advice is to turn permalinks off or deal with the fact that it does not show you what it is going to be.

As for the dependency thing, @ifox?

PR #2696, opened 2024-11-20
Review end to end the slug saving flow

### LINKED ISSUE #2681: All module slugs get reset on reorder.

Tofandel:
A few things this PR tackle:

1. Performance of saving slugs: we now check if the slug attributes were modified to know if the slug needs to be recomputed, this means no query at all if no slug change. A lot of unnecessary or duplicated queries were also removed. A single slug creation before needed between 10 and 16 queries and now only runs between 3 and 6

2. Get slugParams was split between 2 functions which for the most part duplicated the logic. It is now all encompassed into the original function, deprecating `getSingleSlugParams`

3. There is a lot of confusion regarding multiple `slugAttributes` https://github.com/area17/twill/issues/755 as well as in the code, some parts of the code used slugAttributes to generate a concatenated slug, some other parts of the code assume that the first is the slug and the rest is additional columns in the slug table (but the later doesn't make much sense because it would just copy an existing attribute into the slug table, so I do not understand what it achieves).
To avoid a breaking change a $slugFields property and $slugDeps property were introduced, if they are not defined, it will fallback to $slugAttributes to determine those

Superseedes #2688 
Fixes #2681 #2693

The new `$slugFields` goes a long way in the path to fully support https://github.com/area17/twill/discussions/860 (right now it will correctly generate it from all the fields in the backend but not yet the browser side)


tests/integration/SlugTest.php
```diff
@@ -117,49 +158,25 @@ public function testCanReuseSoftDeletedSlugWithHistory(): void
...
         $this->assertTrue($this->module->getRepository()->restore($model->id));
 
-        $model = $this->module->getModelClassName()::find($model->id);
+        $model = $model->fresh();
 
         $this->assertCount(2, $model->slugs()->get());
         $this->assertEquals('slug-update-2', $model->getSlug());
     }
-
-    public function testReactivateSlug(): void
```
Tofandel, line 141: This test was removed because it entirely duplicated the `testSlugLooping`

ifox:
Thank you so much for diving into this @Tofandel.

We do have a few projects that use multiple slug attributes. This is used in tandem with parent-child modules, so that children of different parents can reuse the same slug without a suffix. So the second attribute we add in that case (and a column in the slugs table) is the parent ID.

Tofandel:
In that case maybe I can instead add a new property for slugAttributes so that we can get the new behavior without a breaking change

src/Repositories/Behaviors/HandleSlugs.php
```diff
@@ -15,34 +15,28 @@ public function beforeSaveHandleSlugs(TwillModelContract $object, array $fields)
...
-                    $slugParams = $this->model->slugAttributes;
-                    $slugData = [];
-
-                    foreach ($slugParams as $param) {
-                        $slugData[] = $fields[$param][$locale] ?? '';
-                    }
-
-                    if (!empty(Arr::join($slugData, '-'))) {
-                        $object->twillSlugData[] = [
-                            'slug' => Str::slug(Arr::join($slugData, '-')),
```
Tofandel, line 40: This is the part of the code that does concatenation on slugAttributes

PR #2705, opened 2024-12-05
Use JS to read image dimensions on upload

13twelve:
#### The issue

Some jpegs are taken in one orientation and then rotated on a device and exported such that the display of the image doesn't seem to match the dimensions. The result is you upload what you think is a portrait image but is actually a landscape image, which EXIF data set to rotate the image. This then gives un-expected results in the image cropping tool and on the front end.  

#### The fix

Halt the fine-uploader by using a promise on the submit callback of fine-uploader, until after the image has loaded and JavaScript has had a change to grab the image dimensions. Previously the upload was completing before the image had been loaded and so image dimensions were not being sent.

And then we're using the supplied values from the front end when storing the values.

### DIFF
--- a/src/Http/Controllers/Admin/MediaLibraryController.php
+++ b/src/Http/Controllers/Admin/MediaLibraryController.php
@@ -185,11 +185,11 @@ class MediaLibraryController extends ModuleController implements SignUploadListe
 
         $uploadedFile = $request->file('qqfile');
 
-        [$w, $h] = getimagesize($uploadedFile->path());
-
         if ($request->input('width') && $request->input('height')) {
             $w = $request->input('width');
             $h = $request->input('height');
+        } else {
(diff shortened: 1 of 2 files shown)


src/Http/Controllers/Admin/MediaLibraryController.php
```diff
@@ -187,6 +187,11 @@ public function storeFile($request)
 
         [$w, $h] = getimagesize($uploadedFile->path());
```
ifox: @13twelve let's move this into an else below to avoid the extra call if we don't need it?

ifox (changes requested):

src/Http/Controllers/Admin/MediaLibraryController.php
13twelve: Updated

ifox (approved):

PR #2732, opened 2025-02-24
add tiptap's subscript and superscript extansion to tiptap wysiwyg

tommycornilleau:
Add subscript or superscript content in TipTap wysiwyg

Subscript uses @tiptap/extension-subscript and wrap wysiwyg text within a `<sub />` tag.
Superscript uses @tiptap/extension-superscript and wrap wysiwyg text within a `<sup />` tag.

Capture d’écran 2025-02-24 à 12 07 04


frontend/js/components/WysiwygTiptap.vue
```diff
@@ -371,6 +381,8 @@
             bold: true,
             italic: true,
             underline: true,
+            subscript: true,
+            superscript: true,
```
ifox: This will enable them by default on all Tiptap fields where the developer passes an empty array of options, which I don't think should be the case in a minor release.

src/Services/Forms/Fields/Wysiwyg.php
```diff
@@ -26,6 +26,8 @@ class Wysiwyg extends BaseFormField
         'bold',
         'italic',
         'underline',
+        'subscript',
+        'superscript',
```
ifox: Same here, if the developer does not specify their options in a field, they will end up with new buttons after updating Twill, which I consider as a "breaking" change. We should remove it from this default configuration for now. We might decide to add them back for the next major release.

package.json
```diff
@@ -25,6 +25,8 @@
     "@tiptap/extension-table-row": "^2.2.1",
     "@tiptap/extension-text-align": "^2.2.1",
     "@tiptap/extension-underline": "^2.2.1",
+    "@tiptap/extension-subscript": "^2.11.5",
+    "@tiptap/extension-superscript": "^2.11.5",
```
ifox, line 29: When new npm dependencies are added, the package.lock file needs to be pushed to the repository.

ifox (changes requested):

frontend/js/components/WysiwygTiptap.vue
tommycornilleau: Indeed, this was probably too much

tommycornilleau:
Changes request has been addressed.
Subscript and superscript are no longer default options and package-lock has been added

ifox (approved):

PR #2752, opened 2025-03-27
Implement endpoint params in Browser form field

### LINKED ISSUE #2751: Endpoint params not implemented in Browser form field
## Description

The documentation for the Browser form field (https://twillcms.com/docs/form-fields/browser.html) specifies a `params` option. However, this option has not been fully implemented.

### Steps to reproduce

Create a browser form field with endpoints and params:
```
<x-twill::browser
    module-name="publications"
    name="publications"
    label="Publications"
    :endpoint="[
        [
            'value' => '/path/to/publication/browser',
            'label' => 'Publications'
        ]
    ]"
    :params="['publication_type' => type]"
/>
```

### Expected result

I expect the params to be appended to the end of the endpoint path as query parameters, for example:
`/path/to/publication/browser?page=1&publication_type=type`

### Actual result

The actual result is that the params are not included in the url query parameters:
`/path/to/publication/browser?page=1`

### Versions

Twill version:  3.5.0
Laravel version: 10.48.29
PHP version: 8.2

zachgarwood:
## Description

This change completes the partially implemented `params` option for the Browser form field.

## Related Issues

Fixes #2751


ifox:
Hi @zachgarwood, I'm probably missing something as your issue describes the problem well so I'm sure you've experienced it, but the params are already supposed to be added to the endpoint(s( URL server side in the Browser field component class: https://github.com/area17/twill/blob/541a8a7017f302d2024acb0e35cedcab23d31ed1/src/View/Components/Fields/Browser.php#L61 and https://github.com/area17/twill/blob/541a8a7017f302d2024acb0e35cedcab23d31ed1/src/View/Components/Fields/Browser.php#L78

It was working the same way in Twill 2. The logic has just been moved from the blade file to the component class.

ifox:
Ok, I understand the problem now. Because you are passing "raw" endpoints, the lines I linked above do not apply. I think the params could still be appended server side instead of through the vuex state in that case.

PR #2763, opened 2025-04-23
Properly set search string in Global Search

### LINKED ISSUE #2762: Global Search not properly getting search string
## Description

Global Search isn't working for me in Twill 3. I verified that it loops through all the items on the dashboard, but the SQL query that is generated doesn't include a condition for `title` for any of the models. It is only checking `where "[model]"."deleted_at" is null`. Looking at the [`cmsSearch()`](https://github.com/area17/twill/blob/3.x/src/Repositories/ModuleRepository.php#L156-L168) method in ModuleController, it's using the FreeTextSearch model and setting:

```
        $searchFilter->queryString($search);
        $searchFilter->searchColumns($fields);
        $searchFilter->searchQuery($query);
```

The FreeTextSearch model, however, has a condition that requires `searchString` to not be empty, and that value is never set in cmsSearch(). It gets set by the `searchFor()` method which never gets called.

Adding:

```
        $searchFilter->searchFor($search);
```

Makes global search work as expected for me.

nikhiltri:
## Description

The [`cmsSearch()`](https://github.com/area17/twill/blob/3.x/src/Repositories/ModuleRepository.php#L156-L168) method in ModuleController uses the FreeTextSearch model and sets:

```
        $searchFilter->queryString($search);
        $searchFilter->searchColumns($fields);
        $searchFilter->searchQuery($query);
```

The FreeTextSearch model, however, has a condition that requires `searchString` to not be empty, and that value is never set in cmsSearch(). It gets set by the `searchFor()` method which never gets called.

Adding:

```
        $searchFilter->searchFor($search);
```

Makes global search work as expected for me.

## Related Issues

Fixes #2762

### DIFF
--- a/src/Repositories/ModuleRepository.php
+++ b/src/Repositories/ModuleRepository.php
@@ -156,7 +156,6 @@ abstract class ModuleRepository
     public function cmsSearch(string $search, array $fields = [], callable $query = null): Collection
     {
         $searchFilter = new FreeTextSearch();
-        $searchFilter->queryString($search);
         $searchFilter->searchFor($search);
         $searchFilter->searchColumns($fields);
         $searchFilter->searchQuery($query);
--- a/src/Repositories/ModuleRepository.php
+++ b/src/Repositories/ModuleRepository.php
@@ -157,6 +157,7 @@ abstract class ModuleRepository
     {
         $searchFilter = new FreeTextSearch();
         $searchFilter->queryString($search);
+        $searchFilter->searchFor($search);
         $searchFilter->searchColumns($fields);
         $searchFilter->searchQuery($query);
 


Tofandel:
You should replace `queryString` altogether with `searchFor` that was the wrong method and it does nothing in this context

src/Repositories/ModuleRepository.php
```diff
@@ -157,6 +157,7 @@ public function cmsSearch(string $search, array $fields = [], callable $query =
     {
         $searchFilter = new FreeTextSearch();
         $searchFilter->queryString($search);
```
Tofandel:
```suggestion
```

ifox (approved):

PR #2755, opened 2025-03-28
Laravel 12 Upgrade

Braunson:
## Description

Updated 3.x to support Laravel 12

- Updates composer and dependencies (including min PHP version to match Laravel 12 - `8.2`)
- Updated workflow test matrix to include Laravel 12

## Related Issues

Fixes #2734

## Notes

Since 12 changes things, we may need to move this to a major release where we drop support for 10.x.. leaving it up to the maintainers (@ifox) and will move this to draft mode.

PR #2765, opened 2025-04-29
Laravel 12 upgrade

Tofandel:
Follow up of #2755 

With the upgrade of larastan a few lint issues popped up, the phpcsfixer config and phpstan config required small updates

Then they were ran, only a few manual fixes were done (eg fireCmsEvent, the method exists would always be true in all supported laravel versions so removed)


Braunson:
@ifox Hoping we can get this in soon :)

ifox (approved):

ifox:
Thank you so much @Braunson and @Tofandel, added PHP 8.4 to the tests matrix and finally merged this. Stay tuned for the release. We know this is way overdue, but we're still here and excited to ship new releases!

Braunson:
@ifox Any ETA on a 12.x release?

PR #2781, opened 2025-09-30
`moduleRoute` - Ensure scalar parameters are wrapped in an array.

### LINKED ISSUE #2780: The `moduleRoute` helper function throws an error
## Description

When the `moduleRoute` helper function is passed a `$parameters` with an `int` or `string` value, it throws an error, despite the function signature specifying those data types:
```php
function moduleRoute(
    string $moduleName,
    ?string $prefix = null,
    string $action = '',
    array|int|string $parameters = [],
    bool $absolute = true
): string 
```

### Steps to reproduce

```
vagrant@homestead:~/website$ php artisan tinker
Psy Shell v0.12.12 (PHP 8.2.28 — cli) by Justin Hileman
> moduleRoute('some_module', parameters: 123);

   TypeError  array_merge(): Argument #2 must be of type array, int given.

> moduleRoute('some_module', parameters: 'string');

   TypeError  array_merge(): Argument #2 must be of type array, string given.
```

### Expected result

I expect `moduleRoute` to construct the url using a scalar `$parameters` as the first route parameter.

### Actual result

The helper function throws an error due to a scalar value being passed to `array_merge` on lines https://github.com/area17/twill/blob/3.x/src/Helpers/routes_helpers.php#L21 and https://github.com/area17/twill/blob/3.x/src/Helpers/routes_helpers.php#L23:

```
TypeError  array_merge(): Argument #2 must be of type array, int given.
```
OR
```
TypeError  array_merge(): Argument #2 must be of type array, string given.
```

### Versions

Twill version: 3.5.2
Laravel version: 11.46.0
PHP version: 8.2.28

zachgarwood:
## Description
In the `moduleRoute` helper function, `$parameters` is allowed to be passed in as an `int` or a `string`, in addition to an `array`; however, the variable is then passed to `array_merge` without any type conversion, and throws an error.

## Related Issues

Fixes #2780


src/Helpers/routes_helpers.php
```diff
@@ -15,6 +14,9 @@ function moduleRoute(
         // Fix module name case
         $moduleName = Str::camel($moduleName);
 
+        // Wrap scalar parameters in an array
+        $parameters = is_array($parameters) ? $parameters : [$parameters];
```
Tofandel:
```suggestion
        $parameters = Arr::wrap($parameters);
```

zachgarwood: Love it! Much simpler!

Tofandel:
I do have a comment on this, aren't parameters an associative array, meaning they need a key?

In which case would passing a string or int to parameters which would result in an indexed array be valid?

zachgarwood:
> I do have a comment on this, aren't parameters an associative array, meaning they need a key?
>
> In which case would passing a string or int to parameters which would result in an indexed array be valid?

If the parameters don't have a string key, they are applied to the URL template after the named params, by position, see [Laravel's `RouteUrlGenerator::replaceRouteParameters()`](https://github.com/laravel/framework/blob/11.x/src/Illuminate/Routing/RouteUrlGenerator.php#L194).

So, given a route defined as:
`Route::get('/parent/{parentId}/child/{childId}/type/{type}', [Controller::class, 'show')->name('show');`
Calling `route('show', parameters: ['type' => 'thing', 123, 456])` would generate the url `http://domain.com/parent/123/child/456/type/thing`.

Getting back to `moduleRoute`, if we only want to pass parameters as associative arrays, here are the lines we would have to update:

- https://github.com/area17/twill/blob/3.x/src/Http/Controllers/Admin/FileLibraryController.php#L148
- https://github.com/area17/twill/blob/3.x/src/Models/Media.php#L81
- https://github.com/area17/twill/blob/3.x/src/Repositories/Behaviors/HandleBlocks.php#L460
- https://github.com/area17/twill/blob/3.x/src/Repositories/Behaviors/HandleBlocks.php#L495
- https://github.com/area17/twill/blob/3.x/src/Repositories/Behaviors/HandleBrowsers.php#L208
- https://github.com/area17/twill/blob/3.x/src/Services/Breadcrumbs/NestedBreadcrumbs.php#L78

Tofandel (approved):

ifox (approved):

PR #2804, opened 2026-02-12
Fix DatePicker timezone handling for non-UTC app.timezone

### LINKED ISSUE #2798: DatePicker: Timezone offset applied incorrectly when parsing dates with time
## Description

The DatePicker component incorrectly applies timezone offsets when parsing dates with time enabled, causing the displayed time to shift by the user's timezone offset from UTC.

For example, if a user in UTC+1 timezone enters `14:00`, the displayed value shifts to `15:00` after saving and reloading.

## Root Cause

Two issues in `frontend/js/components/DatePicker.vue`:

### 1. dateFormat uses 'Z' (ISO 8601 with timezone)

```javascript
dateFormat: (self.enableTime && self.noCalendar) ? 'H:i:S' : (self.enableTime ? 'Z' : 'Y-m-d'),
```

The `'Z'` format outputs ISO 8601 dates with timezone information, but the backend stores and returns dates without timezone info. This mismatch causes parsing issues.

### 2. parseDate incorrectly adds 'Z' suffix

```javascript
return parse(date + 'Z', fullFormat + 'X', Date.UTC());
```

This appends a `'Z'` (UTC indicator) to dates that are actually in local time, causing the timezone offset to be applied twice.

## Proposed Fix

1. Change `dateFormat` to use explicit format without timezone:
```javascript
dateFormat: (self.enableTime && self.noCalendar) ? 'H:i:S' : (self.enableTime ? 'Y-m-d H:i:S' : 'Y-m-d'),
```

2. Simplify `parseDate` to not apply UTC conversions:
```javascript
parseDate: function (date, format) {
  // European format: dd.MM.yyyy HH:mm:ss
  if (/^\d{2}\.\d{2}\.\d{4} \d{2}:\d{2}:\d{2}$/.test(date)) {
    // Manual parsing for European format
    var parts = date.split(' ');
    var datePart = parts[0].split('.');
    var timePart = parts[1].split(':');
    return new Date(datePart[2], datePart[1] - 1, datePart[0], timePart[0], timePart[1], timePart[2]);
  }
  // ... similar for other European formats
  
  // ISO formats - date-fns v1 parse() handles these correctly
  if (/^\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}$/.test(date)) {
    return parse(date);
  }
  // ... etc
  
  return new Date(date);
}
```

## Environment

- Twill version: 3.5.2
- PHP version: 8.2
- Browser: Chrome (also reproducible in Firefox/Safari)
- Timezone: Europe/Berlin (UTC+1)

## Steps to Reproduce

1. Set your system timezone to something other than UTC (e.g., UTC+1)
2. Create a model with a `datetime` field using DatePicker with `enableTime: true`
3. Enter a time value (e.g., 14:00)
4. Save the record
5. Reload the page
6. Observe the time has shifted by your timezone offset (shows 15:00 instead of 14:00)

baeroe:
## Description

When `app.timezone` is set to a non-UTC timezone (e.g. `Europe/Berlin`), the DatePicker applies timezone offsets incorrectly, causing a double-offset when loading dates. This happens because the backend stores dates in `app.timezone` but the frontend always assumes UTC.

This PR fixes the issue by:

- **Always storing UTC in the database** — `HandleDates::prepareDatesField()` now calls `->utc()` after `Carbon::parse()`, ensuring dates are converted to UTC before Eloquent writes them
- **Outputting UTC ISO 8601 to the frontend** — Both `_date_picker.blade.php` and `form.blade.php` (publication dates) now format Carbon instances as `->utc()->toIso8601String()`, giving the browser an unambiguous timestamp
- **Robust `parseDate` in DatePicker.vue** — Replaced length-based format detection with regex-based matching that correctly handles ISO 8601 with timezone, ISO without timezone, date-only, European dot-separated (`dd.MM.yyyy`), and US slash-separated (`MM/dd/yyyy`) formats
- **Fixed `PubAccordion.vue`** — `startDateForDisplay` now detects ISO 8601 strings with timezone info instead of blindly appending `Z`
- **Added non-UTC timezone browser tests** — New `FrontendDateConversionNonUtcTest` and `ScheduleDatesNonUtcTest` verify correct behavior with `app.timezone=Europe/Berlin`

## Related Issues

Fixes #2798

### DIFF
--- a/frontend/js/components/DatePicker.vue
+++ b/frontend/js/components/DatePicker.vue
@@ -15,7 +15,6 @@
 <script>
   import 'flatpickr/dist/flatpickr.css'
 
-  import parse from 'date-fns/parse'
   import FlatPickr from 'flatpickr'
 
   import FormStoreMixin from '@/mixins/formStore'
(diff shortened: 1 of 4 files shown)


ifox:
Hi @baeroe, thanks for taking care of this!

It seems like your description doesn't match the diff:
>HandleDates::prepareDatesField() now calls ->utc() after Carbon::parse()

There are no changes to `HandleDates`.

>New FrontendDateConversionNonUtcTest and ScheduleDatesNonUtcTest

These new tests are not in this PR.

PR #2820, opened 2026-03-27
Feature/3.x ru field translations

gvidobluefox:
• ## What Was Done

  Updated Russian translations in `lang/ru/lang.php` based on `lang/en/lang.php` and other locale files.

  ## Main Changes

  - added missing keys from the English locale
  - aligned the `ru` locale structure with `en`
  - translated missing strings into Russian
  - improved wording based on other locale files where it helped with meaning
  - replaced remaining English values in the Russian locale
  - standardized email labels as `E-mail`
  - changed `alt-text` to the more explicit Russian wording meaning `image description`

  ## Validation

  - verified PHP syntax with `php -l lang/ru/lang.php`

  ## Result

  The Russian locale is now more complete, better aligned with `en`, and more consistent in terminology.


lang/ru/lang.php
```diff
@@ -1,68 +1,103 @@
...
+        'unknown-author' => 'Неизвестно',
     ],
     'dialog' => [
         'cancel' => 'Отмена',
         'ok' => 'ОК',
         'title' => 'Отправить в корзину',
     ],
     'editor' => [
-        'cancel' => 'Отмена',
+        'cancel' => 'Закрыть',
```
ifox: @gvidobluefox the previous translation was better for this, no? It looks like `Закрыть` means close, not cancel. In the UI, nothing gets closed when using the button that references this key, it cancels the changes made to the selected.

lang/ru/lang.php
```diff
@@ -1,68 +1,103 @@
...
+            'restored' => 'Восстановлено',
+            'deleted' => 'Удалено',
+            'login' => 'Вход',
+            'logout' => 'Выход',
+            'duplicated' => 'Скопировано',
+        ],
+        'activity-row' => [
+            'edit' => 'Редактировать',
+            'view-permalink' => 'Показать ссылку',
+            'by' => '. Редактор - ',
```
ifox, line 55: Not sure about the punctuation for this. Claude suggests `пользователем`.

lang/ru/lang.php
```diff
@@ -111,11 +149,38 @@
...
+            'published' => 'Успешно опубликовано',
+            'unpublished' => 'Успешно снято с публикации',
+            'error' => 'Не удалось изменить статус публикации выбранных записей',
+        ],
+        'bulk-restore' => [
+            'success' => 'Выбранные сущности успешно восстановлены',
+            'error' => 'Не удалось восстановить выбранные сущности',
+        ],
+        'bulk-selected-item' => 'выбрано',
+        'bulk-selected-items' => 'выбрано',
```
ifox, line 181: EN distinguishes singular/plural. According to Claude, using the same string for both in RU will produce wrong grammar.

lang/ru/lang.php
```diff
@@ -200,17 +342,21 @@
...
     'notifications' => [
         'reset' => [
             'action' => 'Сброс пароля',
             'content' => 'Вы получили это письмо, потому что мы получили запрос на сброс пароля. Если вы не запрашивали сброс пароля, никаких дальнейших действий не требуется.',
+            'subject' => ':appName | Сброс пароля',
         ],
         'welcome' => [
-            'content' => 'Вы получили это письмо, потому что для вас была создана учетная запись :name.',
+            'action' => 'Выберите пароль',
+            'content' => 'Вы получили это письмо, потому что для вас была создана учетная запись на :name.',
```
ifox: Claude said:

> "на :name" is wrong. EN says "on :name" but the correct Russian preposition for a software platform is "в :name"

ifox (changes requested):
Thanks @gvidobluefox. I do not speak Russian myself, but I got Claude to review this, so I added a couple comments. Let me know what you think, but this looks good to merge once you've updated or denied Claude's claims!

lang/ru/lang.php
gvidobluefox: Yes, you're right. Thank you for your attention. I looked at the functionality in more detail and saw that the changes were indeed being canceled. I'll fix it with the next commit.

lang/ru/lang.php
gvidobluefox: In Russian, names are inclined, but there are no declensions, so it doesn't look very pretty. In the case of the proposed punctuation, the text turns out to be beautiful without declension, for example:

_Updated 2 minutes ago. The editor - Admin._
_Обновлено 2 минуты назад. Редактор - Admin._

lang/ru/lang.php
gvidobluefox: In this case, besides plural forms, we also need to consider the grammatical gender of the items. In English, you can simply say “selected,” but in Russian it’s more complex.

Let’s take three example entities that might exist in an admin panel:

Товар (Product) — masculine
Книга (Book) — feminine
Сообщение (Message) — neuter

If we look at the singular form, it would be:

1 товар выбран
1 книга выбрана
1 сообщение выбрано

However, in Russian, the word “selected” changes depending on the gender of the noun, so each of these would have a different form.

Because of this, when dealing with something like “bulk-selected-item,” you also need to adjust it based on grammatical gender. To simplify such cases in Russian interfaces, it’s common practice to normalize everything to the neuter gender.

lang/ru/lang.php
gvidobluefox: Yes, that's also true. I thought that :name was the username, but upon closer inspection, I saw that it was the name of the Laravel application.

PR #2823, opened 2026-03-31
Mitigate Axios supply-chain compromise (#2822)

### LINKED ISSUE #2822: Axios supply-chain compromise
## Description

Use of Axios opens users to supply-chain attack, see https://github.com/axios/axios/issues/10611

### Versions

Twill version:  3.x

zachgarwood:
Set the specified version of axios to a known safe version and do not allow transient dependencies to override this version.

See https://github.com/axios/axios/issues/10611

Fixes #2822

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,7 +23,7 @@
         "@tiptap/starter-kit": "^2.2.1",
         "@tiptap/vue-2": "^2.2.1",
         "alpinejs": "^3.13.5",
-        "axios": "^0.30.2",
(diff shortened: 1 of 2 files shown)


ifox (approved):
Thanks for this! 

Since Twill is versioning both the lock file and a compiled build of the frontend assets, the vulnerability should not impact consumers, but could definitely impact maintainers and contributors.

A lot of unnecessary dependencies like Axios will be removed in Twill 4, so pinning the version is a good approach for now.

PR #2831, opened 2026-05-04
Laravel 13 support

ifox:
This PR bumps Twill to 3.6.0 and widens Composer constraints so installs can target Laravel 13 (and matching testbench). CI now runs Laravel 13 too.

`cartalyst/tags` is abandoned, see ([cartalyst.com](https://cartalyst.com/)), so we stop depending on the Packagist package and ship the same API ourselves under `src/Support/Cartalyst/Tags/`, still autoloaded as `Cartalyst\Tags\...`. That way we can support Laravel 13 on Twill 3 without a breaking change for people already using those classes.

We block  `kalnoy/nestedset:7.0.1` because a bad release was getting resolved on Laravel 10/11 when it shouldn’t. Context: [laravel-nestedset#641 (comment)](https://github.com/lazychaser/laravel-nestedset/issues/641#issuecomment-4286282469).


Tofandel:
> So we stop depending on the Packagist package and ship the same API ourselves under src/Support/Cartalyst/Tags/, still autoloaded as Cartalyst\Tags\...

Then in the composer.json, there should be a `"replace": [ "cartalyst/tags" ]` to instruct composer about this so the package doesn't need to be installed if it was required somewhere else

Jigsaw5279:
Any timeline for the official 3.6.0 release..?

PR #2845, opened 2026-09-04
Make admin email login case-insensitive

### LINKED ISSUE #2789: Admin user authentication treats email addresses as case-sensitive
## Description

Twill's admin authentication appears to treat email addresses as case-sensitive, causing login failures when users enter their email in a different case than how it was originally stored.

### Steps to reproduce

1. Create a Twill admin user with email `User@example.com`
2. Attempt to log in with `user@example.com`
3. Login fails

### Expected result

Email addresses should be treated as case-insensitive for authentication. A user registered as `John.Smith@example.com` should be able to log in with `john.smith@example.com`.

- **RFC 5321** recommends treating email addresses as case-insensitive for practical purposes
- All major email providers (Gmail, Outlook, Yahoo) treat addresses as case-insensitive
- Mobile devices often auto-capitalize the first letter, causing unexpected login failures
- This is a common UX friction point that leads to support requests

### Actual result

Login fails when the email case doesn't exactly match the stored value.

### Versions

- Twill version: 3.5.2
- Laravel version: 11
- PHP version: 8.3

wakqasahmed:
## Description

`LoginController::credentials()` passes the raw email input straight to the auth guard, so login is case-sensitive whenever the `email` column's collation is case-sensitive (e.g. Postgres, or MySQL with a `_bin`/case-sensitive collation) — a user created as `User@example.com` can't log in as `user@example.com`. Changed `credentials()` to match on `LOWER(email)` via a closure credential (Laravel's `EloquentUserProvider` has supported closures for custom query building since Laravel 8), so the match no longer depends on collation. Added an integration test that forces a case-sensitive collation on `email` for the duration of the test and restores it after — the default collation in this repo's own MySQL test DB is already case-insensitive, so a plain "log in with different case" test wouldn't have actually caught the regression.

Used AI assistance (Claude) to write and test this; reviewed the diff and ran the full LoginTest suite plus php-cs-fixer before opening.

## Related Issues

Fixes #2789

### DIFF
--- a/src/Http/Controllers/Admin/LoginController.php
+++ b/src/Http/Controllers/Admin/LoginController.php
@@ -306,7 +306,17 @@ class LoginController extends Controller
      */
     protected function credentials(Request $request): array
     {
-        return array_merge($request->only($this->username(), 'password'), ['published' => 1]);
+        $credentials = $request->only($this->username(), 'password');
+
+        $email = $credentials[$this->username()] ?? null;
+
+        if ($this->username() === 'email' && is_string($email)) {
+            $credentials[$this->username()] = function ($query) use ($email) {
+                $query->whereRaw('LOWER(email) = ?', [mb_strtolower($email)]);
+            };
+        }
+
+        return array_merge($credentials, ['published' => 1]);
     }
 
     protected function autologin(): bool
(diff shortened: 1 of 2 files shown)


wakqasahmed:
Hi @ifox — noticed this PR doesn't have a reviewer assigned yet — it's been sitting for a while, CI is green and it's mergeable. Would you (or whoever's best placed) be able to take a look, or point me to who should? Thanks!

wakqasahmed:
Hi @ifox — circling back on this one — still no reviewer assigned, quiet for a while since the last activity. No pressure, just don't want it to get lost.

